PR #2, opened 2018-04-16
Check remote calls and load remote module types on demand

zuiderkwast:
Adds a module `gradualizer_db`, whichs extracts specs and types from other modules, apps or the whole OTP.

A gen_server is used which needs to be started using gradualizer_db:start_link().

The main operation is to load a module on-demand, when a module is needed. Functions to load and save types and specs to files are included but they are not very fast. Perhaps on-demand parsing is the best so far.

Integrated in typechecker and used for remote calls. Not used yet for references to remote types.

Mostly tested by itself, manually. Not very well-tested with the typechecker, since something you added (maps:update_with/4) makes it not run on my old Erlang 18 installation anymore.

### DIFF
--- a/gradualizer_db.erl
+++ b/gradualizer_db.erl
@@ -1,0 +1,391 @@
+%% Copyright 2018 Viktor Söderqvist
+%% See the LICENSE file shipped with this file.
+
+%% @doc Collect exported functions and types from multiple files.
+%%
+%% For exported functions with missing spec, a spec is generated with any()
+%% as the type for all parameters and return values.
+-module(gradualizer_db).
+
+%% API functions
+-export([start_link/0,
+         get_spec/3, get_type/3,
+         save/1, load/1,
+         import_files/1, import_app/1, import_otp/0]).
+
+%% Callbacks
+-behaviour(gen_server).
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
+         code_change/3]).
+
+%% Types for the Erlang Abstract Format
+-type line() :: non_neg_integer().
+
+-type type() :: {type, line(), atom(), [type()] | any} |
+                {var, line(), atom()} |
+                {integer, line(), integer()} |
+                {atom, line(), atom()}.
+
+%% Gen server local registered name
+-define(name, ?MODULE).
+
+%% Internal data
+-record(typeinfo, {exported :: boolean(),
+                   opaque   :: boolean(),
+                   params   :: [{var, _, atom()}],
+                   body     :: type()}).
+
+%% Public API functions
+
+start_link() ->
+    gen_server:start_link({local, ?name}, ?MODULE, #{}, []).
+
+-spec get_spec(M :: module(), F :: atom(), A :: arity()) ->
+        {ok, [type()]} | not_found.
+get_spec(M, F, A) ->
+    gen_server:call(?name, {get_spec, M, F, A}).
+
+-spec get_type(Module :: module(), Type :: atom(), Params :: [type()]) ->
+        {ok, type()} | not_found.
+get_type(M, T, A) ->
+    gen_server:call(?name, {get_type, M, T, A}).
+
+-spec save(Filename :: any()) -> ok | {error, any()}.
+save(Filename) ->
+    gen_server:call(?name, {save, Filename}).
+
+-spec load(Filename :: any()) -> ok | {error, any()}.
+load(Filename) ->
+    gen_server:call(?name, {load, Filename}).
+
+import_files(Files) ->
+    gen_server:call(?name, {import_files, Files}, infinity).
+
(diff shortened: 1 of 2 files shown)


josefs:
This is great stuff, Victor!

A few comments:

- I don't think this approach will work for types and records in other modules, since those are typically handled with header files. And that's totally fine! It just means that we'll have to have a way to deal with header files as well.
- I'm afraid I broke your pull request. I made an overhaul of how the environment is passed around to support records. That overhaul touched a lot of lines. Can I ask you to resolve the conflict?
- I would rather I hadn't have to use the maps:update_with/4. All I want is a function which combines two maps and have a way to control what happens when the same key occurs in both maps. Unfortunately, that function doesn't exist so I had to roll it myself. Hence the maps:update_with/4. If you have a way to do it that works better for you I'll gladly consider it.
- The copyright statement at the top the file you added is incompatible with the license file. Think about how you want to resolve that.

Again, great stuff!

zuiderkwast:
> I don't think this approach will work for types and records in other modules, since those are typically handled with header files. And that's totally fine! It just means that we'll have to have a way to deal with header files as well.
Yes, include files is a problem. Even within OTP, there are odd stuff in Makefiles for some applications. I added warnings for epp missing includes and other epp errors and also a pragmatic include path that solves most problems. Try `gradualizer_db:import_otp()` to see which errors still occur.

To minimize the problems with unexported types and records in other modules, I chose *not to expand* them until that is needed. Normally, an abstract datatype (opaque) shouldn't be decomposed in other modules. Thus, an exported type that includes an unexported type from some unresolved include file may still pass without a problem.

> 'm afraid I broke your pull request. I made an overhaul of how the environment is passed around to support records. That overhaul touched a lot of lines. Can I ask you to resolve the conflict?
That's ok. I have resolved it.

> I would rather I hadn't have to use the maps:update_with/4. All I want is a function which combines two maps and have a way to control what happens when the same key occurs in both maps. Unfortunately, that function doesn't exist so I had to roll it myself. Hence the maps:update_with/4. If you have a way to do it that works better for you I'll gladly consider it.

It's fine. I have updated to OTP 19.

> The copyright statement at the top the file you added is incompatible with the license file. Think about how you want to resolve that.

I have removed it. Perhaps it's enough to put the authors name in the LICENSE file, but I think the names of all those who write substantial parts of the code should be in there as attribution is a requirement of the license.

In many projects, MIT and similar small licenses are included in each file to make sure you always get it with the source code. For large license texts e.g. GPL, the recommended way is to put a note in each file referring to the license text and how to get it from FSF if you didn't get it with the source code. That's why I wrote my name and a reference to the LICENSE file there.

----

Still TODO for this PR is to modify the types returned by get_type and get_spec so that the typechecker knows in which module the type is. Something like turning user-defined types into remote types (i.e. `my_type()` -> `mod:my_type()`) inside the returned types.

zuiderkwast:
Yes, I hadn't finished the changes that I wrote about before :-) Now I have pushed them.

Only get_spec/3 is called by typechecker, for remote function calls. The get_type functions are yet to be integrated. Two 'special' types are introduced for local (possibly unexported) types and record types annotated with the module in which they are defined.

Here are now some things to try, manually:

```Erlang
1> gradualizer_db:start_link().
{ok,<0.59.0>}
2> gradualizer_db:get_exported_type(sets, set, []).
{ok,{remote_local_type,74,sets,set,[{var,74,'_'}]}}
3> gradualizer_db:get_exported_type(sets, set, [{any, 0, []}]).
{ok,{opaque,sets,set,[{any,0,[]}]}}
4> gradualizer_db:get_exported_type(ordsets, ordset, []).      
not_found
5> gradualizer_db:get_exported_type(ordsets, ordset, [{any, 0, []}]).
{ok,{type,31,list,[{any,0,[]}]}}
6> gradualizer_db:get_spec(maps, update_with, 4).                    
{ok,[{type,162,bounded_fun,
           [{type,162,'fun',
                  [{type,162,product,
                         [{var,162,'Key'},
                          {var,162,'Fun'},
                          {var,162,'Init'},
                          {var,162,'Map1'}]},
                   {var,162,'Map2'}]},
            [{type,163,constraint,
                   [{atom,163,is_subtype},
                    [{var,163,'Key'},{type,163,term,[]}]]},
             {type,164,constraint,
                   [{atom,164,is_subtype},[{var,164,'Map1'},{var,164,'Map1'}]]},
             {type,165,constraint,
                   [{atom,165,is_subtype},[{var,165,'Map2'},{var,165,'Map2'}]]},
             {type,166,constraint,
                   [{atom,166,is_subtype},
                    [{var,166,'Fun'},
                     {type,166,'fun',
                           [{type,166,product,...},{ann_type,166,...}]}]]},
             {type,167,constraint,
                   [{atom,167,is_subtype},
                    [{var,167,'Init'},{type,167,term,[]}]]}]]}]}

```

PR #5, opened 2018-06-04
Fix var bindings unioning

gomoripeti:
When returned from a `lists:unzip3` `Varbinds` is a list of maps, not just a
map. It could probably help if the variable names could highlight this
difference (eg Varbind vs Varbinds, or Varbinds vs VarbindsList). I am
happy to do the renaming everywhere.

### DIFF
--- a/typechecker.erl
+++ b/typechecker.erl
@@ -645,7 +645,7 @@ type_check_expr(Env, {call, _, Name, Args}) ->
 			       end, ArgTys, Args)
 		 ),
 	    { ResTy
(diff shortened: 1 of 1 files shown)


josefs:
Good catch and thanks a lot for the patch! We really need a type system to catch these kinds of errors :-)

It would be fantastic if you wanted to do the renaming everywhere. I'm painfully aware that the naming is inconsistent in the code base.

gomoripeti:
(varbinds renaming is done as part of https://github.com/josefs/Gradualizer/pull/8)

PR #6, opened 2018-06-05
Add support for annotated types in specs

gomoripeti:
The varname of an annotated type only serves documentation purposes
and has no role in the type system.


josefs:
Actually, as part of my ongoing work on typevariables and polymorphism, I'd like the varname in an annotated type to matter! If I have a type which says `{A :: term(), A :: number()}` as in your example, then `A` needs to have a type which is both a subtype of `term()` and of `number()`. So I'd like **not** to throw the varname away.

josefs:
I've added support for annotated types now, by collecting them as constraints on type variables. This supersedes your patch. However, the tests you wrote in typechecker_tests are still very useful. If you update your patch I'd love to include them.

gomoripeti:
great!
I admit I was too much thinking in the narrow way of the meaning of type-specs for Dialyzer. But obviously beyond the syntax type-specs can mean much more or different things in a different type system.

I will update the PR to only add the tests on top of latest master.

PR #8, opened 2018-06-10
Various small fixes

gomoripeti:
- Consistent naming of varbinds variables
- address Dialyzer warnings
- address some crashes during self-type checking
  current output is:
```
> typechecker:type_check_file("typechecker.erl").
The variable 'Ty1' on line 47 has type type() but is expected to have type type() in typelib.erl
```
I don't know if it is expected or I introduced a regression. The type `type()` both in `typechecker.erl` and `typelib.erl` should be subtype of `erl_parse:abstract_type().`

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Right now the user experience is not very polished. Here's how to get started:
 
   You can try typechecking some of the example modules in the `tests`directory.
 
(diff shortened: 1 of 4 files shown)


josefs:
Great work! I'm aware of the current error that Gradualizer gives when checking itself. You did not introduce it. I'll look into it.

PR #9, opened 2018-06-12
 Handle top-level bounded functions 

gomoripeti:
The second commit is just an optional idea, to reduce cases that need to be handled by merging different Abstract Formats.


josefs:
Thanks for another nice PR! I like the idea of always working with bounded functions to help simplify the code!

PR #10, opened 2018-06-14
 OTP application file structure 

gomoripeti:
use erlang.mk as build tool.

(Because erlang.mk wants to compile all erlang files under every subdirectory of `test/` dir, I had to fix a few example files)


josefs:
This is really useful stuff! Thanks!

For posterity and the sake of openness, I'll repeat what I said per email:

I would prefer that we created a separate module called gradualizer which contains the main entry point and keep the name of the typchecker module. It will be useful to separate out the gradualizer module if/when we add command line parameter handling and such. Those things don't belong in the typechecker module.

gomoripeti:
rebased to latest master
name of `typechecker` module is kept (as that is the one containing the main type checking logic)

josefs:
Nice job! I've merged your changes.
One comment: I see that you've called the command `gradualize` without the `r` at the end. I'm in two minds about this. One the one hand, it's fun that the command is a verb so that you can "gradualize your files". On the other hand, it's somewhat confusing that the command is not called the same things as the general application.
Perhaps you can explain your thoughts on the name, Peter?

gomoripeti:
I only chose this name because erlang.mk calls target running dialyzer also `make dialyze`.
Maybe this is a general `make` style to call targets which are not build artifacts as verbs or commands.
(checking rebar3 it is called `rebar3 dialyzer` so it's not super-consistent across build tools)

josefs:
Ok. Based on your explanation here's what I suggest: we follow the convention used by dialyzer and call the make target `gradualize` and call the tool `gradualizer`. 
I think that looks kind of nice, and it has the advantage of being aligned with dialyzer so people will recognize it. How does that sound to you?

PR #12, opened 2018-06-15
mix compatibility

tim2CF:
With mix utility Gradualizer can be used by Elixir developers


tim2CF:
And with `mix.exs` file package can be published in `hex.pm`. It can be done with rebar3 plugin as well. Do you have any plans about it? Publishing in hex.pm is nice opportunity to get new users / maintainers. Elixir guys usually are looking for packages in hex.pm, not in github.

josefs:
Enabling Gradualizer for Elixir programmers is really cool! I'm excited by these additions!

This PR has already gone stale, and Peter is going to make a few more changes to his PR. After that, I'm going to ask you to update this PR so that we can include it.

My plan w.r.t. hex has been that I plan to upload once the Gradualizer has full support of all of Erlang. If you disagree and think that it should be uploaded sooner, just let me know. I'm open to persuasion.

A minor comment for the future: Please separate whitespace changes into separate commits.

tim2CF:
Good! There is only one reason to upload Gradualizer to hex.pm sooner - to occupy public package name, because you cannot publish OTP app to hex.pm if app with the same name has been published before by some other developer

josefs:
Now that we've merged Peter's PR, would you be willing to update this PR so that we can include it?

tim2CF:
yes, I'll update!

PR #13, opened 2018-06-22
Small fixes

gomoripeti:
(I am afraid I still cannot push to this repo directly, so submitting PR)

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -362,8 +362,8 @@ normalize({user_type, P, Name, Args} = Type, TEnv) ->
                     throw({undef, user_type, {Name, length(Args)}})
             end
     end;
-normalize({remote_type, _P, [Module, Name, Args]} = RemoteType, TEnv) ->
-    case gradualizer_db:get_exported_type(Module, Name, Args) of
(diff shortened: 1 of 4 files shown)


src/typechecker.erl
```diff
@@ -1436,6 +1435,16 @@ glb_types({type, _, N, Args1}, {type, _, N, Args2}) ->
...
 
+get_rec_field_type(FieldName,
+                   [{typed_record_field,
+                     {record_field, _,
+                      {atom, _, FieldName}, _}, Ty}|_]) ->
+    Ty;
+get_rec_field_type(FieldName, [_|RecFieldTypes]) ->
+    get_rec_field_type(FieldName, RecFieldTypes);
+get_rec_field_type(_, []) ->
+    {error, not_found}.
```
josefs: I would suggest throwing an exception here instead.

josefs:
Not sure why you can't push directly to the repo. Github says "Awaiting gomoropeti's resonse". What happens if you go to https://github.com/josefs/Gradualizer/invitations ?

src/typechecker.erl
gomoripeti: 👍 pushed a change to throw an exception.

gomoripeti:
I thought it was only a notification, but I indeed had to push the button. Now I can happily push, thank you :)

PR #14, opened 2018-06-24
Extend API

gomoripeti:
Introduce `gradualizer` module which exports following API functions:
- `type_check_module/1/2` (useful for eg. mix task)
- `type_check_file/1/2` that works both for source and beam files
  depending on extension
- `type_check_dir/1/2` that type checks all source and/or beam files in
  the given directory

The main entry point of the core `typechecker` module is `type_check_forms/2`.

@josefs, please let me know what do you think about this reorganization, options
@tim2CF I took your idea, and expanded it a bit, does this look good for a mix task?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ DIALYZER_OPTS = -Werror_handling
 include erlang.mk
 
 gradualize: app
-	$(ERL) -pa $(CURDIR)/ebin -eval 'typechecker:type_check_file("src/typechecker.erl"), halt().'
+	$(ERL) -pa $(CURDIR)/ebin -eval 'gradualizer:type_check_file("src/typechecker.erl"), halt().'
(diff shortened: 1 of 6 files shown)


josefs:
Great! This is the kind of structure I've been envisioning. Solid job!

josefs:
@gomoripeti, I realize now you might have wanted to wait for comments from @tim2CF. Sorry that I got so trigger happy with the merge button. I was just excited about these changes.

gomoripeti:
It's ok :) I just wanted to indicate that the other PR is superseded.

(I also have to admit that the changed default behaviour of not stopping on first error unleashed an infinite chain of crashes on not yet handled ASTs in the self-check so `make gradualize` is not that useful any more. But at least we see what we are working towards)

PR #17, opened 2018-06-26
Fix normalizing local user types

### LINKED ISSUE #15: "... has type type() but is expected to have type type()"
When I run `make gradualize` to gradualize the gradualizer itself, it crashes in typechecker.erl with the following error:

> The variable 'Ty1' on line 60 has type type() but is expected to have type type()

I tried to solve it but I got confused. This is how far I got.

The first `type()` is defined in typechecker (the currently checked module) while the second is defined in `typelib`. (This can be seen by uncommenting some commented code in `typelib:pp_type/1` which checks for filename annotation in the form and adds it to the output.)

So far so good. Next, to see what these `type()` expand to, I modified the exception caught in `subtype/4` from `nomatch` to `{nomatch, T1, T2}` and the corresponding `throw` and printed out what exactly mismatches when comparing two types piece by piece. Now I can see that they are both normalized/expanded to `{ann_type, anno(), [af_anno() | abstract_type()]}` (where the types are local to erl_parse). There is no function clause in `compat_ty` for the case when both types are ann_type, but there is a clause that matches identical types `compat_ty(T, T, A, _) -> ...` which should match here. I don't understand why it doesn't match. I am confused.

What am I missing?

gomoripeti:
When a user type in the same module as type checked was encountered it
was not normalized further recursively. This was not a problem for user
types in other modules or remote types.

Fixes self-typecheck issue:
> The variable 'Ty1' on line 60 has type type() but is expected to have type type()

closes #15

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -324,7 +324,8 @@ normalize({user_type, P, Name, Args} = Type, TEnv) ->
             case TEnv#tenv.types of
                 #{TypeId := {Vars, Type0}} ->
                     VarMap = maps:from_list(lists:zip(Vars, Args)),
-                    typelib:substitute_type_vars(Type0, VarMap);
+                    Type1 = typelib:substitute_type_vars(Type0, VarMap),
+                    normalize(Type1, TEnv);
                 _NotFound ->
                     throw({undef, user_type, {Name, length(Args)}})
             end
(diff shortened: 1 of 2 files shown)


zuiderkwast:
That's it! I confirm that this is correct. Thanks Peter!

PR #18, opened 2018-07-06
 Improve handling remote and user types

gomoripeti:


test/should_pass/user_types.erl
```diff
@@ -0,0 +1,15 @@
+-module(user_types).
+
+-export([f/1]).
+
+%% these exported types are used in remote_types module
+-export_type([my_tuple/0, my_list/0, my_union/0]).
+
```
zuiderkwast: So this file is a test of it's own as well as a dependency for another test. That's interesting... I think it's fine. It's good that it is noted in the comment.

src/typechecker.erl
```diff
@@ -89,8 +88,8 @@ subtypes([Ty1|Tys1], [Ty2|Tys2], TEnv) ->
 %% The function compat_ty is just a convenience function to be able to
 %% pattern match on types in a nice way.
 compat(T1, T2, A, TEnv) ->
-    Ty1 = normalize(T1, TEnv),
-    Ty2 = normalize(T2, TEnv),
+    Ty1 = typelib:remove_pos(normalize(T1, TEnv)),
```
zuiderkwast: This makes a lot of sense to me! Good solution.

src/typechecker.erl
```diff
@@ -1531,8 +1531,17 @@ add_type_pat_tuple(Pats, {type, _, union, Tys}, TEnv, VEnv) ->
 			    , length(TS) == length(Pats)])),
     lists:foldl(fun ({Pat, Union}, Env) ->
 			add_type_pat(Pat, Union, TEnv, Env)
-		end, VEnv, lists:zip(Pats, Unions)).
-
+                end, VEnv, lists:zip(Pats, Unions));
+add_type_pat_tuple(Pats, Ty, TEnv, VEnv)
+  when element(1, Ty) =:= user_type;
+       element(1, Ty) =:= remote_type ->
+    case normalize(Ty, TEnv) of
```
zuiderkwast: I think normalizing Ty should be done earlier. I think it applies for most clauses of `add_type_pat`, at least for tuple, binary, cons and record. Perhaps it can be done for all these cases of `add_type_pat` using a recursive clause which is placed before the clauses for tuple, cons, bin and record:

```Erlang
add_type_pat(Pat, Ty, TEnv, VEnv)
  when element(1, Ty) =:= user_type;
       element(1, Ty) =:= remote_type ->
    case normalize(Ty, TEnv) of
        Ty ->
            %% no change, unknown case
            throw({unsupported_type, Ty});
        Ty2 ->
            add_type_pat(Pat, Ty2, TEnv, VEnv)
    end;
```

src/typechecker.erl
```diff
@@ -1531,8 +1531,17 @@ add_type_pat_tuple(Pats, {type, _, union, Tys}, TEnv, VEnv) ->
...
+        Ty ->
+            %% no change, unknown case
+            throw({unsupported_type, Ty});
```
zuiderkwast: It would expand to itself for remote opaque types. Remote opaque types shouldn't be used in pattern matching against tuples and other decomposing patterns. Perhaps the error message can reflect that.

src/typechecker.erl
```diff
@@ -1014,7 +1014,10 @@ type_check_expr_in(Env, ResTy, {tuple, LINE, TS}) ->
 	    % tuple in a union. That's potentially inefficient.
 	    % Maybe we should have a flag to allow for approximation here
 	    % in the same way that dialyzer does it.
-	    {VBs, Cs2} = type_check_tuple_in(Env, ResTy, TS),
+
+        %% normalize ResTy to resolve user and remote types
+        NormResTy = normalize(ResTy, Env#env.tenv),
```
zuiderkwast: I think normalizing should be done for several clauses of `type_check_expr_in/4`, at least for cons, bin, tuple and record. I noticed this when I implemented `{bin, ...}`. Perhaps we can handle this in a single place, by adding a clause of type_check_expr_in for user and remote types, normalize and then a recursive call? Similar reasoning here as for normalizing in add_type_pat.

zuiderkwast:
To answer this
> maybe there should be a separate expand_user_types function, to avoid executing other normalizations too early

For bin expressions and pattens, we need normalization of "built in type aliases", e.g. `binary()` => `<<_:0, _:_*8>>`. I don't think it is a problem to do a full normalization.

test/should_pass/user_types.erl
gomoripeti: This is correct.
These should pass/should fail files are very handy to add just random examples, but a bit harder to add well structured test cases covering many variations. But at this stage of the project it is good enough.

src/typechecker.erl
gomoripeti: good point!

gomoripeti:
- full normalization: you are right, it's not only expandable types. I also hit built-in aliases like `arity()`
- where to normalize?: 
  - one option is to do it before calling `type_check_exp_in` something like
    ```
    type_check_expr_in(Env, ResTy, Expr) ->
      do_type_check_expr_in(Env, normalize(ResTy), Expr).
    ```
    one small issue is that the error messages will show the normalized `ResTy` and not the original one (I don't know if this is an issue)
  - other option I can think of is to never pattern match on `ResTy`, only on `Expr`, and use `subtype/3`, `compatible/3` and similar utility functions that decompose types (eg `get_list_element_type(Ty)`). Something like handling type as an opaque structure in most parts of the code.

zuiderkwast:
> * where to normalize?:
>
>    * one option is to do it before calling type_check_exp_in something like
>      ```
>      type_check_expr_in(Env, ResTy, Expr) ->
>          do_type_check_expr_in(Env, normalize(ResTy), Expr).
>      ```
>      one small issue is that the error messages will show the normalized ResTy and not the original one (I don't know if this is an issue)

If we do this, we can catch the type error and re-throw it with the original type and the same trace:

```Erlang
type_check_expr_in(Env, ResTy, Expr) ->
    try
        do_type_check_expr_in(Env, normalize(ResTy), Expr)
    catch
        throw:{type_error, ..., Ty}:Stacktrace when Ty /= ResTy ->
            erlang:raise(throw, {type_error, ..., ResTy}, Stacktrace)
    end.
```

Yes, I think the original type is good to have in the error message. The normalized type could be good to have as well, though, to see what exactly which part inside the type is mismatching.

When comparing two types in `subtype` in the recursion (e.g. each element in a tuple type is compared), it could be useful to have the original types as well as the mismatching "part" of the type. Something like a trace consisting of types if this is done recursively.

>    * other option I can think of is to never pattern match on ResTy, only on Expr, and use subtype/3, compatible/3 and similar utility functions that decompose types (eg get_list_element_type(Ty)). Something like handling type as an opaque structure in most parts of the code.

Sure, but I don't think moving the normalization to a help function like `get_list_element_type` is very different from doing normalization in the type_check_expr_in clause directly.

src/typechecker.erl
josefs: I agree, it's a very nice and surgical fix.

josefs:
Regarding where to normalize. @gomoripeti can I ask you to explain in more detail your worry about calling normalize the way we do it now? I'm sure your worries are well founded, it's just that I'm not sure I understand exactly what it is that you're worried about.

> - other option I can think of is to never pattern match on ResTy, only on Expr, and use subtype/3, compatible/3 and similar utility functions that decompose types (eg get_list_element_type(Ty)). Something like handling type as an opaque structure in most parts of the code.
> 
If we were to change where we normalize, this would be my prefered solution. We're fairly close to this solution anyway, and I think the places where we actually do pattern match on types would be easy to handle the way you suggest. I like the idea of treating types as abstract when traversing the expression tree.

@zuiderkwast Your point about having to change the exception so that we report the right type is well made. It's important that we don't lose the quality of the error messages.

gomoripeti:
problem with only normalizing types within `subtype` call is that all the bellow functions/functionalities have to handle user/remote types and all built-in aliases themselves:

- `type_check_expr_in`
- `type_check_cons_in`
- `type_check_tuple_in`
- `type_check_assocs
- `check_clauses
- `type_check_arith_op_in`
- `type_check_in`t_op_in`
- `type_check_rel_op_in`
- `type_check_list_op_in`
- `type_check_block_in`

for example `type_check_cons_in` need to know what is the element type of the list, and for this first it needs to expand a user type. If they are guaranteed to only receive normalized types, they only need to handle a subset of cases.

I'm pushing a new version of this branch implementing option 1. (This allows to use utility functions, but leaves the option for pattern matching, and it is not really different from option 2. as Viktor pointed out). I haven't implemented the "catch and re-throw" yet (and the "type-trace"), although I think that's a good idea, the error tuples might need to be unified first a bit to make this effortless.

src/typechecker.erl
```diff
@@ -942,55 +942,58 @@ type_check_lc(Env, Expr, [{generate, _, Pat, Gen} | Quals]) ->
 
 
 
+type_check_expr_in(Env, ResTy, Expr) ->
+    NormResTy = normalize(ResTy, Env#env.tenv),
+    do_type_check_expr_in(Env, NormResTy, Expr).
```
zuiderkwast: This is good, except the type_error will contain the normalized type. It is better to do it in one place than in all the `*_in` functions. The original type can easily be restored with using the try-catch and re-throw method I sketched earlier (although using get_stacktrace/0 for pre-OTP-21), at least temporarily until we find a better solution. As a rule for this, let's say that type to replace is always the last element in the tuple if the first element is `type_error`.

gomoripeti: ok, I will add another commit with re-throw and assuming last type_error element is the one to replace

gomoripeti: Just noting: keeping the original type has the funny side-effect of error messages like this:
```
The operator 'andalso' on line 7 is given a non-boolean argument of type bool()
```
whereas showing the normalized type better reveals the problem (`type_check_logic_op_in` does not handle union as boolean())
```
The operator 'andalso' on line 7 is given a non-boolean argument of type true | false
```

This can be remedied by printing the whole type stack as you proposed.

src/typechecker.erl
```diff
@@ -1442,7 +1461,8 @@ merge_types(Tys) ->
 add_types_pats([], [], _TEnv, VEnv) ->
     VEnv;
 add_types_pats([Pat | Pats], [Ty | Tys], TEnv, VEnv) ->
-    add_types_pats(Pats, Tys, TEnv, add_type_pat(Pat, Ty, TEnv, VEnv)).
+    NormTy = normalize(Ty, TEnv),
+    add_types_pats(Pats, Tys, TEnv, add_type_pat(Pat, NormTy, TEnv, VEnv)).
```
zuiderkwast: What about the normalization here? Do we need to restore the original type here too?

gomoripeti: ops, forgot about this place, fixed in a new commit

gomoripeti:
I created the macro, because I assume this re-throw will be needed in more places, and it is not possible to do the try-catch wrapping from a helper function.

PR #20, opened 2018-07-13
BEAM files support for gradualizer_db

tim2CF:
- `gradualizer_db` can read forms from `.beam` files
- with `.beam` files support `gradualizer_db` can work with Elixir projects
- `.editorconfig` file for code styles consistency

### DIFF
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -13,8 +13,7 @@
          type_check_module/1,
          type_check_module/2,
          type_check_dir/1,
-         type_check_dir/2,
-         get_forms_from_beam/1
+         type_check_dir/2
         ]).
 
 -type options() :: proplists:proplist().
@@ -32,9 +31,9 @@ type_check_file(File, Opts) ->
     Forms =
         case filename:extension(File) of
             ".erl" ->
-                get_forms_from_erl(File);
+                gradualizer_file_utils:get_forms_from_erl(File);
             ".beam" ->
-                get_forms_from_beam(File);
+                gradualizer_file_utils:get_forms_from_beam(File);
             Ext ->
                 throw({unknown_file_extension, Ext})
         end,
@@ -82,30 +81,4 @@ type_check_dir(Dir, Opts) ->
               end, ok, filelib:wildcard(filename:join(Dir, "*.{erl,beam}")));
         false ->
             throw({dir_not_found, Dir})
-    end.
-
-%% Helper functions
-
-get_forms_from_erl(File) ->
-    case epp:parse_file(File, []) of
-        {ok, Forms} ->
-            Forms;
-        {error, enoent} ->
-            throw({file_not_found, File});
-        {error, Reason} ->
-            throw({file_open_error, {Reason, File}})
-    end.
-
-get_forms_from_beam(File) ->
-    case beam_lib:chunks(File, [abstract_code]) of
-        {ok, {_Module, [{abstract_code, {raw_abstract_v1, Forms}}]}} ->
-            Forms;
-        {ok, {_Module, [{abstract_code,no_abstract_code}]}} ->
-            throw({forms_not_found, File});
-        {error, beam_lib, {file_error, _, enoent}} ->
-            throw({file_not_found, File});
-        {error, beam_lib, {file_error, _, Reason}} ->
-            throw({file_open_error, {Reason, File}});
-        {error, beam_lib, Reason} ->
-            throw({forms_error, Reason})
     end.
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -28,7 +28,7 @@ type_check_file(File) ->
 %% @doc Type check a source or beam file
 -spec type_check_file(file:filename(), options()) -> ok | nok.
 type_check_file(File, Opts) ->
-    Forms =
+    ParsedFile =
         case filename:extension(File) of
             ".erl" ->
                 gradualizer_file_utils:get_forms_from_erl(File);
@@ -37,8 +37,13 @@ type_check_file(File, Opts) ->
             Ext ->
                 throw({unknown_file_extension, Ext})
         end,
-    Opts2 = proplists:expand([{print_file, [{print_file, File}]}], Opts),
-    typechecker:type_check_forms(Forms, Opts2).
+    case ParsedFile of
+        {ok, Forms} ->
+            Opts2 = proplists:expand([{print_file, [{print_file, File}]}], Opts),
+            typechecker:type_check_forms(Forms, Opts2);
+        Error ->
+            throw(Error)
+    end.
 
 
 %% @doc Type check a module
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -13,7 +13,8 @@
          type_check_module/1,
          type_check_module/2,
          type_check_dir/1,
-         type_check_dir/2
+         type_check_dir/2,
+         get_forms_from_beam/1
         ]).
 
 -type options() :: proplists:proplist().
--- a/src/gradualizer_db.erl
+++ b/src/gradualizer_db.erl
@@ -11,6 +11,7 @@
          get_record_type/2,
          get_modules/0, get_types/1,
          save/1, load/1,
+         import_module/1,
          import_erl_files/1, import_beam_files/1,
          import_app/1, import_otp/0]).
 
@@ -21,6 +22,9 @@
 
 %% Types for the Erlang Abstract Format
 -type type() :: erl_parse:abstract_type().
+
+%% Compiled regular expression
+-type regexp() :: {re_pattern, _, _, _, _}.
 
 %% Gen server local registered name
 -define(name, ?MODULE).
@@ -107,6 +111,10 @@ import_app(App) ->
 import_otp() ->
     call(import_otp, infinity).
 
+-spec import_module(module()) -> ok | not_found.
+import_module(Module) ->
+    call({import_module, Module}, infinity).
+
 %% ----------------------------------------------------------------------------
 
 %% Gen_server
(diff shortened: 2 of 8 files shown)


src/gradualizer_db.erl
```diff
@@ -279,17 +280,29 @@ autoimport(M, #state{opts = #{autoimport := true},
...
+              _ ->
+                  not_found
+          end
     end.
 
 import_files([File | Files], State) ->
-    EppOpts = [{includes, guess_include_dirs(File)}],
-    {ok, Forms} = epp:parse_file(File, EppOpts),
+    {ok, Forms} =
+        case re:run(File, beam_file_regexp()) of
```
zuiderkwast: I think it is better to make two separate functions for importing from an .erl file and importing a .beam file. It is less logic and the import function doesn't need to use a regex to find out what kind of file it is importing, which is something we already know where the function is called.

src/gradualizer_db.erl
```diff
@@ -279,17 +280,29 @@ autoimport(M, #state{opts = #{autoimport := true},
...
+            {match, _} ->
+                {ok, gradualizer:get_forms_from_beam(File)};
```
zuiderkwast: What if the beam files doesn't contain the abstract code? That's not `ok`...

src/gradualizer_db.erl
```diff
@@ -279,17 +280,29 @@ autoimport(M, #state{opts = #{autoimport := true},
...
+            nomatch ->
+                EppOpts = [{includes, guess_include_dirs(File)}],
+                epp:parse_file(File, EppOpts)
+        end,
     [{attribute, _, file, _},
      {attribute, _, module, Module} | Forms1] = Forms,
     check_epp_errors(File, Forms1),
```
zuiderkwast: Check_epp_errors should probably be done only when we use epp, i.e. not for beam files.

src/gradualizer_db.erl
```diff
@@ -459,10 +472,30 @@ get_src_map() ->
...
+    BeamDirs = code:get_path(),
+    BeamFiles = lists:flatmap(fun (Dir) -> filelib:wildcard(Dir ++ "/*.beam") end, BeamDirs),
+    BeamPairs = lists:map(fun (Filename) ->
+            {match, [Mod]} = re:run(Filename, beam_file_regexp(), [{capture, all_but_first, list}]),
+            {list_to_atom(Mod), Filename}
+        end,
+        BeamFiles),
+    maps:from_list(BeamPairs).
+
+-spec beam_file_regexp() -> {re_pattern, _, _, _}.
```
zuiderkwast: The return type should be `re:mp()`. From the docs of `re`:

> `mp() = {re_pattern, term(), term(), term(), term()}`
>
>Opaque data type containing a compiled regular expression. mp() is guaranteed to be a tuple() having the atom re_pattern as its first element, to allow for matching in guards. The arity of the tuple or the content of the other fields can change in future Erlang/OTP releases.

src/gradualizer.erl
```diff
@@ -13,7 +13,8 @@
          type_check_module/1,
          type_check_module/2,
          type_check_dir/1,
-         type_check_dir/2
+         type_check_dir/2,
+         get_forms_from_beam/1
```
zuiderkwast: If this function must be exported, I think it should be be moved to another module. In my opinion, it is also OK to duplicate some lines from this function. The case when the beam file doesn't contain abstract code is not an error in this case, but just a case where we should try the fallback instead.

src/gradualizer_db.erl
```diff
@@ -279,17 +280,29 @@ autoimport(M, #state{opts = #{autoimport := true},
...
 
 import_module(Mod, State) ->
-    case State#state.srcmap of
+    case State#state.beammap of
         #{Mod := Filename} ->
             State1 = import_files([Filename], State),
             {ok, State1};
         _ ->
-            not_found
+          case State#state.srcmap of
```
zuiderkwast: Indentation

src/gradualizer_db.erl
```diff
@@ -459,10 +472,30 @@ get_src_map() ->
                    RevPath         -> lists:reverse("lre.*/" ++ RevPath)
                end || Path <- code:get_path()],
     SrcFiles = lists:flatmap(fun filelib:wildcard/1, SrcDirs),
-    {ok, RE} = re:compile(<<"([^/.]*)\.erl$">>),
     Pairs = [begin
-                 {match, [Mod]} = re:run(Filename, RE,
+                 {match, [Mod]} = re:run(Filename, erl_file_regexp(),
```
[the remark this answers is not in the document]
tim2CF: Are there regexp sigils in Erlang? For example in Elixir you can just write `~r/^hello$/` and it will be compiled in compile-time and inlined to code as value

tim2CF:
Hi @zuiderkwast 
All issues fixed except tests
What is proper pattern to test things like this?
Make function `&:gradualizer_db.import_module/2` public (exported) and implement `eunit` test case?

src/gradualizer_db.erl
```diff
@@ -189,8 +189,8 @@ handle_call({load, Filename}, _From, State) ->
                                        types  = maps:merge(Ty1, Ty2),
                                        loaded = maps:merge(Loaded1, Loaded2)},
                 {reply, ok, NewState}
-            catch error:E ->
-                {reply, {error, E, erlang:get_stacktrace()}, State}
+            catch _:Exception:Stacktrace ->
```
gomoripeti: (my opinion) please don't use the new syntax introduced only in Erlang/OTP 21.0. It would be good to keep supporting older Erlang versions (current master still runs fine on OTP 19+). The old syntax is still valid (only results in a warning). 

Let's use this directive proposed by @zuiderkwast (in https://github.com/josefs/Gradualizer/pull/18) in the top of the module to silence this warning for the time being.
```
-ifdef(OTP_RELEASE).
-compile([{nowarn_deprecated_function,{erlang,get_stacktrace,0}}]).
-endif.
```

If we start to need stacktraces in many places we can start using https://hex.pm/packages/stacktrace_compat, but for now I wanted to avoid any external dependency.

(this PR is great work btw 👍)

tim2CF: Thanks, let's use it!
There are 2 other compiler warnings:
```
src/ty.erl:6: Warning: export_all flag enabled - all functions will be exported
src/typechecker.erl:9: Warning: export_all flag enabled - all functions will be exported
```
I think it's nice to make proper `export` here

src/gradualizer_db.erl
```diff
@@ -483,19 +484,20 @@ get_src_map() ->
...
     BeamPairs = lists:map(fun (Filename) ->
-            {match, [Mod]} = re:run(Filename, beam_file_regexp(), [{capture, all_but_first, list}]),
+            {match, [Mod]} = re:run(Filename, RE, [{capture, all_but_first, list}]),
             {list_to_atom(Mod), Filename}
         end,
         BeamFiles),
     maps:from_list(BeamPairs).
 
--spec beam_file_regexp() -> {re_pattern, _, _, _}.
+-spec beam_file_regexp() -> {re_pattern, _, _, _, _}.
```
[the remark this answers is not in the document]
tim2CF: I'm not sure it can be used, I don't see type exporting here
https://github.com/erlang/otp/blob/master/lib/stdlib/src/re.erl

src/gradualizer_file_utils.erl
```diff
@@ -0,0 +1,64 @@
...
+        {error, beam_lib, {file_error, _, Reason}} ->
+            {file_open_error, {Reason, File}};
+        {error, beam_lib, Reason} ->
+            {forms_error, Reason}
+    end.
+
+-spec get_forms_from_erl(file:filename()) -> abstract_form() | no_return().
+get_forms_from_erl(File) ->
+    case get_forms_from_erl_safe(File) of
+        {ok, Forms} ->
```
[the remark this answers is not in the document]
tim2CF: Sure we can
In Elixir it's common pattern to have 2 versions of utility function: safe and unsafe, for example

```
Map.fetch(map(), key()) :: {:ok, value()} | :error

Map.fetch!(map(), key()) :: value() | no_return()
```

So first idea was to make `get_forms_from_erl` and `get_forms_from_erl!` functions, but it looks not very good because `!` is keyword in Erlang. I searched some info about safe/unsafe functions naming conventions in Erlang, but it seems there is not any

tim2CF:
updated

src/gradualizer_db.erl
```diff
@@ -21,6 +27,9 @@
 %% Types for the Erlang Abstract Format
 -type type() :: erl_parse:abstract_type().
 
+%% Compiled regular expression
+-type regexp() :: {re_pattern, _, _, _, _}.
```
zuiderkwast: I couldn't understand why you didn't want to use `re:mp()`, but now I see: It is not exported! So I understand now.

    5> gradualizer_db:get_exported_type(re,mp,[]).    
    not_exported

(That's annoying as from the docs it looks like it is the type to use. Perhaps I will do a PR at erlang/otp and export this type.)

test/gradualizer_db_tests.erl
```diff
@@ -0,0 +1,7 @@
+-module(gradualizer_db_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+
+import_module_test() ->
+    ?assertMatch(not_found, gradualizer_db:import_module(hello)),
+    ?assertMatch(ok, gradualizer_db:import_module(lists)).
```
zuiderkwast: Looks good. I like the test. To test beam files with and without the AST, I guess we would have to add compiling to the test, but that's a bit more work. Maybe in the future then.

zuiderkwast:
Thanks @tim2CF!

Issue #23, opened 2018-07-16
Type refinement in conditional expressions


gomoripeti:
let's see the following correct function with spec
```
-spec f(atom() | integer()) -> true | integer().
f(AI) ->
    is_atom(AI)     %% Expr1
      orelse
        AI + 1.     %% Expr2
```
- the type of variable `AI` in `Exrp2` must be the subtype of `number() :: integer() | float()` as the first argument of `+`
- the type of variable `AI` in `Exrp2` is `atom() | integer()` based on the spec which is not a subtype of `integer() | float()`

How could Gradualizer realise that `Expr2` is only executed when type of `AI` is `integer()`?

- if type checking was supersmart it would realise that `if NOT is_atom(AI) => typeof(AI) == (atom() | integer()) & (not atom()) == integer()` - however this is not always possible

Let's see a more generic example
```
-spec f(input()) -> any().
f(A) ->
  cond(A)
    orelse
      body(A).

-spec cond(input()) -> boolean().
-spec body(required()) -> any().
```
In this case what the type checking could do is realise that `body(A)` is only executed in a subset of cases therefor type of `A` in `body(A)` is just a subset or subtype of `input()`. In case `input() & required() == none()` (the intersection of the two types is empty) then we can say that `body(A)` will always fail, but otherwise we cannot know (and have to leave type checking at runtime). This method requires to be able to calculate the **intersection** of two types.

What is the right way to handle these kind of constructs? (`if`, `case`, `andalso`, `orelse`, multiple function clause)

zuiderkwast:
Just a comment on `orelse` and `andalso`. @josefs regards them as boolean operators where both operands should be booleans. Otherwise it's a type error. I regard them as conditional constructs when only the first operand must be boolean. I think the latter has become common in Erlang code during the last years. There is even a line in this project where it is used like this:

     File =/= undefined andalso io:format(...),

Perhaps we should decide about which semantics are desirable.

gomoripeti:
I also regard them as conditional constructs, as that is how the Erlang language (compiler/VM) handles them. This is also how I implemented checking them in #24, exactly to address the self-gradualizing failure on the line that you highlighted.

Nevertheless the problem domain in this ticket is more generic, how to handle that the type set of a variable is possibly reduced within a conditional clause body relative to its global type.

josefs:
[...]

PR #24, opened 2018-07-17
Refactor type_check_logic_op[_in]

gomoripeti:
Fixes self-check warning:
The operator 'andalso' on line 1733 is given a non-boolean argument of type ok

Type checking can still return false warnings in some scenarios described in https://github.com/josefs/Gradualizer/issues/23

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -775,7 +775,7 @@ type_check_expr(Env, {op, P, 'not', Arg}) ->
     end;
 type_check_expr(Env, {op, P, BoolOp, Arg1, Arg2}) when
       (BoolOp == 'andalso') or (BoolOp == 'and') or
-      (BoolOp == 'orelse')  or (BoolOp == 'or') ->
+      (BoolOp == 'orelse')  or (BoolOp == 'or') or (BoolOp == 'xor') ->
     type_check_logic_op(Env, BoolOp, P, Arg1, Arg2);
 type_check_expr(Env, {op, P, RelOp, Arg1, Arg2}) when
       (RelOp == '=:=') or (RelOp == '==') or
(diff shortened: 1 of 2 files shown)


gomoripeti:
thanks @josefs for the feedback. I opened this PR (and not pushing straight) exactly to get feedback. I will rework accordingly.

So the way forward is that if people use the lazy bool ops with a non-bool in a fully untyped program, they will get a warning. Then they need to add type annotation in the right place to address the warning.
However this type annotation might need to be added to expressions, not just on function top level. Which is a longer-term feature I guess. (the `ty.erl` module is related)

josefs:
@gomoripeti , I'm not quite sure what you mean by "if people use the lazy bool ops with a non-bool in a fully untyped program, they will get a warning". Right now, if a programmer doesn't specify any types, then there will be no errors or warnings when e.g. passing a non-boolean second argument to `andalso`. It works just fine!

It is when adding type annotations that the programmer has to be careful, if s/he wants to keep using a non-boolean argument then there needs to be an `any()` type at some appropriate place to make sure the gradualizer doesn't complain. But I think you understand this part already.

gomoripeti:
Actually in some cases there is a type error even without type annotation, eg:
```
f() -> false andalso {true}.

=> The operator 'andalso' on line 1 is given a non-boolean argument of type {any()}
```
It depends on whether `type_check_expr` infers some type for the expression or just returns `any()` (in most of the cases the later)

A more practical example is when the second argument is a library function which happens to be typed within OTP, (this case is not exactly non-typed, but the type annotation of the called function is out of the control of the developer) eg:
```
f() -> true andalso atom_to_list(asd).

=> The operator 'andalso' on line 1 is given a non-boolean argument of type string()
```

Can these examples be considered bugs, so the correct behaviour should be as you described?

josefs:
The first example is clearly a bug and I'm surprised that the gradualizer complains here. Good find!

As for the second example, it's behaving as intended. The function `f/0` may not have a type spec, but `atom_to_list/1` does, so that type information should be propagated. I was not very careful when writing "if a programmer doesn't specify any types". I meant something more along these lines: if there are no type specs in the program, or in any of the library functions that are called, then there will be no errors or warnings.

zuiderkwast:
It is not true then that any untyped program can start using gradualizer without getting any type errors, and then gradually add specs. Most OTP library functions are speced already.

It seems a bit weird that the above examples give type errors but this one doesn't:

    f() -> foo andalso banana.

I think that not propagating any types, not even the return types from spec'ed functions, would be more intuitive for `type_check_expr` if we want to stick to "if a programmer doesn't specify any types". If we do propagate types, then why not also propagate the type of literals? Library functions are outside the programmer's control, just like literals and other language constructs.

josefs:
@zuiderkwast you make a good point about OTP being annotated. I think we should consider having a flag to gradualizer which can effectively turn off type annotations in OTP. That will allow programmers to start from a clean slate with absolutely no type annotations in their (possibly already existing) projects and then gradually add type information.

zuiderkwast:
Such a flag @josef, would it turn off type annotation only in OTP or also in 3rd party libraries? I think that most projects use 3rd party libraries and that these most often are annotated too. 

Would the type annotations be turned off only in `type_check_expr` but still use them in `type_check_expr_in`? I think that would probably be good, but I wouldn't mind some more reasoning about it.

Let's continue discussion in #26?

josefs:
@zuiderkwast, yes, I think it'd make sense to silence the type specs in 3rd party libraries as well. But that could perhaps be a third party flag.

As for where the type specs would be silenced, I'd imagine it would be everywhere. I think it would be rather hard to explain the effects of having the type spec only be turned off in `type_check_expr`.

gomoripeti:
As concluded in #23 and also on erlang mailing list short-circuit operators should not support any term() type as second argument. It is only an implementation detail that there is no check in the VM for the type of the second arg (which is an optimisation to allow tail recursion) and it is considered a bad practice to exploit this.

Therefore closing this PR and will submit a new one which fixes warnings like:
`The operator 'andalso' on line ... is given a non-boolean argument of type true | false`

Issue #26, opened 2018-07-22
Role of `type_check_expr`


gomoripeti:
In https://github.com/josefs/Gradualizer/pull/24#issuecomment-406861255` was mentioned that  `type_check_expr` should not return type in any case (currently it return a type other than any if expr is fun objection, function call, expr with clauses (try, case, receive etc), record or tuple).
The general behaviour of `type_check_expr` is that it calls itself on subtypes of the expr and (maybe) type-checks the returned types. If the returned type is always any then what can this function check?

On a different note even `type_check_expr_in` calls `type_check_expr` to infer some type (eg in `case Cond of Pattern -> ...` `Pattern` is checked against the inferred type of `Cond`). Type checking could be more powerful if more precise type would be inferred for more expressions (eg a float literal could be inferred to be of type `float()`)

But maybe the two functionalities could be split:
- `type_check_expr` checks the type of expression in case there is no type information available (basically same as checking against the any() type - checking a dynamic program)
- `infer_type_of_expr` would infer the type of the expression (this would have obviously limited scope and wouldn't go beyond the boundaries of function calls)

But I might misunderstand the roles of the functions (I ignored collecting constraints). What am I missing?

josefs:
Where did I write anything along the lines of "type_check_expr should not return type in any case"? If I did then I was clearly delusional. 
The function `type_check_expr` has a very specific role. It tries its best to propagate the return type of functions which are called within an expression. Sometime `type_check_expr` **does** return a type which is different from `any()` and in those cases the type can always be traced back to a call to a type spec'ed function.

In general, the split between the functions `type_check_expr` and `type_check_expr_in` follows the principle of **bidirectional typing**, which, among other things, maximizes the propagation of type information. There are plenty of papers on this subject if you're interested.
That being said, the names of the functions are pretty bad. `type_check_expr` really should be called something akin to `infer_expr` and `type_check_expr_in` should be something like `type_check_expr`.

gomoripeti:
(it was actually Victor who mentioned return-no-type)

I see, so type information must always stem from specs (of functions and also records, if I see correctly) and type should be inferred from no other untyped construct.
(This is a little heureca moment for me and explains a few places in the code I didn't understand -why we fall back to `any()` so eagerly. I had the assumption that we should use the most type info we can sqeeze from an expression, eg from literals - but this is not the case.)

Is the reason for this (not to infer type from other constructs) is to maximize the power/effect of user intent?

[...]

PR #27, opened 2018-07-23
Only infer tuple type if it has a subexpr typed by the user 

gomoripeti:
As discussed in https://github.com/josefs/Gradualizer/issues/26

I am a bit hesitant if I did the correct thing. I implemented along the lines: "tries its best to propagate the return type of functions which are called within an expression" that is if ***any*** of the elements of the tuple has a non-any type then the tuple will also have non-any type inferred.

However the intended behaviour might be if ***all*** elements of the tuple have non-any type...
The second behaviour is visible in the implementation eg. for lists (cons) where currently both the head and the tail must be non-any to infer a non-any type for the cons cell.
Also `not Expr` will always be inferred as `any()` even `not f()` where `f/0` is spec'ed.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1425,18 +1425,18 @@ check_guards(Env, Guards) ->
 				    end, GuardSeq))
 		end, Guards)).
 
-type_check_function(FEnv, TEnv, {function,_, Name, NArgs, Clauses}) ->
-    case maps:find({Name, NArgs}, FEnv) of
+type_check_function(Env, {function,_, Name, NArgs, Clauses}) ->
+    case maps:find({Name, NArgs}, Env#env.fenv) of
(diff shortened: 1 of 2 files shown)


src/typechecker.erl
```diff
@@ -601,6 +601,13 @@ int_range_to_types({I, J}) when I < J ->
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+%% @doc Infer type of expression
+%%
+%% Type information should always stem from user type annotation (function specs
+%% and type annotated record definitions). If an expression does not have a
+%% subexression that has a type inferred from these sources, its inferred type
+%% will be `any()'.
```
josefs: "subexression" => "subexpression"

PR #28, opened 2018-08-22
Short-circuit logic ops propagate variable bindings from their first arg to their second one.

gomoripeti:
Also
- use `subtype` function when type-checking logic op (that is more generic and future proof)
- support type checking matches and blocks

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1244,6 +1244,9 @@ do_type_check_expr_in(Env, Ty, {var, LINE, Var}) ->
 	false ->
 	    throw({type_error, tyVar, LINE, Var, VarTy, Ty})
     end;
+do_type_check_expr_in(Env, Ty, {match, _, Pat, Expr}) ->
+    {VarBinds, Cs} = type_check_expr_in(Env, Ty, Expr),
+    {add_type_pat(Pat, Ty, Env#env.tenv, VarBinds), Cs};
 do_type_check_expr_in(Env, Ty, I = {integer, LINE, Int}) ->
     case subtype(I, Ty, Env#env.tenv) of
 	{true, Cs} ->
@@ -1529,6 +1532,8 @@ do_type_check_expr_in(Env, ResTy, {op, P, Op, Arg1, Arg2}) when
       Op == '++' orelse Op == '--' ->
     type_check_list_op_in(Env, ResTy, Op, P, Arg1, Arg2);
 
+do_type_check_expr_in(Env, ResTy, {block, _, Block}) ->
+    type_check_block_in(Env, ResTy, Block);
 do_type_check_expr_in(Env, ResTy, {'catch', _, Arg}) ->
     % TODO: Should we require ResTy to also include the possibility of
     % exceptions? But exceptions can be of any type! That would mean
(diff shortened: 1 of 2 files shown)


src/typechecker.erl
```diff
@@ -1385,7 +1397,7 @@ type_check_block_in(Env, ResTy, [Expr]) ->
 type_check_block_in(Env, ResTy, [Expr | Exprs]) ->
     {_, VarBinds, Cs1} = type_check_expr(Env, Expr),
     {VB, Cs2} = type_check_block_in(Env#env{ venv = add_var_binds(Env#env.venv, VarBinds) }, ResTy, Exprs),
-    {VB, constraints:combine(Cs1, Cs2)}.
+    {union_var_binds([Env#env.venv, VarBinds, VB]), constraints:combine(Cs1, Cs2)}.
```
gomoripeti: I'm not sure which variable bindings `type_check_expr[_in]` and all similar functions should return
- only new ones bound in the expression checked
- a cumulated set that also includes variable bindings from the input `Env`

josefs: It should only return new one bound in the expression being checked. That means, `Env#env.venv` should not be in there.

You've clearly identified a bug here! The `VarBinds` clearly needs to be returned as well, just as you have written. Good job!

test/should_pass/bool.erl
```diff
@@ -5,3 +5,16 @@
...
+b2() ->
+    begin
+        A = f(),
+        is_integer(A)
+    end
+        orelse
+        is_float(A).
+
+-spec f() -> number().
+f() -> 1.
```
josefs: This test is great and reminds me we don't have that many tests checking that we get the scoping of variables correctly. Erlangs scoping is rather weird and inconsistent, and I'm sure there are places (just like this one) where I've gotten it wrong. This test is a good start. One could also imagine a test for `or` that makes sure that we *don't* propagate the variable bindings across the operator.

gomoripeti: Yes, Erlang scoping is sometimes surprising, and it's hard to come up with thorough testing. (Unfortunately I don't have time to write extensive tests for Gradualizer, instead when I hit an issue, I try to add at least one example testcase which fails without the fix, regression testing style) I think Gradualizer has scoping right in most of the cases, I didnt find any issue on real code-bases. This testcase is a bit contrived, I only found it based on your comment in the code. But I will quickly scan the code if I see any other suspicious places.

josefs: Sorry, I didn't mean to imply that you had to write these test for this pull-request. I was merely thinking out loud about the lack of tests for the scoping rules. No need to write any additional tests. I think your approach of adding regressions as you find them is fine.

josefs: Sorry, I didn't mean to imply that you had to write new test as part of this pull-request. I was merely thinking out loud about our lack of tests for scoping rules. I agree with you that we have it mostly right, and that's good enough for now. I also agree that adding regressions as you find them is a good way forward.

zuiderkwast:
Looks good. I don't have any comments. Submit?

gomoripeti:
finally had the time to resolve conflict with master

PR #30, opened 2018-09-21
Improve type checking if/case/try expressions

gomoripeti:


gomoripeti:
these changes were sitting in my WIP branch since a while, it's good that they don't conflict with your ongoing work.

Issue #29, opened 2018-09-20
Making it a real Erlang application


Zalastax:
I want to work on making Gradualizer a real Erlang application so I'm wondering what you have in mind.
Is it a traditional long-running OTP application you have in mind or a CLI application that starts and stops between each request?
Perhaps an implementation using the [Language Server Protocol](https://langserver.org/)?

zuiderkwast:
I assume a CLI tool is what Josef had in mind.

If you want to make an LSP "plugin" for Vlad's [sourcer](https://github.com/erlang/sourcer), it would be pretty nice I guess. I don't know if there is a plugin mechanism of any sort. It might be some more work...

I would like to see plugins for rebar and erlang.mk as well so that you can run `rebar3 gradualize` and `make gradualize`.

It's all up to you!

gomoripeti:
Just adding to the brainstorming, Gradualizer could also be inserted as a parse transform, which does not actually transform the AST but emits errors in case of a type error. This way it could be part of compiling a module. (In a very futuristic setup it could even modify the AST applying small optimisations based on extra type information it gather during analysis - but I don't think this is the primary goal.)

But before that, I agree, it's better to keep Gradualizer as a standalone step and trigger it via CLI/build tool plugins.

Zalastax:
Regarding integration with rebar3: it seems like a plugin needs to be [its own OTP application](https://www.rebar3.org/docs/plugins#section-tutorial) (eventual sourcer integration probably needs this as well). I assume that means gradualizer would need to be an [umbrella project](https://www.rebar3.org/docs/basic-usage#section-new-app-or-release)? Is that fine with you @josefs and can erlang.mk support that?

josefs:
If making Gradualizer an umbrella project is what it takes to integrate with rebar3, I'm all for it.

As for whether erlang.mk can support umbrella projects, I don't know. @gomoripeti, do you know?

gomoripeti:
I'm not a big fan of library application as an umbrella. For example rebar3 does not support an umbrella app as a dependency. Also creating a hex package with the old rebar3 hex plugin from apps in an umbrella is problematic.

OTOH Gradualizer is already an OTP application, wouldn't just adding a ` gradualizer_prv` module do the trick? (as a rebar3 plugin - I don't know how erlang.mk plugins work)

Zalastax:
[...]

PR #31, opened 2018-09-23
Add a command line interface

Zalastax:
See #29

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,12 @@
 # start a shell
 # > make shell
 #
+# build cli
+# > make escript
+#
+# run cli
+# > gradualizer
+#
 # run unit tests
 # > make eunit
 #
@@ -24,6 +30,9 @@ PROJECT = gradualizer
 
 PLT_APPS = kernel stdlib compiler crypto
 DIALYZER_OPTS = -Werror_handling
+
+ESCRIPT_NAME = gradualizer_cli
+ESCRIPT_FILE = $(PROJECT)
 
 include erlang.mk
 
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -16,6 +16,8 @@
          type_check_dir/2
         ]).
 
+-export_type([options/0]).
+
 -type options() :: proplists:proplist().
 
 %% API functions
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ gradualizer.d
 cover/
 eunit.coverdata
 test/ct.cover.spec
+/gradualizer
--- a/README.md
+++ b/README.md
@@ -11,7 +11,17 @@ This tool is a type checker for Erlang.
 
 # Usage
 
-Right now the user experience is not very polished. Here's how to get started:
+To run gradualizer as from the command line:
+
+* Compile the project as an escript:
+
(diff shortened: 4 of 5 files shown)


josefs:
Looks good! I agree with Peter's comments. I just want to add that you don't have to add the `--pa` for this PR to be accepted. That can be added later.

src/cli.erl
```diff
@@ -0,0 +1,69 @@
+-module(cli).
+-export([main/1]).
+-include("gradualizer.hrl").
+
+-spec main([string()]) -> ok | no_return().
```
zuiderkwast: `ok | no_return() = ok`

src/gradualizer.erl
```diff
@@ -16,8 +16,7 @@
          type_check_dir/2
         ]).
 
--type options() :: proplists:proplist().
-
+-include("gradualizer.hrl").
```
zuiderkwast: Use `-export_type` if we need the type in other modules. Include files are not needed for this purpose.

src/cli.erl
```diff
@@ -0,0 +1,69 @@
...
+            end, {true, 0}, Rest)
+    end,
+    halt(Exit).
+ 
+-spec get_ver(atom()) -> string().
+get_ver(App) ->
+    case lists:keyfind(App, 1, application:loaded_applications()) of
+        {_, _, Ver} ->
+            Ver;
+        false ->
```
zuiderkwast: Why accept the false case here and hide the error in an empty string? Since we require the application to exist, it's better to crash IMO.

And why a generic get_ver when we only care about the gradualizer application? I would shorten all this to `{_, _, Ver} = lists:keyfind(gradualizer, 1, application:loaded_applications())`.

Makefile
```diff
@@ -25,6 +31,9 @@ PROJECT = gradualizer
 PLT_APPS = kernel stdlib compiler crypto
 DIALYZER_OPTS = -Werror_handling
 
+ESCRIPT_NAME = cli
```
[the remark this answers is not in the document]
Zalastax: cli just means that it should use the cli.erl file. Is there a possibility that it would name clash with a cli.erl in a future dependency?

src/cli.erl
Zalastax: Ah, I'll move the non-termination condition to a normal comment instead.

src/cli.erl
Zalastax: Makes sense. A generic function was used because I found the example online. It's also easier for me to filter out the noise that way - inlining the function makes print_version/0 read less like prose. I've been getting the same feedback from my team and I'm not a convert yet.

Zalastax:
Take a look again!
Adding the `-pa` was quite important to even get the `should_pass` tests to pass.
I would have liked if the help text would wrap based on the terminal width but I couldn't get `io:columns()` to work.

josefs:
Excellet work @Zalastax! @zuiderkwast, are you happy with the changes?

zuiderkwast:
Sure! Merge it!

josefs:
Welcome as a contributer @Zalastax!

Zalastax:
Thanks! There's more to come 😄

PR #36, opened 2018-09-30
Create a rebar3 plugin

Zalastax:
Tested in both a standalone app structure and umbrella structure.
The following needs fixing but I've never done that so leaving that to someone else (I allow edits from maintainers):
```Erlang
% TODO: Include rebar.hrl and remove the following type definition
% -include_lib("rebar/src/rebar.hrl").
-type rebar_dict() :: dict:dict().
```

Some files had CRLF line endings and are now LF.


josefs:
This looks good to me! Though I have to ask; the diffs that github shows me looks a little weird. E.g. for the README file, it shows the diff as if all the content was removed and exchanged for something different, which is clearly not what has been done. @Zalastax, do you know what's going on with the diffs?

@zuiderkwast and @gomoripeti, would you like to speak up about his PR? I don't have any experience with creating rebar3 plugins so I figure someone who's a little more qualified might take a look.

Zalastax:
@josefs I split the PR into two commits: one with the actual changes and a second with just whitespace changes. Hope that clears up the confusion!

examples/rebar3/README.md
```diff
@@ -0,0 +1,32 @@
...
+{plugins, [
+  {gradualizer, {git, "git://github.com/josefs/Gradualizer.git", {branch, "master"}}}
+]}.
+```
+
+# Options
+
+Gradualizer checks every source file in your app(s), unless configured not to.  
+Configuration is read from `gradualizer_opts` in `rebar.config` which
+should be a `proplists:proplist()`
```
gomoripeti: maybe add `The following options are supported:`

gomoripeti:
In the "Files changed" tab there is a "Diff settings" with "Hide whitespace changes".

But when using eg git blame retrospectively it's definitely better to have the whitespace changes in a separate commit.

josefs:
I thought I had tried to hide the whitespace changes, but apparently not. Anyway, thanks for clearing that up and separating the whitespace changes into a separate commit.

zuiderkwast:
Looks really good to me! I've never made rebar plugins myself either. I agree with Péter's comments, e.g. using `dict:dict()` directly instead of `rebar_dict()`.

src/rebar_prv_gradualizer.erl
```diff
@@ -0,0 +1,94 @@
...
+            []
+    end.
+
+-spec format_error(any()) -> iolist().
+format_error(_) ->
+    "Gradualizer found errors.".
+
+-spec resolve_src_dirs(rebar_dict()) -> {[file:name()], [file:name()]}.
+resolve_src_dirs(Opts) ->
+    SrcDirs = rebar_dir:src_dirs(Opts, ["src"]),
```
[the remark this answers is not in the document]
Zalastax: Thanks! I has to dig a bit so I opened [an issue](https://github.com/erlang/rebar3/issues/1900) in rebar3 about improving the documentation.

Zalastax:
I'm happy now! Take it for a spin and merge when confident.

examples/rebar3/src/rebar3.erl
```diff
@@ -0,0 +1,7 @@
+-module(rebar3).
```
zuiderkwast: Slightly confusing example module name. It may be mistaken for the tool with the same name. How about my_example or something like that?

examples/rebar3/src/rebar3.app.src
```diff
@@ -0,0 +1,11 @@
+{application, rebar3,
```
zuiderkwast: Slightly confusing name of the example project.

examples/rebar3/src/rebar3.erl
Zalastax: rebar3_example

gomoripeti:
I quickly tried this branch on an (umbrella) project.
- I noticed that in case of an umbrella the `CWD` seems to be the dir of each contained application and not the root of the repo. So an include pattern `"src/mod.erl"` works but none of `"mod.erl"`, `"*mod.erl"` or `"apps/myapp/src/mod.erl"`. What do you think, would it be more handy to use regexp pattern instead of `filelib:wildcard` pattern for include and exclude? This way `.*` would match the whole path string and not just a path component, and also don't have to know what is the `CWD`, just write `"mod.erl"` or `"mod_.*.erl"`.

- What I see is gradualizer(_db) cannot find function specs and exported types in other modules in the same application. This is strange as I thought `rebar3` takes care of setting up code path and then `gradualizer_db` would be able to fetch types and specs from other modules. I did not dig into why this does not work.

Zalastax:
@gomoripeti my vision is that apps in umbrella projects that modify those patterns would have individual `rebar.config` files and that exclude/include should be relative to the application directory. However, I think it would be nice to allow for some merging strategy between the top level config and the app configs but perhaps duplication can be avoided through some other means. I'm personally happy with the way it works right now.

I believe I tested your second point and had it working - so I'm surprised. Did you include the other app as a dependency?

Zalastax:
@gomoripeti I looked at [what dialyzer is doing](https://github.com/erlang/rebar3/blob/416176290b20e1e68c5901f83cccef71ec2bc322/src/rebar_prv_dialyzer.erl#L88) and adapted the same strategy.

I have two concerns at the moment:
* Is there a problem with adding all dependencies/sibling applications to the code path? What if one fails? What if this application uses a dependency that is actually specified in a sibling (can they have different versions?) (what if that sibling is not included in the release?)?
* The CLI does not automatically add to the path, perhaps it should?

Only the first problem concerns this issue so if you think this last commit solves the issue - merge

gomoripeti:
@Zalastax your reasoning with include/exclude makes sense 👍 

- I think rebar3 should do the dependency/version resolution so we shouldn't be concerned about conflicting versions. Also I think (opinion) what is included in the release and what not is out of scope of the gradualizer plugin.
- the CLI does not have the context and the notion of the whole project as rebar3 so I think it's fine that it does not add anything to the path automatically.

josefs:
I'm really excited by this PR! Excellent work @Zalastax!

PR #37, opened 2018-10-03
Fix dialyzer warnings

gomoripeti:
- `expect_fun_type_union` can never return a `type_error` tuple
- output type of `expect_fun_type` can never match the input type
- `type_check_call_ty_union` must return a result type too
 
@josefs in the last case I just highlighted the problem, but actually I don't know how to calculate the result type of a `call_ty_union`, so this PR is incomplete in this state. If you could guide me here, that would be great.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -801,7 +801,7 @@ expect_fun_type(Tys) when is_list(Tys) ->
     end;
 expect_fun_type({type, _, union, UnionTys} = Union) ->
     case expect_fun_type_union(UnionTys) of
-	{type_error, _} ->
+	[] ->
 	    {type_error, Union};
 	[Ty] ->
 	    Ty;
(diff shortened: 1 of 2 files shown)


src/typechecker.erl
```diff
@@ -1272,12 +1272,18 @@ type_check_call_ty_intersect(Env, [Ty | Tys], Args, E) ->
     end.
 
 type_check_call_ty_union(_Env, [], _Args, _E) ->
```
josefs: Ugh. I've clearly written some nonsense code here.

Anyway.

The type that we want to return from type_check_call_ty_union/4 is the union of all the types we get from the calls to type_check_call_ty/4. I think it'd be best to structure type_check_call_ty_union/4 as follows: add an accumulating parameter which is a list of all the types we get from type_check_call_ty/4. When we come to the base case we simply take the list of types and turn it into a union type. I think it'd be a good idea to normalize the resulting union type also.

With this structure you don't have to create the new function calc_union_res_ty/2, and you don't have to return something convoluted in the base case.

Can you make sense of my explanation?

gomoripeti: thanks, it makes sense, I was not sure if `type_check_call_ty` returns a normal `type()` or something fun specific (like these `fun_ty*` tuples). I will update this PR as you suggested

josefs:
Sorry for being slow to react here. I had somehow missed that you provided fixes so quickly.

PR #44, opened 2018-10-08
More fixes to fun_ty_union

gomoripeti:
Changes in PR #37 became obsolete when merged as commit 6085414 added
constraints to the `fun_ty_union` tuple.

There were no conflicts so this was not obvious.


gomoripeti:
thanks Josef, I think this was a classic race condition, only visible after merging.

Issue #49, opened 2018-10-11
Unified error terms and text


gomoripeti:
It would be good to have an option to return errors as erlang terms (when called from Erlang, not cli). 
- various frontends can format the errors according to their taste (eg Elixir syntax formatting).
- also testing negative cases could be more precise

This requires unifying the error terms, each one having the same structure (common fields in same position: eg filename, line number, expected type etc.)

Also the default formatter (used by the cli) could have a format similar to other tools, like the compiler: `file-name:line-number:error-text`

zuiderkwast:
For an type error of an expression E not matching a type T, I suggest the simple structure `{type_error, E, T}` where E and T are abstract syntax. Line number can be extracted from E. This would cover many of the currently different type errors.

Zalastax:
@josefs can we get your blessing to make this happen? I expect that I can dish out a PR addressing this issue before the week ends if we can agree on the format.

zuiderkwast:
I can answer yes on this question. Almost certain that Josef doesn't mind if you sort this out!

josefs:
@Zalastax, go wild! I'd love to see this happen.

Issue #48, opened 2018-10-11
Vim integration through Neomake


erszcz:
I've started work on integrating Gradualizer with Vim.

![Vim/Neomake showing Gradualizer hints](https://s3.amazonaws.com/uploads.hipchat.com/15025/1268738/6Ka3fYr7X0LbRCQ/Screen%20Shot%202018-10-11%20at%2014.45.11.png)

A Neomake maker runs the `gradualizer` EScript to obtain results for a file open in the current Vim buffer. The code path is extended with per-project `ebin` dirs, so that the analysis results are relatively accurate.

For now, the changes are available below - it still requires some polishing, so I'm not creating PRs yet:
- https://github.com/erszcz/Gradualizer/tree/vim-integration
- https://github.com/erszcz/neomake/tree/erlang-gradualizer

One thing that popped up in Gradualizer output is that line numbers are not placed consistently - I had to add a shell script to reformat them - that's the only change on the `vim-integration` branch above.

@josefs do you think this could be addressed directly in Gradualizer (likely `gradualizer_cli.erl`)? A format like this would be ideal (with _on line 102_ being redundant, but it doesn't hurt):

```
src/docsh_lib.erl:102: The type beam_lib:beam/0 on line 102 is not exported
src/docsh_lib.erl:114: The tuple on line 114 does not have type docsh_beam:debug_info() | false
```

Zalastax:
Great initiative!
I think it would be even better if we supported gathering up the errors as a list- then you could use that directly instead of parsing the errors as text.

By the way, does vim work with sourcer? I would love it if we could get great support in many editors and I think using a language server is the best way of doing so.

erszcz:
Thanks!

> I also think it would be even better if we supported gathering up the errors as a list

Indeed, that would be convenient for integration on the Erlang level.

> By the way, does vim work with sourcer?

Judging by erlang/sourcer#19 it doesn't :/
Integration directly with Neomake is pretty straightforward - enough that I can pull it off singlehandedly in a reasonable amount of time.
While I understand the benefits of a language server, I also have the impression that the learning curve might be a lot steeper (not just the server structure / extension points, but possibly also async jobs on Vim side - I'm a long time user, but a Vimscript newbie).

josefs:
This is just wonderful, @erszcz! I think it'd be great if we can add support for this kind of usage. I have spent very little energy and brain cycles on the error messages so I welcome if someone were to make something more useful out of them.

[...]

PR #55, opened 2018-10-25
Add gradualizer.sh to normalise line number position

erszcz:
Following #48 this PR adds a script to reformat Gradualizer messages for easier consumption (until #49 is addressed). The NeoMake part is almost ready and waiting for CI/review at https://github.com/neomake/neomake/pull/2115.

Once Gradualizer itself changes the message format it will be possible to drop this script and adjust the NeoMake maker to just use `gradualizer` escript directly.

Of course, in order for NeoMake to find `gradualizer.sh` the script has to be available in PATH.

### DIFF
--- a/gradualizer.sh
+++ b/gradualizer.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+#set -x
+
+pushd `dirname $0` > /dev/null 2>&1
+BASEDIR=`pwd`
+popd > /dev/null 2>&1
+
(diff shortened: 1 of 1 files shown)


Zalastax:
Would it be possible to maintain this script in your vim plugin for a while? I hope to have #49 adressed very soon.

josefs:
I agree with @Zalastax here. Especially since he's volunteered to fix the underlying problem himself :-)

erszcz:
Fine with me, I've added the file to NeoMake's Erlang maker.

PR #57, opened 2018-10-25
Fix infering type of bool/numeric negation

gomoripeti:
Arg of not should be sutype of `boolean()` and not the other way
around. Same goes for arg of '-', which should be a subtype of
`number()`.

At the same time negation of literal bool/int and range types is
improved. (Maybe this should only work in case `--infer` is true and
otherwise fall back to eg. -(1) => integer()?)

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -592,6 +592,13 @@ upper_bound_more_or_eq({_, A}, {_, B}) ->
 int_max(A, B) when A == pos_inf; B == pos_inf   -> pos_inf;
 int_max(A, B) when is_integer(A), is_integer(B) -> max(A, B).
 
+int_negate(pos_inf) ->
+    neg_inf;
+int_negate(neg_inf) ->
+    pos_inf;
+int_negate(I) when is_integer(I) ->
+    -I.
+
 %% Integer type to range, where the bounds can be infinities in some cases.
 -spec int_type_to_range(type()) -> int_range().
 int_type_to_range({type, _, integer, []})              -> {neg_inf, pos_inf};
@@ -601,23 +608,7 @@ int_type_to_range({type, _, pos_integer, []})          -> {1, pos_inf};
 int_type_to_range({type, _, range, [{integer, _, I1},
                                     {integer, _, I2}]})
                                         when I1 =< I2  -> {I1, I2};
-int_type_to_range({type, _, range, _} = Range) ->
-    %% lower and upper bounds might have operators
-    case normalize(Range, #tenv{}) of
-        Range ->
-            error({not_int_type, Range});
-        NormRange ->
-            int_type_to_range(NormRange)
-    end;
-int_type_to_range({integer, _, I})                     -> {I, I};
-int_type_to_range(Int) ->
-    %% integer might have operators
-    case normalize(Int, #tenv{}) of
-        Int ->
-            error({not_int_type, Int});
-        NormInt ->
-            int_type_to_range(NormInt)
-    end.
+int_type_to_range({integer, _, I})                     -> {I, I}.
 
 %% Converts a range back to a type. Creates two types in some cases.
 -spec int_range_to_types(int_range()) -> [type()].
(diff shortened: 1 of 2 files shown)


src/typechecker.erl
```diff
@@ -600,7 +600,23 @@ int_type_to_range({type, _, pos_integer, []})          -> {1, pos_inf};
 int_type_to_range({type, _, range, [{integer, _, I1},
                                     {integer, _, I2}]})
                                         when I1 =< I2  -> {I1, I2};
-int_type_to_range({integer, _, I})                     -> {I, I}.
+int_type_to_range({type, _, range, _} = Range) ->
+    %% lower and upper bounds might have operators
+    case normalize(Range, #tenv{}) of
```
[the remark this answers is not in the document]
gomoripeti: good suggestion, I was also thinking that this normalization here (with an empty tenv) is not the best thing

zuiderkwast:
> (Maybe this should only work in case --infer is true and
> otherwise fall back to eg. -(1) => integer()?)

Without --infer, we should get -(1) => any() I guess. Now infer is merged so it's possible to implement this difference.

gomoripeti:
I will move the normalization up as suggested (please be patient)

I agree with the need for a general rule for normalization. I earlier had an idea to normalize the whole form before starting type checking, and store the original ast/type in the annotations. (So errors can print the original type and not the normalized one which might look strange) However `erl_anno` is very strict about what can be stored in the annotation and does not allow just any key-value. Maybe some info could be stored in the text field...but this is not ideal.

src/typechecker.erl
```diff
@@ -653,18 +644,18 @@ negate_num_type({type, _, TyName, []} = Ty) when
...
+    %% some kind of range type like `1..3' or `neg_integer()'
+    {L, U} = int_type_to_range(RangeTy),
+    L2 = int_negate(U),
+    U2 = int_negate(L),
+    case int_range_to_types({L2, U2}) of
+        [Ty] ->
+            Ty;
+        Tys = [_, _] ->
+            %% in some cases the result is two mutually exclusive type
+            {type, erl_anno:new(0), union, Tys}
```
zuiderkwast: The only case I can think of where a non-union results in a union is the negation of `non_neg_integer()` => `neg_integer() | 0`. Are there any other cases?

How about negating a union of int types such as `neg_integer() | 0` or `1..3 | 5..7`? That would be possible as well, just negate each term in the union and normalize again. That's a future improvement perhaps.

gomoripeti: you are right about non-union=>union.

and good point about union of integer types. actually they currently crash, so I'll try to address that case too.

gomoripeti:
I realise that calculating type of negation is just a special case of operations on types. Then we could also calculate/infer `constant` `op` `int_type` or even `int_type` `op` `int_type`, but how far can this go. This has for sure it's limits, what are the theoretical ones and where should be the practical ones?

Zalastax:
@gomoripeti Let's leave it as barebones as we possibly can for now. Taking it to its logical conclusion I think we'll either reach full partial evaluation or some ad-hoc dependent types - interesting topics, but hairy. TS and Flow have explored this space a fair bit so we should look there (among others) when ready.

zuiderkwast:
I don't see a problem with evaluating expressions in the type language. It is not the same thing as partial evaluation of expressions (or inlining helper functions). So, as long as it is strait-forward and tested and someone wants to do it, I'm positive about it. (Perhaps we should consider splitting the typechecker module soon though. It's pretty large already.)

Zalastax:
I'm positive as well but I want it to be well thought out and fit with the vision. Go for the low hanging fruit but no further. I know several type systems that accidentally implement turing completeness/lambda calculus so my wish is that we tread carefully.

josefs:
I expect there will be users who will want to have very fine-grained tracking of numerical types, if only because it is possible. I have implemented similar analyses myself in the past and it gets very complicated very quickly with diminishing returns. So I agree with @Zalastax in that I don't mind that we add these operators for now. But there will come a point where it's not going to be worth adding more complicated operations. 

The question is where to draw the line.

My main concern is having a story that is easy to explain to the user. I could imagine supporting the following story: we have precise operations on singleton types but whenever the types become more complicated, we give up trying to be precise and give a more conservative type. For example, suppose that `X :: 4` and `Y :: 2`, then we would have `X * Y :: 8`. However, if, instead, `X :: 4 | 5` then `X * Y :: positive_integer()` (or possibly just `integer()`).

gomoripeti:
I totally agree. The hard part is being consistent- eg gradualizer could handle all integer types as integer(). Or one step further all singleton types as single values and all others as integer(). But gradualizer already handles ranges in some places, and my heart would bleed if that would be removed for consistency :)

My concern with being conservative in case of ranges is this:
```
-spec f(X :: 4 | 5, Y :: 2) -> any().
f(X, Y) ->
  V = X * Y,
  g(V).

-spec g(8 | 10) -> any().
```
In this case we would get a type error as `integer()` (the inferred type of V) is not a subtype of `8 | 10`.

So what we would need to infer is V is any subtype of `integer()` (something like `any() & integer()` where `&` is intersection)

josefs:
We should definitely try to avoid heartbleed :-)

I'm not necessarily against computing precise types for unions of singletons, as in your example @gomoripeti. The main worry I have is that the unions can grow very large. For example:
`X :: 1 | 2`, `Y :: 3 | 5` and `Z :: 6 | 9`. The expression `X * Y * Z` will have type `18 | 27 | 30 | 36 | 45 | 54 | 60 | 90`. In the worst case we can get an exponential number of singletons in the range, with the exponent being the number or arithmetic operations. Granted, most erlang programs don't have very complicated arithmetic and even those that do might not run into this problem. But it'd be a shame if there are programs that we cannot typecheck just because we run out of memory or whatever.

Ok, having written the above I feel like my argument for is not supporting precise types for unions of singletons is rather weak. I simply can't imagine the above scenario causing a problem in practice. So I've convinced myself that I'm for at least this kind of precise types :-)

zuiderkwast:
Because we typecheck one function at a time, the compexity doesn't grow very large, as long as the functions are not expremely complex.

In dialyzer, they get very compex types since they "infer" the whole program and therefore they need to limit the preciseness, e.g. approximate `{A, B} | {C | D}` with `{A | C, B | D}` and things like that. I think we can be more precise than dialyzer. :-)

PR #59, opened 2018-10-28
Unify error terms

Zalastax:
See #49.

This work turned out more tedious than I expected but the codebase was fortunately already structured so that this could be done without intrusive changes.

There's still no documentation, I'd like `type_error()` to have a tight type, and I'd like to get rid of `handle_type_error` by changing what is thrown so I'm leaving #49 open - I will unassign myself from it though.

The different reasons are fairly ad-hoc - it would be nice if someone thinks of a good way to group them and that they document that.
Some error messages are / were quite cryptic so check that I didn't misinterpret any. It would be nice to unify error messages as well (especially when a newline should be used and where to print the line number).

### DIFF
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -8,6 +8,8 @@
 %%%   error is from.
 -module(gradualizer).
 
+-include("typelib.hrl").
+
 -export([type_check_file/1,
          type_check_file/2,
          type_check_module/1,
@@ -15,22 +17,20 @@
          type_check_dir/1,
          type_check_dir/2,
          type_check_files/1,
-         type_check_files/2
+         type_check_files/2,
+         format_error/1,
+         format_errors/2
         ]).
-
--export_type([options/0]).
-
--type options() :: proplists:proplist().
 
 %% API functions
 
 %% @doc Type check a source or beam file
--spec type_check_file(file:filename()) -> ok | nok.
+-spec type_check_file(file:filename()) -> [typechecker:type_error()].
 type_check_file(File) ->
     type_check_file(File, []).
 
 %% @doc Type check a source or beam file
--spec type_check_file(file:filename(), options()) -> ok | nok.
+-spec type_check_file(file:filename(), typechecker:options()) -> [typechecker:type_error()].
 type_check_file(File, Opts) ->
     ParsedFile =
         case filename:extension(File) of
@@ -43,20 +43,30 @@ type_check_file(File, Opts) ->
         end,
     case ParsedFile of
         {ok, Forms} ->
-            Opts2 = proplists:expand([{print_file, [{print_file, File}]}], Opts),
-            typechecker:type_check_forms(Forms, Opts2);
+            case typechecker:type_check_forms(Forms, Opts) of
+                [] -> [];
+                Errors ->
+                    NoReportErrors = proplists:get_bool(no_report_errors, Opts),
+                    if NoReportErrors -> Errors;
+                       true ->
+                            PrintFile = proplists:get_bool(print_file, Opts),
+                            io:format("~s", [format_errors(PrintFile,
+                                                           [{File, Errors}])]),
+                            Errors
+                    end
+            end;
         Error ->
             throw(Error)
     end.
 
 
 %% @doc Type check a module
--spec type_check_module(module()) -> ok | nok.
+-spec type_check_module(module()) -> [typechecker:type_error()].
 type_check_module(Module) ->
     type_check_module(Module, []).
 
 %% @doc Type check a module
--spec type_check_module(module(), options()) -> ok | nok.
+-spec type_check_module(module(), typechecker:options()) -> [typechecker:type_error()].
 type_check_module(Module, Opts) when is_atom(Module) ->
     case code:which(Module) of
         File when is_list(File) ->
@@ -66,14 +76,12 @@ type_check_module(Module, Opts) when is_atom(Module) ->
     end.
 
 %% @doc Type check all source or beam files in a directory.
-%% (Option `print_file' is implicitely true)
--spec type_check_dir(file:filename()) -> ok | nok.
+-spec type_check_dir(file:filename()) -> [{file:filename(), [typechecker:type_error()]}].
 type_check_dir(Dir) ->
     type_check_dir(Dir, []).
 
 %% @doc Type check all source or beam files in a directory.
-%% (Option `print_file' is implicitely true)
--spec type_check_dir(file:filename(), options()) -> ok | nok.
+-spec type_check_dir(file:filename(), typechecker:options()) -> [{file:filename(), [typechecker:type_error()]}].
 type_check_dir(Dir, Opts) ->
     case filelib:is_dir(Dir) of
         true ->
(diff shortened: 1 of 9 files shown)


Zalastax:
Error formatting used to get hit by the code coverage by accident - now it doesn't.
Ideas for testing it? Is it even useful?

src/typelib.erl
```diff
@@ -25,6 +25,14 @@ pp_type(Type) ->
     %	File      -> S ++ " in " ++ File
     %end.
 
+pp_intersection_type([]) ->
+    "";
+%% TODO: pp_type seems to have problems printing bounded types.
```
gomoripeti: yes, bounded funs can only appear in specs but not in types (in the erlang AST) so they are not part of `erl_parse:abstract_type()`.  maybe we could extend `typchecker:type()` or have another type that includes it

Zalastax: Btw, this function is just cut and paste from typechecker.erl so if we want to fix the problem of having it let's do that in a follow up PR.

josefs: Indeed @gomoripeti, we need to distinguish between toplevel function specs and types which fall under `erl_parse:abstract_type()`. In particular, we need proper pretty printing of the former. The code here is just a hack that I made so that the typechecker wouldn't crash.

src/gradualizer.erl
```diff
@@ -83,23 +78,234 @@ type_check_dir(Dir, Opts) ->
...
+        [LINE, typelib:pp_type(Ty)]);
+format_error_(LINE, {logic_op, LogicOp, Ty}, non_boolean) ->
+    io_lib:format("The operator ~p on line ~p is given a non-boolean argument "
+        "of type ~s~n",
+        [LogicOp, LINE, typelib:pp_type(Ty)]);
+format_error_(LINE, {logic_op, LogicOp}, {non_boolean_exp, Ty}) ->
+    io_lib:format("The operator ~p on line ~p is used in a context where it is "
+        "required to have type ~s~n",
+        [LogicOp, LINE, typelib:pp_type(Ty)]);
+format_error_(LINE, {logic_op, LogicOp, Ty1, Ty2}, incompatible) ->
```
zuiderkwast: Apart from the line number element, is there a rule of thumb for what the elements are used for? Here, they are two values which tagged with a descriptive atoms instead of a flat structure of variable number of elements. I don't see how this is better.

src/gradualizer.erl
```diff
@@ -83,23 +78,234 @@ type_check_dir(Dir, Opts) ->
...
+format_error_(LINE, {remote_type, Name, Arity, Module}, not_exported) ->
+    io_lib:format("The type ~s:~s/~p on line ~p is not exported~n",
+        [erl_pp:expr(Module), erl_pp:expr(Name), Arity, LINE]);
+format_error_(LINE, {var, Var, VarTy}, {type_mismatch, Ty}) ->
+    io_lib:format("The variable ~p on line ~p has type ~s "
+        "but is expected to have type ~s~n",
+        [Var, LINE, typelib:pp_type(VarTy), typelib:pp_type(Ty)]);
+format_error_(LINE, {char, Char}, {type_mismatch, Ty}) ->
+    io_lib:format("The character ~p on line ~p does not have type ~s~n",
+        [erl_pp:expr(Char), LINE, typelib:pp_type(Ty)]);
```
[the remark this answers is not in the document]
Zalastax: It might be possible to merge the cases but that would require thinking - which I did not really apply here. Same goes for the wrapper - the translation is mostly mechanical. Could the erl_pp representation change? I'd like to keep using `erl_pp:expr(Char)` without knowing the internal representation.

src/gradualizer.erl
```diff
@@ -83,23 +78,234 @@ type_check_dir(Dir, Opts) ->
...
+format_error_(LINE, {cons, Cons}, {type_mismatch, Ty}) ->
+    io_lib:format("The expression ~s on line ~p does not have type ~s~n",
+        [erl_pp:expr(Cons), LINE, typelib:pp_type(Ty)]);
+format_error_(LINE, nil, {type_mismatch, Ty}) ->
+    io_lib:format("The empty list on line ~p does not have type ~s~n",
+        [LINE, typelib:pp_type(Ty)]);
+format_error_(LINE, {arguments, LenArgs}, {length_mismatch, LenTy}) ->
+    io_lib:format("The clause on line ~p is expected to have ~p ~s "
+        "but it has ~p~n ",
+        [LINE, LenTy, pluralize("argument", "arguments", LenTy), LenArgs]);
```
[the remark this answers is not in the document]
Zalastax: That's the only new feature - the rest is either pure translation or removal of dead code. I hope it's fine that I leave it in the same commit 🙏

src/typechecker.erl
```diff
@@ -2734,197 +2739,109 @@ aux([{attribute, _, compile, export_all} | Forms], Acc) ->
...
+    {type_error, LINE, {mfa, M, F, A, ResTy}, {type_mismatch, FunTy}};
+handle_type_error({type_error, fun_res_type, LINE, Func, FunResTy, ResTy}) ->
+    {type_error, LINE, {result, Func, FunResTy}, {type_mismatch, ResTy}};
+handle_type_error({type_error, expected_fun_type, LINE, Func, FunTy}) ->
+    {type_error, LINE, {function, Func, FunTy}, not_function_type};
+handle_type_error({type_error, no_type_match_intersection,
+                  LINE, Func, FunTy}) ->
+    {type_error, LINE, {call, Func}, {intersection_mismatch, FunTy}};
+handle_type_error({type_error, boolop, BoolOp, LINE, Ty}) ->
+    {type_error, LINE, {bool_op, BoolOp, Ty}, not_boolean};
```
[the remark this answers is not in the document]
Zalastax: That would indeed be better and I mentioned that in my initial comment 😃. I did not have time to check if changing what's thrown would alter behavior. Is it OK that we change change that in a follow up PR?

src/gradualizer.erl
```diff
@@ -83,23 +78,234 @@ type_check_dir(Dir, Opts) ->
...
+
+-spec format_error(typechecker:type_error()) -> string().
+format_error({type_error, LINE, Cause, Mismatch}) ->
+    format_error_(erl_anno:line(LINE), Cause, Mismatch).
+
+-spec format_error_(non_neg_integer(), any(), any()) -> string().
+format_error_(LINE, {call, Func, Arity}, undef) ->
+    io_lib:format("Call to undefined function ~p/~p on line ~p~n",
+        [Func, Arity, LINE]);
+format_error_(LINE, {call, Func, Arity, Module}, undef) ->
```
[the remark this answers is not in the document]
Zalastax: True dat.
Resolve the conversations you think can be addressed later (and take note of them for an issue we should open) and I'll fix what's left.

src/gradualizer.erl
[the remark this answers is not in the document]
zuiderkwast: Sure, that's what I mean. `erl_pp:expr/1` can be used for any expression (AST).
```Erlang
format_error({expr_type_mismatch, Expr, Ty}) ->
    io_lib:format("The expression ~p on line ~p does not have type ~s~n",
        [erl_pp:expr(Expr), get_line(Expr), typelib:pp_type(Ty)]);
```

Zalastax:
I see three paths forward here:
* Merge as is (with small fixes) and work  on improving the error format + removing `handle_type_error` little by little.
* I tear down the rewrapper function `handle_type_error` and expose the old format to `format_error_`. This would lead to a smaller diff which we can merge while we design an official error format.
*  
I discussed today with @zuiderkwast about what should be done for this to become mergeable. The path we agreed on was to have everyone collaborate on what `type_error()` should look like

josefs:
Thanks for biting the bullet here @Zalastax and make some progress on the error handling.

I spoke with @zuiderkwast IRL today and he argued for option 3 (correct me if I midunderstood you @zuiderkwast) were we try to decide upon a format for the errors before pushing any code to the repo. The argument would be that we don't want people to start building tools on top of an unfinished design. Given that we already have eager customers for this interface, I think it makes sense to present them with a more or less complete design.

Some comments on the actual design. 

- I'm not sure we need the atom `type_error` as the "constructor". @Zalastax, you claim that there are a few places that would need it. Can you point them out? 
- I like the spirit of the suggestion from @zuiderkwast to include the original `expr` where the error originated. My main concern is that this expression can be rather big and if we were to print it, the type error would be humongous. A pretty printer which could elide parts of the expression if it was too big would be nice to have. But that would surely take some work. So I'm not sure what the best way forward is. Any ideas?

@zuiderkwast, when we spoke today I got the impression that you had a fairly concrete idea of how you'd like type errors to be represented. Would you care to share it here?

Zalastax:
I'm fine with any of the options as long as we make decent progress and there aren't too many merge conflicts. Given the history of my previous contributions I have no worries on that regard.

@josefs I believe it is needed in these places:
https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L2045
https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L2127
https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L2138
https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L2243
https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L1393

Zalastax:
**Please edit this comment to collaborate on designing the type_error() type**
```Erlang
-type type_error() :: {type_error, erl_anno:anno(), any(), any()}.
```

Look at `gradualizer:format_error_` and `typechecker:handle_type_error` when designing. When the design nears completion I will make a list of which cases have not been considered.

gomoripeti:
some random ideas:

the format could also be:
- a map with mandatory and optional fields
- a tuple with mandatory fields and a last map element with optional fields

When debugging gradualizer itself, it's very useful to know where the error was thrown from. So it would be useful to have a descriptor-tag (more than just `type_error`) or file:line info for this (this later can be only present in the extra-info map when some debug option is enabled). (this is an argument against simple `{type_error, Expr, ExpectedType}` format)

Error location:
- file and line can be extracted from the `Expr` AST (`erl_anno:file/1`, `erl_anno:line/1`) if present.
- I don't know if this is useful or convenient but location could additionally also contain module:function/arity.

I think it is useful in some cases to include both the expected and the actual (inferred) type of the expression (if applicable). Eg
- `variable A is expected to have type X` - "well, I think it has type X"
- instead `variable A is expected to have type X but it has type X | Y` - "oh, right, I didn't think it can also be Y"
(this is an argument against simple `{type_error, Expr, ExpectedType}` format)

(just as a reference what the compiler and dialyzer returns:)
```
compile:file(_, [return_error]) -> {error, Errors, Warnings}
  when
      Erros :: [{Filename, [ErrorInfo]}],
      ErrorInfo :: {ErrorLine, Module, ErrorDescriptor}

compile:format_error(ErrorDescriptor)
```

```
%% The following type has double role:
%%   1. It is the set of warnings that will be collected.
%%   2. It is also the set of tags for warnings that will be returned.
-type dial_warn_tag() :: ?WARN_RETURN_NO_RETURN | ?WARN_RETURN_ONLY_EXIT | ...

%% This is the representation of each warning as they will be returned
%% to dialyzer's callers
-type file_line()    :: {file:filename(), non_neg_integer()}.
-type dial_warning() :: {Tag :: dial_warn_tag(), Loc :: file_line(), Msg :: {atom(), [term()]}}.

%% This is the representation of each warning before suppressions have
%% been applied
-type m_or_mfa()     :: module() % warnings not associated with any function
                      | mfa().
-type warning_info() :: {file:filename(), non_neg_integer(), m_or_mfa()}.
-type raw_warning()  :: {Tag :: dial_warn_tag(), Loc :: warning_info(), Msg :: {atom(), [term()]}}.

dialyzer:format_warning(raw_warning() | dial_warning())
```

Zalastax:
A map makes a lot of sense since it gives us freedom to add new details without breaking backwards compatibility. The other alternative for backwards compatibility would be to provide a bunch of projections (e.g. `line_number(type_error()) -> integer()`) from type_error() and make the type opaque.

zuiderkwast:
> Would you care to share it here?

Ok, what I had in mind:

```Erlang
-type_error() ::
    {expr_type_mismatch, expr(), type()} |  %% e.g. "The character ~s on line ~p does not have type tuple()"
    {type_mismatch, type(), type()} | %% e.g. "The type ~s on line ~p is not a list type"
    {expr_unexpected_type, expr(), ExpectedType :: type(), ActualType :: type()} |
    {undefined_type, type()} | %% undefined local type, remote type or record
    {unexported_type, type()} | %% one error that needs some more info (perhaps not needed)
    {undefined_function, expr()} | %% Call or fun ref to undef function or module
    {clause_bad_arity, Clause, ExpectedArity} |
    ...
```

So, the idea is to omit some of the information when it is *easy enough* to get it out of the expression or type. Where we catch the errors, we can also *insert* a file name annotation before we return the errors.

For the first message, the line number and the word "character" needs to be extracted from the expressions, e.g.

```Erlang
pp_error({expr_type_mismatch, Expr, Type}) ->
    io_format("The ~p ~s on line ~p does not have type ~s",
              [element(1, Expr), erl_anno:line(element(2, Expr)),
               erl_pp:expr(Expr), pp_type(Type)]);
```

Some error messages contain some info about the surrounding context, e.g. that it is inside a bit expression or that negation is not possible on the type. I hadn't thought of this when I got the idea above.

The idea of using the trace within the typechecker for debugging the typechecker is an idea that we can perhaps solve in another way, e.g. not catch the type error if an option is set in the env. Then we get it as a crash report with full trace.

Using typed maps is also an interesting idea. The idea of merging errors some to keep it simpler can still be applied if we use a map.

OvermindDL1:
Also, just printing the error is not very useful for tools, is there any way that we can get a term representation of the errors for processing further in a compilation pipeline?

Zalastax:
@OvermindDL1 the goal of this PR is to create a structured format for errors so that the pipeline can decide what to do with the error. The reason it's so difficult is because we know people will start using it so it needs to be sane from the get go. Is there some information that would be extra useful for your use case? We'd love to hear your ideas on how the error format could look as well!

OvermindDL1:
> Is there some information that would be extra useful for your use case? 

Specifically I'd love to be able to print the erlang terms as elixir terms in the messages, so even something as simple as a list of something like `["something happened at ", {some, [terms]}, " at line ", 42, ...]` would be perfect, though something that actually documents the information in better detail so I can parse out the proper terms to reprint as elixir terms while also getting the line and column information to jump straight to in IDE's would be most helpful (perhaps a `map` of information?).

Zalastax:
@josefs @zuiderkwast should we go for a representation based completely on a map structure or some fixed fields + a map?

@OvermindDL1 After getting a good representation I'll work on creating a pretty printer that represents the output as a string with "holes" to fill in. What about returning arguments intended for [io_lib:format/2](http://erlang.org/doc/man/io_lib.html#format-2)? E.g.
```Erlang
{"something happened at ~p at line ~p", [{some, [terms]}, 42]}
```

OvermindDL1:
> What about returning arguments intended for io_lib:format/2? E.g.

That would be great actually!  I could both parse and transform them via that!

zuiderkwast:
In order to pretty-print `{some, [terms]}`, you need to know if it's an expression or a type or something else, to use the correct pretty-printing function.

I like a structured format that we can use in the tests and such things, but we can provide functions to access the format strings and make them usable.

I like the idea of a map, rather than a tuple with a map. A map is good for the places in the code where we catch an exception and replace an element in it (a normalized type with the non-normalized type). Maps are actually very handy in pattern matching and they can be nicely typed. The next question: Which keys should the map contain?

josefs:
Go ahead and use a map as the representation. My preference is for using a tuple as I think it's more ergonomic to work with but I don't have a strong opinion on this and since everybody else is in favor of using a map then I'll fold.

OvermindDL1:
I was just going to match on the message to determine each type, but if that information is encoded in better ways than I'm all for it.  I prefer tuples too personally though.

Zalastax:
I'm going to fix this PR soon (Wednesdays hackathon perhaps), i.e. fix conflicts, translate to map, and create a function that returns arguments for io:format. The map format will be ad-hoc but I believe it will be easier to agree this time around. I don't know yet if I will remove the need for a translation layer in this PR - though if you object I'll have to comply. If the type was opaque I could have bowed for using tuples (even though I think they are almost always inferior) but with an open Interface tuples will brake compatibility even when just adding data.

zuiderkwast:
Nice! I am thinking about writing a function to "describe" an expression that can be used instead of pretty-printing it when that would be too long. The function would return a string like "list comprehension" or "character" or possibly pretty-printing the expression in cases when it is short enough. Then not all but several of the errors can be pretty-printed using the same code, e.g.

```Erlang
io_lib:format("~p: The ~s is expected to have type ~s but it has type ~s",
              [get_line(Expr), describe_expr(Expr), pp_type(Ty1), pp_type(Ty2)])
```

How about that?

Zalastax:
Feedback welcome again!
I hope the map format is to your liking and that we're able to fine tune rather than redo again.
The goal is that `handle_type_error` is removed but it's easier to do so in a follow-up commit.
Some `error_type`s might be mergeable or tupelifiable - please suggest which; keep in mind that it we'd like the format of the data to be fairly backwards compatible and should not require reading the full context of where the error is thrown to understand what is what.

UlfNorell:
I think it's important to have test coverage of `format_error`. It's calling `erl_pp:expr/1` and `typelib:pp_type/1` and similar functions that can crash (and have done so) if given unexpected input.

We could move the printing into `gradualizer` (it's now duplicated over `gradualizer_cli` and `rebar_prv_gradualizer`) and have an option to print/return errors. We might want to copy the Erlang compiler in this, using `report_errors` and `return_errors`.

src/typechecker.erl
```diff
@@ -20,6 +20,36 @@
...
+                    | logic_op | {rel_op, arguments | expression}
+                    | {list_op, arguments | expression}
+                    | operator_pattern | pattern | tuple
+                    | {bit, arguments | expression} | generator | bin_generator
+                    | check_clauses | record | record_pattern | record_update
+                    | {comprehension, bit_string | list | binary}
+                    | lambda | cyclic_type_vars,
+      line           => erl_anno:anno(),
+      module         => atom(),
+      name           => atom(),
```
josefs: I would suggest we call this field 'function' or something along those lines which is a little more descriptive than 'name'.

src/typechecker.erl
```diff
@@ -20,6 +20,36 @@
...
+      arity          => non_neg_integer(),
+      actual_type    => type() | {type(), type()},
+      expected_type  => type(),
```
josefs: The expected/actual type works well for type checking, when we have a known type which we want to check against an expression. But what about inference? Suppose we're checking the type of an addition expression and one of the operators turns out to be a tuple. What does these fields represent in this situation?

src/typechecker.erl
Zalastax: In that case I should also change
```Erlang
handle_type_error({type_error, tyVar, LINE, Var, VarTy, Ty}) ->
```
to return
```
    #{error_type    => var,
      line          => LINE,
      expression    => Var,
      actual_type   => VarTy,
      expected_type => Ty
      };
```

josefs: Sorry, I don't follow. Can you elaborate?

Zalastax: At the moment it uses the `name` field which if renamed would not make sense there.
I don't want a field for just this case so moving `Var` to `expression` is my preferred solution.

josefs: Right. So you had chosen 'name' specifically because it's ambiguous and can be used in many different context. 
Yeah, using 'expression' for variables sounds like a good solution.

src/typechecker.erl
zuiderkwast: 1. It would be nice if fields with a name like ..._type is always a type(). Perhaps we can encode two types as a tuple type (`{type, _, tuple, [T1, T2]}`) to keep the structure, but a probably better suggestion is that we add another field `expected_operand_types :: [type()]` which can be used for binary operators.
2. I see that actual_type and expected_type are optional. That means the pretty-printing should be able to handle cases where these are missing, independently of type_error. This is probably fine. We can just make sure that pretty-printing is agnostic to this and just prints the fields that are present in the error map.

src/typechecker.erl
```diff
@@ -20,6 +20,36 @@
...
+                    | {call, args | arity | context} | mfa
+                    | return | expected_fun_type | op_type_too_precise
+                    | {arith_op, arguments | expression}
+                    | {int_op, arguments | expression}
+                    | non_number_argument_to_plus | non_number_argument_to_minus
+                    | non_boolean_argument_to_not
+                    | unary_op | non_integer_argument_to_bnot
+                    | logic_op | {rel_op, arguments | expression}
+                    | {list_op, arguments | expression}
+                    | operator_pattern | pattern | tuple
```
zuiderkwast: These are way too many IMO. Some of these (var, char, atom, string, int, float, cons, nil, ......) can either appear as a pattern or as an expression. Exactly which pattern or expression it is is already encoded in the expression field, but we can't see if it appears as a pattern or as an expression because these are syntactically the same. So, what we really need to distinguish between here is `expression | pattern`. Then as a helper for pretty-printing we can have simple function for describing an expression, e.g.

```Erlang
describe_expr(Expr) ->
    case element(1, Expr) of
        lc    -> "list comprehension";
        'case'-> "case expression";
        var   -> "variable " ++ erl_pp:expr(Expr);
        Other -> atom_to_list(Other)
    end.
```

Possibly, other error_types can be merged too, but I haven't thought much about which and how.

src/gradualizer.erl
```diff
@@ -83,23 +80,500 @@ type_check_dir(Dir, Opts) ->
...
+format_error(#{error_type := int,
+      line          := LINE,
+      expression    := I,
+      expected_type := Ty
+      }) ->
+    io_lib:format("The integer ~p on line ~p does not have type ~s~n",
+              [I, LINE, typelib:pp_type(Ty)]);
+format_error(#{error_type := float,
+      line          := LINE,
+      expression    := F,
```
zuiderkwast: I think we need to handle that `expression` and most of the fields are optionals. If we want some fields to be mandatory depending on the value of the type_error field, it should be reflected in the types. Otherwise, throwing type errors will be very error-prone.

But (rather than going back to a union type just to be able to express this dependency) I suggest we make format_error capable of printing all the information in the type_error, regardless of whether it makes sense or not. Only the order of fields needs to be defined somewhere. Then we just need one pretty-printer per field.

Example:

```Erlang
format_error(Error = #{error_type = T}) ->
    [format_error_type(T),
     lists:map(fun (F) -> F(Error) end,
               [format_expression/1,        %% "" when expression is not present
                format_expected_type/1,
                format_actual_type/1, ...])].

File: src/gradualizer.erl [hunk 11]
```diff
@@ -83,23 +80,500 @@ type_check_dir(Dir, Opts) ->
...
+                    NewErrors -> [{File, NewErrors} | Errors]
                 end;
-            (_, nok) ->
-                nok
-        end, ok, Files).
+            (_, Errors) ->
+                Errors
+        end, [], Files).
+
+-spec format_error(typechecker:type_error()) -> string().
```
zuiderkwast: `io_lib:format` doesn't return a string() but rather an iolist(), a nested list of strings.

Zalastax:
@zuiderkwast my job would definitely be easier if this was merged as fast as possible but it's worth lingering while there's still good feedback coming in. If we need more/other data to be thrown we can do that in this PR, I just don't want to start working on removing the extra conversion step yet.

src/typechecker.erl
Zalastax: What about `{undef, record}`, `{undef, Type}`, `unexported_remote_type`?
Perhaps the whole MFA should be in `expression`? I avoided doing so since I don't know what the type of most things in `expression` is.

Zalastax:
@UlfNorell it would be nice to test that error printing doesn't crash but after thinking about it I would prefer not doing so via your suggested solution that uses `report_errors` and `return_errors`. The issue I'm having is that doing so complicates the implementation of multiple functions in `gradualizer`, makes for more complicated types, mixes concerns, and provides little value. I'd much prefer adding a few helper functions, but even that I don't like very much. Perhaps just modify `test.erl` to call the formatter every time - only testing that it doesn't crash.

src/gradualizer.erl
Zalastax: We could of course start specializing the types, e.g.
```Erlang
type type_error() ::
    #{error_type := string,
      line           := erl_anno:anno(),
      expected_type  := type(),
      expression     := string() % I guess...
} | % the rest but excluding string
```
I still find the map very valuable since it lets us add fields and reading code using it is a lot easier than tuples.

src/typechecker.erl
zuiderkwast: I agree about encoding stuff in the other fields so the error_type can just be `undefined` or `unexported`, but in these cases we don't have an expression. These errors are are coming from a bad type in a spec or a type definition. The AST types are defined here: https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_parse.yrl#L605

src/gradualizer.erl
zuiderkwast: I thought about that, but I think a union of maps can be too complex. Especially it might not be obvious that some key is mandatory in all maps in the union.

UlfNorell:
I'm fine with skipping the `return_errors` flag. As you say this complicates the types. But I think having a `report_errors` flag that gets checked in `gradualizer:type_check_file/2` will simplify the code: you can get rid of the printing in `rebar_prv` and `cli` and just pass in `report_errors`, and the test code can do the same. It would also help people (like myself) who wants to run gradualizer from the Erlang shell. I would even go so far as to suggest `report_errors` should default to `true`.

src/typechecker.erl
zuiderkwast: I don't know. Maybe put the bad record or type in a `type` field? Or `bad_type`...

Zalastax:
Tests now hit formatting, thanks to error reporting via `gradualizer` and it being default on (disable with `no_report_errors`).
I can maintain the branch for a while but I won't take the steering wheel so I still want help with simplifications!

Zalastax:
I performed some simplifications (search for `error_type  =>` in typechecker (e.g. [here](https://github.com/josefs/Gradualizer/blob/a185a2422faade4b895cd8e0957ce27dc2f740d9/src/typechecker.erl#L2102-L2105)) and `error_type := type_error` in gradualizer (e.g. [here](https://github.com/josefs/Gradualizer/blob/a185a2422faade4b895cd8e0957ce27dc2f740d9/src/gradualizer.erl#L342-L347)))  and I'm warming up to emitting the AST types directly - it works great when the AST type directly determines the error message. I'm not set on what to do for https://github.com/josefs/Gradualizer/blob/a185a2422faade4b895cd8e0957ce27dc2f740d9/src/typechecker.erl#L1359-L1364 but I believe I'll restructure it so that the original expression can be thrown and the `error_type` will either be `argument_length_mismatch` or `type_error`.

Let me know if this looks like the right way to go or not!

zuiderkwast:
If you don't amend and force-push every time, we can see what has changed since last time. We can squash them when submitting.

src/typechecker.erl
```diff
@@ -1457,39 +1488,47 @@ type_check_expr(Env, {op, _, '!', Proc, Val}) ->
...
     {Ty, VB, Cs1} = type_check_expr(Env, Arg),
     case subtype(Ty, {type, P, boolean, []}, Env#env.tenv) of
         {true, Cs2} ->
             NormTy = normalize(Ty, Env#env.tenv),
             {negate_bool_type(NormTy), VB, constraints:combine(Cs1, Cs2)};
         false ->
-            throw({type_error, non_boolean_argument_to_not, P, Ty})
+            throw({type_error, #{error_type  => type_error,
+                                 expression  => E,
+                                 actual_type => Ty}})
```
zuiderkwast: I like the idea of structuring these types of errors in a uniform way. Here though, some information is lost and there is some assymetry. `E` is the expression `not Arg`. `Ty` is the type of `Arg`. `Arg` is not included in the error and the expected type `boolean()` is not included.

Perhaps we should have `expression => Arg`, `expected_type => type(boolean)` and `context => E`. By context I mean surrounding expression or pattern. Then we have all the information that was encoded in the tag before and can render it as e.g. "The variable Arg in the unary not expression on line 42 has type string() but is expected to have type boolean()".

Apart from this, I think the wrapping tuple is redundant.

Zalastax: Rebasing would get increasingly tricky without amending so I unfortunately think it's the way of working that suits me best here.

Zalastax:
Rebasing would get increasingly tricky without amending so I unfortunately think it's the way of working that suits me best here.

zuiderkwast:
This is a very big project, especially as nobody knows what we really want in the end....

Have you considered starting in the other end, by small changes that can be quickly accepted one by one? E.g. merging similar error forms into fewer ones, reordering fields in the tuples to make them more uniform, align the formulations in the error messages, etc.

src/typechecker.erl
Zalastax: I agree on all your comments.
Wrapping the tuple might be redundant but I know that there is at least one place that catches based on that first element of the tuple so I wasn't sure.

Zalastax:
Smaller steps might help, especially with reviewing.    If the other maintainers agree on doing so I will proceed.
What I'd like to start on in that case is to use the format you suggested where applicable (expression, expected_type, actual_type, context) - either with a map or with a record that has some fields undefined as needed.

zuiderkwast:
> Smaller steps might help, especially with reviewing. If the other maintainers agree on doing so I will proceed.

The smaller the steps, the easier to agree, for sure.

> What I'd like to start on in that case is to use the format you suggested where applicable (expression, expected_type, actual_type, context) - either with a map or with a record that has some fields undefined as needed.

I suggest even smaller steps. How about just making sure all type errors on this form has an expected type and an actual type included? Currently, some don't. Once we start changing the tuple into something else, it somehow requires changing all the type errors everywhere to keep things uniform. We could prepare as much as possible before taking that step.

Zalastax:
> I suggest even smaller steps. How about just making sure all type errors on this form has an expected type and an actual type included? Currently, some don't. Once we start changing the tuple into something else, it somehow requires changing all the type errors everywhere to keep things uniform. We could prepare as much as possible before taking that step.

I'll start with unifying the tuple as much as possible, i.e. pretend that it is the record I suggested wherever that's possible without actually creating the record definition.

PR #64, opened 2018-10-31
Improve testing

Zalastax:
Dialyzer + known problems. See commits + referenced issues.

### DIFF
--- a/test/test.erl
+++ b/test/test.erl
@@ -6,16 +6,39 @@ should_pass_test_() ->
     %% user_types.erl references remote_types.erl
     %% it is not in the sourcemap of the DB so let's import it manually
     gradualizer_db:import_erl_files(["test/should_pass/user_types.erl"]),
-    run_tests_in("test/should_pass", ok).
+    map_erl_files(fun(File) ->
+            ?_assertEqual(ok, gradualizer:type_check_file(File))
+        end, "test/should_pass").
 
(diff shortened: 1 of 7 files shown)


josefs:
This is great! I really like having the known_problems directory. If people wish to contribute it's easy to point them there. It can also function as a backlog of sorts, parallel to the issue track here on github.

test/test.erl
```diff
@@ -6,16 +6,39 @@ should_pass_test_() ->
...
-run_tests_in(Dir, ExpectedRes) ->
-    Files = filelib:wildcard(filename:join(Dir, "*.erl")),
+% Test succeeds if Gradualizer crashes or if it doesn't type check.
+% Doing so makes the test suite notify us whenever a known problem is resolved.
+known_problem_should_pass_test_() ->
+    map_erl_files(fun(File) ->
+        Result =
+            try gradualizer:type_check_file(File) of
+                V -> V
+            catch E -> nok
```
gomoripeti: this will only catch exception of type `throw` (doesn't type check) but not `error`(crash). So a crash of gradualizer would crash the unit test. (I would use `catch _:_ -> nok`) Or I might miss something.

Zalastax: Aha, the point was to catch errors. If you have time please patch it, otherwise I'll do it tomorrow.

PR #66, opened 2018-11-01
Replace erlang.mk with rebar3

### LINKED ISSUE #50: erlang.mk bug
Attempting to run `make escript` on RedHat linux (which is locked to `make` version `GNU Make 3.82` with no alternatives on the main installs), and running `make escript` gives this:
```zsh
╰─➤  make escript
erlang.mk:30: Please upgrade to GNU Make 4 or later: https://erlang.mk/guide/installation.html
erlang.mk:30: Please upgrade to GNU Make 4 or later: https://erlang.mk/guide/installation.html
 GEN    escript-zip

WARNING: No more files
gradualizer

make: *** [escript-zip] Error 1
```
And the supplied link is useless as it says nothing about how to workaround the issue on modern RedHat installs (which regulate what can and cannot be installed).

Zalastax:
Closes #50.

Not sure what to do about the unit test. Should we add the beam file to the test directory and commit it?


Makefile
```diff
@@ -26,34 +29,51 @@
...
+	rebar3 do eunit -c, cover
 
-EUNIT_OPTS = verbose
+.PHONY: showcover
+showcover: cover
+	open _build/test/cover/index.html
 
-# Unit test files to run
-EUNIT_TEST_SOURCES = $(shell ls $(TEST_DIR)/*.erl)
+.PHONY: dialyze
```
gomoripeti: this `dialyze`/`gradualize` was just an `erlang.mk` naming convention. let's just use `dialyzer`/`gradualizer targets.

Makefile
```diff
@@ -26,34 +29,51 @@
...
+dialyze:
+	rebar3 dialyzer
 
-# Override erlang.mk variable to only run unit test modules
-# (results in a bit less noisy output)
-# (Other it tried to compile and run all source files
-#  recursively in all subdirs of test)
-EUNIT_TEST_MODS = $(notdir $(basename $(EUNIT_TEST_SOURCES)))
+.PHONY: check
+check: dialyze tests gradualize
```
gomoripeti: I would prefer first running the tests and then dialyzer, as unit tests are faster.

rebar.lock
```diff
@@ -0,0 +1 @@
+[].
```
zuiderkwast: Wat?

Makefile
```diff
@@ -26,34 +29,51 @@
...
 
-include erlang.mk
+.PHONY: escript
+escript:
+	rebar3 escriptize
+	cp _build/default/bin/gradualizer .
 
+.PHONY: gradualize
 gradualize: escript
 	./gradualizer -pa $(CURDIR)/ebin/ $(CURDIR)/ebin/*beam
```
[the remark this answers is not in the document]
Zalastax: Unfortunately no (as for as I know).

rebar.lock
Zalastax: No dependencies = empty lock file. I find it strange that project_plugins are not locked, but so it is.
I'd prefer to leave it in on the principle that rebar.lock should not be added to .gitignore.

Makefile
Zalastax: We can't keep both around? 😸😈
Or perhaps we could use `dialyze` in the make file and those that want to use `dialyzer` can use `rebar3 dialyzer`?
**Important stuff!**

josefs: I like the suggestions of using `gradualize` in the makefile and `gradualizer` in rebar3. Having the make target `gradualizer` to run the gradualizer on itself is rather confusing because its name suggest that it would build the command line program "gradualizer".

Zalastax: What about `dialyze`?

Makefile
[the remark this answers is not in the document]
gomoripeti: makes sense, it was too wild of an idea :)

Makefile
josefs: Yes, I like `dialyze` too! We should make sure to use the same naming convention for both targets.

gomoripeti: let's keep the old `gradualize`/`dialyze` that we all got used to and love in the make file then.

rebar.lock
zuiderkwast: Ok

Issue #39, opened 2018-10-03
Improve some OTP type specs


josefs:
There are modules in OTP which have really unfortunate type specs, from the point of view of Gradualizer. Take the function `hd/1` for example. It has the following spec:

`hd([term()]) -> term()`

In Gradualizer, the type `term()` is the top of the subtyping hierarchy, which means that the code that consumes the output of `hd/1` has to account for every kind of value there is! In practice it means that all uses of this function will signal a type error. That's no good! A better type would be:

`hd([any()]) -> any()`

One way to solve this problem is to provide our own type specs for problematic modules in OTP. We could instrument `gradualizer_db` to use its own set of specs instead of these modules.

A non-exhaustive list of problematic OTP modules

- lists
- proplists
- maps
- erlang
- mnesia
- ets
- qlc

zuiderkwast:
The long-term solution would be to update these in OTP with more specific types such as `hd([A]) -> A`. It would not impossible to get such a PR accepted, I think.

josefs:
@zuiderkwast, I agree. We should absolutely try to fix the types on OTP. The plan I outlined above is the short- to medium term solution.

jachto:
Where comes the otp code/types into Gradualizer?

KennethL:
I encourage you to make a pull request with the specs you want to improve in OTP. Preferrably one per OTP application.
I think we will accept most of them without problem. For example hd([A]) -> A is something we have thought of as an improvement anyway.

josefs:
@jachto, the Gradualizer doesn't have any special mechanism for importing types from OTP. The file gradualizer_db handles the lookup of types in other modules, and if you want to dig into the details you can start with the function import_module/2.

josefs:
Thanks for the encouragement, @KennethL! It's good to know that you're open to improving the specs. We'll make sure to submit pull request in due course.

tim2CF:
Hello! I think I have similar problems with standard Elixir libraries. Let's say I have a function 

```
  @spec foo :: non_neg_integer()
  def foo do
    [:bar, :buz]
    |> Enum.count
  end
```

Then Gradualizer will say 

```
The expression [bar,buz] on line 0 does not have type t()
```

Where actually **t()** is **Enumerable.t()** which is **term()**. What will be the best solution here? To create `./priv/elixir_spec_fix.erl`? Or maybe we should develop mechanism to override types locally in context of project where Gradualizer is executed for type checks?

[...]

PR #67, opened 2018-11-05
Override incorrect or inexact OTP specs

zuiderkwast:
The file priv/otp_spec_fix.erl contains specs that will be loaded
before any other specs. Later, when loading specs, already defined
specs are not overwritten.

### DIFF
--- a/src/gradualizer_db.erl
+++ b/src/gradualizer_db.erl
@@ -125,6 +125,8 @@ import_module(Module) ->
 
 -type opts() :: #{autoimport => boolean()}.
 -define(default_opts, #{autoimport => true}).
+-define(prelude_erl, filename:join(code:priv_dir(gradualizer),
+                                   "otp_spec_fix.erl")).
 
 -record(state, {specs   = #{} :: #{mfa() => [type()]},
                 types   = #{} :: #{mfa() => #typeinfo{}},
@@ -146,7 +148,8 @@ init(Opts0) ->
                  _ ->
                     State1
              end,
-    {ok, State2}.
+    State3 = import_erl_files([?prelude_erl], State2),
+    {ok, State3}.
 
 -spec handle_call(any(), {pid(), term()}, state()) -> {reply, term(), state()}.
 handle_call({get_spec, M, F, A}, _From, State) ->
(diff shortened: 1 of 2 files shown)


gomoripeti:
great! just a few remarks:
- "not overwritten." although this makes implementation simple, this feels to me opposite of how code-loading works - a bit not-least-surprising. Probably in practice there is no spec clash when checking a static, unchanging code base. A warning message would definitely be good.
- isn't there a need to support different specs for different OTP versions? (eg a spec is fixed in a new OTP version so it can be removed from otp_spec_fix for that version. Or a function is extended in a new OTP version so also the fixed spec needs extension just for that new version)

zuiderkwast:
> * "not overwritten." although this makes implementation simple, this feels to me opposite of how code-loading works - a bit not-least-surprising. Probably in practice there is no spec clash when checking a static, unchanging code base. A warning message would definitely be good.

Yes, it's simple. Just adding a warning here would give a warning for all functions we have overrided when those modules (erlang, lists, etc.) are loaded. That's not desirable either. Do you have a suggestion?

> * isn't there a need to support different specs for different OTP versions? (eg a spec is fixed in a new OTP version so it can be removed from otp_spec_fix for that version. Or a function is extended in a new OTP version so also the fixed spec needs extension just for that new version)

It's possible to achieve in the otp_spec_fix.erl using `-ifdef`, `-if`, etc.

```Erlang
-ifdef(OTP_RELEASE).
%% This macro is only defined in OTP >= 21.0
-spec erlang:hello_joe() -> {system_working, seems_to_be}.
-endif.
```

src/gradualizer_db.erl
```diff
@@ -491,12 +501,12 @@ collect_specs(Module, Forms) ->
...
-        Module -> ok;
-        _ -> error_logger:info_report([{spec_for, {M,F,A}},
-                                       {found_in, Module}])
-    end,
+normalize_spec(Spec = {{_M, _F, _A}, _Types}, _Module) ->
+    %case M of
+    %    Module -> ok;
+    %    _ -> error_logger:info_report([{spec_for, {M,F,A}},
+    %                                   {found_in, Module}])
+    %end,
```
josefs: I suggest removing this code instead of keeping it in comments.

zuiderkwast: Done

josefs:
@gomoripeti, regarding your comments, are they something you want to see fixed? Or should we go ahead and merge this PR and continue the discussion afterwards?

zuiderkwast:
OK, good. I made hd and tl only accept non-empty lists (as thery do in OTP already). More type overrides can be added whenever there is a need. @josefs, you have some in mind I guess.

josefs:
Indeed, I have quite a lot of overrides that I'd like to add. I just have very little time to code right now, so don't hold your breath.

gomoripeti:
hijacking the conversation:

what would be a good spec for `proplists:get_value/2`?
in OTP it is `get_value(Key :: term(), List :: [term()]) -> term().`
Obviously that `term()` return value is a bit too much. (eg if I call it with `List :: [{atom(), integer()}]`)

josefs:
A safe but conservative option is to just replace all occurrences of `term()` with `any()`.

If we want to be a bit more precise we could give proplists the type `[atom()|tuple()]`.

zuiderkwast:
Since this is closed now, can we discuss individual specs to override in #39 instead?

PR #78, opened 2018-11-07
Add known problems

Zalastax:
Test cases inspired by a crash when checking OTP's `log_mf_h`.


test/test.erl
```diff
@@ -34,7 +34,7 @@ known_problem_should_fail_test_() ->
         Result =
             try gradualizer:type_check_file(File) of
                 V -> V
-            catch E -> ok
+            catch _:_ -> ok
```
zuiderkwast: Does this mean that typechecker crash is OK for should_fail tests?

Zalastax: Yes, kinda. Crash is OK both for `should_fail` and `should_pass`.

My idea is that you design a test case that you want to place in `should_pass` or `should_fail`. If that test case does not have the behaviour you want, either because it crashes or because it returns the wrong result, you move it to the same `should_pass` or `should_fail` directory but in the `known_problems` directory. The test case will report whenever the known problem is fixed and can be moved out from `known_problems`. To get this behaviour the test for `should_fail` tests that there is a crash or `ok` is returned and the test for `should_pass` tests there is a crash or `nok` is returned.

zuiderkwast: OK! Makes sense.

PR #79, opened 2018-11-08
Known problem: not yet supported union of maps == map of unions 

gomoripeti:
Also move already existing map examples from `misc` to `should_pass`/`should_fail`.
And show file name when should_pass/should_fail/known_problem fails.

I just had this example lying around and with the advent of known_problems there is now a place to add it.

### DIFF
--- a/test/known_problems/should_pass/map_union.erl
+++ b/test/known_problems/should_pass/map_union.erl
@@ -1,0 +1,6 @@
+-module(map_union).
+
+-compile(export_all).
+
+-spec union_value2(#{a => b | c}) -> #{a => b} | #{a => c}.
+union_value2(M) -> M.
--- a/test/test.erl
+++ b/test/test.erl
@@ -7,12 +7,12 @@ should_pass_test_() ->
     %% it is not in the sourcemap of the DB so let's import it manually
     gradualizer_db:import_erl_files(["test/should_pass/user_types.erl"]),
     map_erl_files(fun(File) ->
-            ?_assertEqual(ok, gradualizer:type_check_file(File))
+            ?_assertMatch({ok, _}, {gradualizer:type_check_file(File), File})
         end, "test/should_pass").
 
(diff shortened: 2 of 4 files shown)


test/test.erl
```diff
@@ -24,7 +24,7 @@ known_problem_should_pass_test_() ->
                 V -> V
             catch E -> nok
             end,
-        ?_assertEqual(nok, Result)
+        ?_assertMatch({nok, _}, {Result, File})
```
Zalastax: Is this for improving the error reporting if the assertion fails?

josefs:
I'll wait to merge until you've had the chance to answer the question from @Zalastax.

test/test.erl
gomoripeti: yes, exactly. this is a bit of poor-man's, badmatch solution, but good enough for now.

example before
```
**error:{assertEqual,[{module,test},
              {line,27},
              {expression,"Result"},
              {expected,nok},
              {value,ok}]}
```
after
```
**error:{assertMatch,[{module,test},
              {line,27},
              {expression,"{ Result , File }"},
              {pattern,"{ nok , _ }"},
              {value,{ok,"test/known_problems/should_pass/map_union.erl"}}]}
```

test/known_problems/should_pass/map_union.erl
```diff
@@ -0,0 +1,6 @@
+-module(map_union).
+
+-compile(export_all).
+
+-spec union_value2(#{a => b | c}) -> #{a => b} | #{a => c}.
```
zuiderkwast: This can be solved in the `normalize` function I suppose. Merging maps within a union, like we merge integer intervals, etc. What do you think about that?

josefs: Solving this in the same way as we handle unions of integers is the right thing to do. I agree.

PR #80, opened 2018-11-08
Handle user-def types expanding to fun types in calls and fun refs

zuiderkwast:
Fixes one of the `known_problems`


src/typechecker.erl
```diff
@@ -837,18 +853,19 @@ expect_fun_type({type, _, 'fun', []}) ->
 expect_fun_type({type, _, 'fun', [{type, _, any}, ResTy]}) ->
     {fun_ty_any_args, ResTy, constraints:empty()};
 expect_fun_type(Tys) when is_list(Tys) ->
+    %% This is a spec, not really a type().
```
josefs: Indeed. As we talked about earlier this week IRL, it'd be good to properly separate function types which comes from `-spec` and can be intersection types, and function types which are part of `type()`.

PR #82, opened 2018-11-11
Use source code escript instead of escript archive

### LINKED ISSUE #74: Error loading otp_spec_fix.erl from escript
When gradualizer is run as an escript (as on travis) it crashes with the below error:
```
{"init terminating in do_boot",{{badmatch,{error,enotdir}},
[{gradualizer_db,import_erl_files,2,[{file,"/home/travis/build/josefs/Gradualizer/_build/default/lib/gradualizer/src/gradualizer_db.erl"},{line,333}]},
 {gradualizer_db,init,1,[{file,"/home/travis/build/josefs/Gradualizer/_build/default/lib/gradualizer/src/gradualizer_db.erl"},{line,151}]}...
```

This can probably have something to do with `code:priv_dir(gradualizer) => "/home/travis/build/josefs/Gradualizer/gradualizer/gradualizer/priv"` because it is running from an escript archive.

(https://travis-ci.com/josefs/Gradualizer/jobs/156473749)

zuiderkwast:
This fixes #74, the problem of packaging the priv dir into an
escript archive.


zuiderkwast:
This adds ./gradualizer to the repo. Before it was built. This means that you will have do delete the local file gradualizer before you pull this change. (Btw, I will remove /gradualizer from .gitignore)

zuiderkwast:
The script should be in the `bin/` directory according to http://erlang.org/doc/design_principles/applications.html#directory-structure-for-a-released-system ...

gradualizer
```diff
@@ -1,10 +1,26 @@
--module(gradualizer_cli).
--export([main/1]).
+#!/usr/bin/env escript
+%% -*- erlang -*-
+%%!
+
+-define(ebin_dir, "_build/default/lib/gradualizer/ebin").
```
gomoripeti: can we remind me of the purpose of the cli script? (apart from running it from a standalone git checkout)

I am a bit concerned hardcoding a rebar3 specific `ebin_dir` here (and also bit concerned that the escript won't be self-contained any more). I'm not sure this will work as a mix or erlang.mk dep (but there could be separate plugins for these tools). Also I guess this won't work if gradualizer is a rebar3 dependency (but there is rebar3 plugin for that case already)

I wonder if there are any other use-cases where it's the cli that should be used but it won't work?

Zalastax: I see the cli just as another way of interacting with Gradualizer. Its purpose is not to specifically to help with the interaction towards mix (which there is already a plugin in the works for) or erlang.mk but it should be possible to use the cli there as with any cli. We require rebar3 to build so those who want to build the cli will need to use rebar3 and we should be fine with that. Hardcoding is unfortunate but there is no good way around it. In light of https://github.com/erlang/rebar3/issues/1936 I believe that what @zuiderkwast has presented here is the correct solution.

gomoripeti: I see, so the cli is not just that single escript but the repo as a whole. And the supported way of using the cli is to: clone the repo+`make app`+(optionally add the repo dir to the PATH).
I agree with you (myself as an ignorant example :) ) to highlight usage in the README. (eg it is not supported to copy or symlink just the escript file somewhere else - `filename:dirname(?FILE)` would give wrong output)

Zalastax: While it would be nice that the CLI is self contained, it's OK if it's not (e.g. all node cli's are shipped this way). As you say, good documentation regarding this is key. Let's revisit this topic once OTP/rebar3 ships with standardized support for `/priv` inside escripts.

zuiderkwast:
I don't like hardcoding the rebar3 specific ebin_dir.

I guess there are different use cases. If we want to distribute gradualizer as a standalone, like rebar3 or as a debian package or something like that, it should be self-contained.

Perhaps implementing the extraction of the file inside the archive, parsing it and passing it to gradualizer_db is a better way.

gomoripeti:
I am actually fine with the currently proposed solution (as long as it is documented)

An alternative might be that the `otp_spec_fixes` is not a random data file but an erlang module, so maybe it could be handled as all other gradualizer modules and maybe in this case the `code` or `beam_lib` modules would be able to retrieve its AST (maybe by first loading it) from the escript zip. I haven't tried this though, just brainstorming.

Zalastax:
I'd like to merge this (after documentation fixes) and when someone feels inclined they can implement that as an improvement. The cli does not work at all at the moment and the path forward regarding such an involved escript as we have is unclear. I want to avoid escript with `/priv` until there is an official stance from rebar3.

Issue #89, opened 2018-11-16
Unexpected "pattern <<"ok">> doesn't have the type undefined | <<_:_*8>>" error


erszcz:
I was happily surprised when Gradualizer pointed it out that I don't have to check for `undefined` if my function expects just `t1/0` (BTW, I was writing production code - this is just a minimised example):

```erlang
-module(test).

-export([f/1]).

-type t1() :: binary().

-spec f(t1()) -> ok | error | other.
f(T) ->
    case T of
        undefined -> error;
        _ -> other
    end.
```

However, it puzzled me way more when I changed the code to the following and got `The pattern <<"ok">> on line 12 doesn't have the type undefined | <<_:_*8>>` error:

```erlang
-module(test).

-export([f/1]).

-type t1() :: binary().
-type maybe(T) :: undefined | T.

-spec f(maybe(t1())) -> ok | error | other.
f(T) ->
    case T of
        undefined -> error;
        <<"ok">> -> ok;
        _ -> other
    end.
```

I'm not sure if it's a known issue. It resembles #16 a bit, but that one is already closed.

zuiderkwast:
Thanks for the example! If you want to add this file to test/known_problems/should_pass/ it is much appreciated.

It is kind of known that we some unhandled cases like this one and similar. In this case, it is probably that the binary pattern checked against a union type or a maybe because it's a user-defined type that isn't expanded to a union in this context. (The function `add_type_pat` expects a normalized type.)

erszcz:
Awesome!

PR #90, opened 2018-11-16
Add binary_literal_pattern.erl to tests which should pass

erszcz:
Adding the example from #89 as requested.


zuiderkwast:
Thanks!

PR #92, opened 2018-11-18
Compute sizes of bit expressions and patterns

### LINKED ISSUE #75: typechecker:handle_type_error({type_error,bin,0,...
```
typechecker:handle_type_error({type_error,bin,0,           
                                                              {user_type,[{file,"test_path.erl"},{location,24}],t,[]}}) (/home/overminddl1/erlang/gradualizer/_build/default/lib/gradualizer/src/typechecker.erl, line 2807)
```
It seems binary types are not being handled by the type error path?  So should be a simple fix.  Although the fact that the binary is not passed in like it is for strings/charlists seems odd?  Also `typelib:pp_type(Ty)` does not seem to add the remote module name to the type so I just get something like `t()` instead of `test_path:t()` or so, and with how common `t` is as a type name that's not terribly helpful, plus just seeing `t` isn't really helpful either, it would be nice to get a list of types drilled down.  :-)

zuiderkwast:
This allows us to do more exact checking of bit expressions and patterns.

Fixes #75, #89.


src/gradualizer_bin.erl
```diff
@@ -0,0 +1,121 @@
+%% Helper module for binaries/bitstrings
+-module(gradualizer_bin).
+
+-export([compute_type/1]).
+
+%% Computes the type of a bitstring expression or pattern based on the sizes
+%% of the elements. The returned type is a normalized bitstring type.
+-spec compute_type(ExprOrPat) -> erl_parse:type()
+        when ExprOrPat :: {bin, _, _},
+             ExprOrPat :: erl_parse:expr().
```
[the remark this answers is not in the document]
zuiderkwast: Oh, good catch! It should be erl_parse:abstract_type() as well.

src/gradualizer_bin.erl
```diff
@@ -0,0 +1,121 @@
...
+
+%% <<_:B, _:_*U>> is represented as {B, U} (fixed base + multiple of unit)
+-type bitstr_view() :: {non_neg_integer(), non_neg_integer()} | none.
+
+bitstr_concat({B1, U1}, {B2, U2}) ->
+    {B1 + B2, gcd(U1, U2)};
+bitstr_concat(none, _) -> none;
+bitstr_concat(_, none) -> none.
+
+-spec bitstr_view_to_type(bitstr_view()) -> erl_parse:type().
```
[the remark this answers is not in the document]
zuiderkwast: Right

test/gradualizer_bin_tests.erl
```diff
@@ -0,0 +1,62 @@
...
+    ].
+
+compute_type_float_test_() ->
+    [
+     ?_assert_bin_type(<<_:64>>,            <<X/float>>),
+     ?_assert_bin_type(<<_:32>>,            <<X:32/float>>),
+     ?_assert_bin_type(<<_:32, _:_*32>>,    <<X:S/float>>),
+     ?_assert_bin_type(<<_:32, _:_*32>>,    <<X:S/float-unit:16>>),
+     ?_assert_bin_type(<<_:64>>,            <<X:S/float-unit:64>>),
+     ?_assert_bin_type(<<_:192>>,           <<"abc"/float>>),
```
gomoripeti: `<<"abc"/float>>` is not supported on OTP 19 and travis fails with it (but ok on OTP 20+)
good to know :), I did not know this syntax (as it turns out equivalent with `<<$a/float, $b/float, $c/float>>` and <<97/float, 98/float, 99/float>>`)

zuiderkwast: Thanks for finding the cause of this! Any suggestion on how to skip these cases on OTP 19?

zuiderkwast: It's a pretty useless syntax btw. Wonder why they added it. Once I suggested on erlang-questions to add `<<BoolExpr:1/boolean>>` but that was not popular at all. :-)

gomoripeti: - maybe using the `OTP_RELEASE` macro (and test only from OTP 21+)
- or as it fails at runtime (not compiler error), there could be a helper function which checks if this syntax throws a badarg

gomoripeti: I remember the `/boolean` issue :) the community couldn't decide if `0` should be true or false...

UlfNorell: > It's a pretty useless syntax btw. Wonder why they added it.

Some git archaeology suggests that it was added unintentionally in erlang/otp@a5fcd4f269.

zuiderkwast: Impressive finding @UlfNorell! And Jose Valim is to blame :) I don't mind that this syntax exists, even if it is useless.

PR #93, opened 2018-11-18
Unfold bounded type variables in type specs

### LINKED ISSUE #84: Crash on type variable
```erlang
-spec fff() -> Result when Result :: integer().
fff() -> 0.

ggg() -> -fff().
```
```
** exception error: no function clause matching typechecker:int_type_to_range({var,6,'Result'}) (../src/typechecker.erl, line 604)
     in function  typechecker:negate_num_type/1 (../src/typechecker.erl, line 658)
     in call from typechecker:type_check_expr/2 (../src/typechecker.erl, line 1219)
     in call from typechecker:do_type_check_expr_in/3 (../src/typechecker.erl, line 1599)
     in call from typechecker:type_check_expr_in/3 (../src/typechecker.erl, line 1595)
     in call from typechecker:check_clause/4 (../src/typechecker.erl, line 2395)
     in call from lists:map/2 (lists.erl, line 1239)
     in call from typechecker:check_clauses/4 (../src/typechecker.erl, line 2376)
```

UlfNorell:
Currently (before this PR) we turn bounded type variables (from `when`-clauses) into constraints, but we never actually try to solve the constraints. For instance, the following passes type checking
```erlang
-spec foo(A) -> integer() when A :: boolean().
foo(X) -> X.
```
My simple solution is to unfold `when`-clauses when looking up type specs, so that the above is equivalent to
```erlang
-spec foo(boolean()) -> integer().
foo(X) -> X.
```
This may or may not be exactly what those `when`-clauses are intended to mean, since `::` is pronounced `is_subtype`, but on the other hand I don't think you *can* give a sensible semantics to them if it really means bounded quantification (do you really want `-spec foo() -> A when A :: integer()` to mean `∀ A ⊆ integer(). () -> A`, and thus not have any implementations?)

If a variable has multiple `::`-bindings I take the intersection (`glb`) of them.

This fixes #84 among other things.


UlfNorell:
I should say that this doesn't address the problem with type variables that aren't bound in `when`-clauses. We still ignore constraints on those:
```erlang
-spec foo(A) -> A.
foo(X) -> {X, X}. %% passes
```

UlfNorell:
Just found this call: https://github.com/josefs/Gradualizer/blob/d74675abc3c739e70e04122354f1cd5a642f1766/src/typechecker.erl#L1327
which probably needs addressing. Also some more tests would be nice.

OvermindDL1:
Wouldn't unfolding prevent named versions from matching?  So this:
```erlang
-spec foo(A, B) -> {A, B} when A :: any() when B :: any().
```
(Or whatever the syntax was, been doing too much elixir lately...)
But that above spec should fail this function:
```erlang
foo(A, B) -> {B, A}
```
Where simple unfolding would match, but named versions should not match.

UlfNorell:
What I'm saying here is that I think it's more useful if we don't treat `when`-blocks as introducing bounded quantification. For a few reasons:
- I don't know how to give a sensible semantics to variables only appearing in the result (or negatively in a higher order argument). This is done all the time, for instance the spec for `erlang:erase/1` is
  ```erlang
  -spec erase(Key) -> Val | undefined when
        Key :: term(),
        Val :: term().
  ```
  If read as bounded quantification this says that for all subtypes of `term()`, `erase` returns either an element of this type or `undefined`, basically saying that it always returns `undefined`.
- I know that the [documentation](http://erlang.org/doc/reference_manual/typespec.html#specifications-for-functions) says that it's bounded quantification, but they don't really give any compelling examples. One is the type of the identity function on tuples:
  ```erlang
  -spec id(X) -> X when X :: tuple()
  ```
  and the other is the function
  ```erlang
  -spec foo({X, integer()}) -> X when X :: atom();
           ([Y]) -> Y when Y :: number()`.
  ```
  which when given a pair of an atom and an integer returns the atom, and when given a list of numbers returns one of the numbers in the list.

  They also say that you are free to ignore when a type variable appears more than once:
  > *However, it is up to the tools that process the specifications to choose whether to take this extra information into account or not.*

Basically as far as I can see, bounded quantification
- doesn't do anything for variables occurring once in arguments
- makes no sense for variables occurring once in result
- is way too strict for variables occurring more than once

UlfNorell:
>> makes no sense for variables occurring once in result

> Eh, it's still documentative naming.

No, that's the semantics I'm saying we should adopt. In the bounded quantification semantics `-spec foo() -> A when A :: integer()` is the same as `-spec foo() -> none()`, which is why I'm saying that it doesn't make sense.

>> is way too strict for variables occurring more than once

> How so?

Would you really expect the function `-spec foo([Y]) -> Y when Y :: number()` to only type check when returning one of the numbers already in the list? No adding them up or multiplying them?

OvermindDL1:
> No, that's the semantics I'm saying we should adopt. In the bounded quantification semantics `-spec foo() -> A when A :: integer()` is the same as `-spec foo() -> none()`, which is why I'm saying that it doesn't make sense.

Wait what, it does *that* right now?  o.O

Does dialyzer not error for such mis-matches then?

> Would you really expect the function `-spec foo([Y]) -> Y when Y :: number()` to only type check when returning one of the numbers already in the list? No adding them up or multiplying them?

Following HM typing that would tell me that foo takes a list of integers and returns an integer, it only matches types, not values.  Does the spec really state that it must be a 'value' that matches in the list, not just the 'type' of the values in the list?

src/typelib.erl
```diff
@@ -106,6 +106,8 @@ annotate_user_types(Filename, {type, Anno, T, Params}) when is_list(Params) ->
     {type, Anno, T, [annotate_user_types(Filename, Param) || Param <- Params]};
 annotate_user_types(Filename, {ann_type, Anno, [Var, Type]}) ->
     {ann_type, Anno, [Var, annotate_user_types(Filename, Type)]};
+annotate_user_types(Filename, Types) when is_list(Types) ->
```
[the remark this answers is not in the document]
UlfNorell: I know it doesn't, and I really tried, but decided in the end that it wasn't worth the hassle. The problem is that for some unfathomable reason the `constraint` type has [a nested list in the arguments](https://github.com/erlang/otp/blob/0fc332c2/lib/stdlib/src/erl_parse.yrl#L1172).

I promise to fix it once we have a decent chance of gradualizing the gradualizer.

UlfNorell:
It didn't enter my mind that this was something that might show up, so it probably doesn't work right now. It would make sense to do it though.

UlfNorell:
> Does dialyzer not error for such mis-matches then?

I'm pretty sure dialyzer doesn't implement bounded quantification.

>> Would you really expect the function `-spec foo([Y]) -> Y when Y :: number()` to only type check when returning one of the numbers already in the list? No adding them up or multiplying them?

> Following HM typing that would tell me that foo takes a list of integers and returns an integer, it only matches types, not values. Does the spec really state that it must be a 'value' that matches in the list, not just the 'type' of the values in the list?

The problem is that values *are* (singleton) types. Under the bounded quantification semantics, the type of `foo` is `∀ Y ⊆ number(). [Y] -> Y`. For instance, we can instantiate `Y` to the singleton type `17` to get `foo : [17] -> 17`. Or for any three numbers `N₁`, `N₂`, `N₃`, `foo : [N₁ | N₂ | N₃] -> N₁ | N₂ | N₃`.

Bounded quantification really only makes sense if you have some operations that all subtypes are closed under.

OvermindDL1:
> I'm pretty sure dialyzer doesn't implement bounded quantification.

I thought so, so just HM style typing should be entirely sufficient I'd think?  If you want bounded/unbounded mixed types that going the polymorphic style typer as in ocaml would solve that fairly well, except you might need to introduce a new type or two to act as the constraint types.

josefs:
Thanks Ulf for making progress on type variable constraints.

When I first looked at this PR I was very skeptical for a number of reasons:

- I had always envisioned using bounded quantification as the semantics, and that's what the documentation says.
- We will have to implement constraint solving anyway, so why have two different ways of resolving the bounds on type variables?
-  I'm not worried about the fact that, semantically, a type like `-spec foo() -> A when A :: integer()` doesn't have any inhabitants. After all, there are many types that are not inhabited. If there are specs like this in OTP, we already have a mechanism in place for overriding them with more sensible specs.

But after thinking some more about it I've come to really like this semantics. Here's what convinced me:
- The real problems with specs like `-spec foo() -> A when A :: integer()` is that people actually write these kinds of specs. Unfolding constraints matches better how erlang programmers actually write specs. I've seen many examples where people use constraints as if they were equalities.

So I'm happy to accept this PR. Great job, Ulf!

I've been meaning to implement that constraint solver for a long time, but there's always been lower hanging fruit to pick first. Given my current schedule it's not likely to happen before Christmas so if anyone wants to step up and have a go at it, that's be more than welcome.

zuiderkwast:
@Zalastax The erlang-questions mailing list, https://bugs.erlang.org and https://github.com/erlang/otp are the official communication channels to improve OTP, AFAIK. Github if it is a strait-forward change. The mailing list if it needs to be discussed.

src/typechecker.erl
```diff
@@ -1136,6 +1139,44 @@ new_type_var() ->
...
+    solve_bounds(TEnv, Env, SCCs, #{}).
+
+solve_bounds(TEnv, Defs, [{acyclic, X} | SCCs], Acc) ->
+    #{ X := Tys } = Defs,
+    Tys1 = subst_ty(Acc, Tys),
+    %% Take intersection after substitution to get rid of type variables.
+    Ty1  = lists:foldl(fun(S, T) -> glb(S, T, TEnv) end, type(term), Tys1),
+    solve_bounds(TEnv, maps:remove(X, Defs), SCCs, Acc#{ X => Ty1 });
+solve_bounds(_, _, [{cyclic, Xs} | _], _) ->
+    throw({cyclic_dependencies, Xs});
```
josefs: Here's a rather late comment, as I didn't realise this until now: If we have 'A :: B' and 'B :: A' then that amounts to A being equal to B. We could detect that but it's probably not worth it.

I think it'd be good to try to report an informative error message here, since it is fairly easy to accidentally write cyclic constraints by mistake for someone who uses constraints excessively. @UlfNorell, would you be willing to tackle that?

UlfNorell: Yes, I'm working on this. There are also some cases of cyclic dependencies in OTP. For instance
```erlang
-spec lists:flatten(DeepList) -> List when
  DeepList :: [term() | DeepList],
  List :: [term()].
```
The first step is to print a nice error, but we might indeed consider doing some work to solve the equations.

UlfNorell: #99

PR #95, opened 2018-11-22
More precise typing rule for shortcut operators (andalso and orelse)

UlfNorell:
Essentially
```erlang
-spec andalso(boolean(), false | T) -> false | T.
```
and same but with `true` for `orelse`.

See discussion in #23

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1575,7 +1575,13 @@ type_check_logic_op(Env, Op, P, Arg1, Arg2) ->
                 false ->
                     throw({type_error, boolop, Op, P, Ty2});
                 {true, Cs4} ->
-                    {normalize({type, erl_anno:new(0), union, [Ty1, Ty2]}, Env#env.tenv)
+                    Inferred =
+                        case Op of
+                            'andalso' -> type(union, [Ty1, {atom, erl_anno:new(0), false}]);
+                            'orelse'  -> type(union, [Ty1, {atom, erl_anno:new(0), true}]);
+                            _         -> type(boolean)
+                        end,
+                    {normalize(Inferred, Env#env.tenv)
                     ,union_var_binds(VB1, VB2, Env#env.tenv)
                     ,constraints:combine([Cs1,Cs2,Cs3,Cs4])}
             end
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1568,7 +1568,10 @@ type_check_logic_op(Env, Op, P, Arg1, Arg2) ->
             throw({type_error, boolop, Op, P, Ty1});
         {true, Cs2} ->
             {Ty2, VB2, Cs3} = type_check_expr(Env#env{ venv = UnionVarBindsSecondArg(Env#env.venv,VB1 )}, Arg2),
-            case subtype(Ty2, {type, P, bool, []}, Env#env.tenv) of
+            % Allow any() in second argument for shortcut operators
+            SndArgTy = if Op == 'andalso'; Op == 'orelse' -> type(any);
+                          true                            -> type(bool) end,
+            case subtype(Ty2, SndArgTy, Env#env.tenv) of
                 false ->
                     throw({type_error, boolop, Op, P, Ty2});
                 {true, Cs4} ->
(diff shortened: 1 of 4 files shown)


src/typechecker.erl
```diff
@@ -1568,7 +1568,10 @@ type_check_logic_op(Env, Op, P, Arg1, Arg2) ->
             throw({type_error, boolop, Op, P, Ty1});
         {true, Cs2} ->
             {Ty2, VB2, Cs3} = type_check_expr(Env#env{ venv = UnionVarBindsSecondArg(Env#env.venv,VB1 )}, Arg2),
-            case subtype(Ty2, {type, P, bool, []}, Env#env.tenv) of
+            % Allow any() in second argument for shortcut operators
+            SndArgTy = if Op == 'andalso'; Op == 'orelse' -> type(any);
+                          true                            -> type(bool) end,
+            case subtype(Ty2, SndArgTy, Env#env.tenv) of
```
gomoripeti: nitpicking for below unchanged line `{normalize({type, erl_anno:new(0), union, [Ty1, Ty2]}, Env#env.tenv)`

I think the inferred type of short-circuit ops should not be `union(Ty1, Ty2)` but `union(false, Ty2)` or `union(true, Ty2)` (respectively for andalso and orelse - as you stated in the typing rule)

because the below code would trigger a type error
```
f() ->
  B = get_bool(),
  I = get_int(),
  R = B andalso I,
  g(R).

-spec get_bool() -> boolean().
-spec get_int() -> integer().
-spec g(false | integer()) -> any().
```
`g/1` is called with `R` which has the inferred type of `boolean() | integer()`, which is not a subtype of expected argument type `false | integer()`

This is a different code path than type checking, where we make the outer result type flow down/inwards/backwards.

UlfNorell: Well spotted, I'll update it.

UlfNorell: Also the `union(Ty1, Ty2)` is wrong for `xor`. This will infer `true xor true :: true`.

UlfNorell: Pushed a fix. I relaxed the inferred type to `boolean()` for non-shortcut ops. While we could technically be more precise, I don't think it's worth it.

test/should_fail/shortcut_ops.erl
```diff
@@ -0,0 +1,16 @@
+-module(shortcut_ops).
+
+-compile(export_all).
+
+-spec andalso_too_precise1(true, number()) -> number().
+andalso_too_precise1(True, N) -> True andalso N.
+
+-spec andalso_too_precise2(false, number()) -> false.
```
zuiderkwast: For `should_fail`, we only require one function to fail in a module. You probably want all these to fail. (We can perhaps change how these tests work, but sometimes you want a (passing) helper function in a test module like this, so requiring all functions to fail is maybe not good either. Perhaps we can require all the exported functions to fail...)

src/typechecker.erl
```diff
@@ -1568,11 +1568,20 @@ type_check_logic_op(Env, Op, P, Arg1, Arg2) ->
             throw({type_error, boolop, Op, P, Ty1});
         {true, Cs2} ->
             {Ty2, VB2, Cs3} = type_check_expr(Env#env{ venv = UnionVarBindsSecondArg(Env#env.venv,VB1 )}, Arg2),
-            case subtype(Ty2, {type, P, bool, []}, Env#env.tenv) of
+            % Allow any() in second argument for shortcut operators
+            SndArgTy = if Op == 'andalso'; Op == 'orelse' -> type(any);
```
zuiderkwast: This could perhaps be cleaner by breaking the shortcut ops out of the `type_check_logic_op` function and creating a separate `type_check_shortcut_ops`. Just an idea.

test/should_fail/shortcut_ops.erl
UlfNorell: I was hoping that after #59 gets merged we could work on checking that we get the right kind and number of errors from each `should_fail` and didn't want to spend work on splitting them up just to merge them back together again.

src/typechecker.erl
UlfNorell: Yes, I did that for `type_check_logic_op_in`. Might save that for future cleanups though.

PR #100, opened 2018-11-26
Fix forgotten variable bindings in case scrutinees

UlfNorell:
This is valid (albeit unusual) code:
```erlang
case X = bla() of _ -> foo() end,
X
```


josefs:
Nice catch!

PR #101, opened 2018-11-26
Fix some crashes

UlfNorell:
Missing cases in `handle_type_error` and crash in `negate_num_type` on `none()`.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3477,10 +3477,10 @@ handle_type_error({type_error, call, _P, Name, TyArgs, ArgTys}) ->
     io:format("The function ~p expects arguments of type~n~p~n but is given "
               "arguments of type~n~p~n",
               [Name, TyArgs, ArgTys]);
-handle_type_error({type_error, call, P, FunTy, Name}) ->
+handle_type_error({type_error, call, P, Ty, Name}) ->
     io:format("The function ~s, called on line ~p doesn't have a function type~n"
              "Rather, it has the following type~n~s~n"
-            ,[erl_pp:expr(Name), P, pp_intersection_type(FunTy)]);
+            ,[erl_pp:expr(Name), P, typelib:pp_type(Ty)]);
 handle_type_error({type_error, call_intersect, P, FunTy, Name}) ->
     io:format("The type of the function ~s, called on line ~p doesn't match "
               "the surrounding calling context.~n"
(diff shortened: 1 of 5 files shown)


UlfNorell:
Tests incoming

src/typechecker.erl
```diff
@@ -798,6 +798,8 @@ negate_num_type({type, P, union, Tys}) ->
     %% contain any unresolved types. So it is ok to normalize the result with an
     %% empty TEnv.
     normalize({type, P, union, [negate_num_type(Ty)||Ty <- Tys]}, #tenv{});
+negate_num_type(None = {type, _, none, []}) ->
+    None;
```
josefs: It could be argued that we should report a type error here instead. What's your argument for returning none() instead of failing?

UlfNorell: No strong reasons. This was the minimal fix to make the crash go away. The call to `negate_num_type` is here: https://github.com/josefs/Gradualizer/blob/d01371f7fa110654fa3e93f08dbf2a61241daa36/src/typechecker.erl#L1462-L1470
To throw an error we'd need to special-case `Ty = none()`, and do the same for the cases above for `+`, `bnot`, and `not`.

Since this was just a minor bugfix PR, I didn't want to change the typing rules.

gomoripeti: This is interesting. If `none()` comes from some type calculation maybe it should cause a type error the first time it is encountered (or maybe at the end of the flow), but probably not in the middle (ie not in negate).

Otoh if `none()` comes straight from a type spec that reflects user intention, so that should not count as a type error.

It would be good to differentiate between the two cases, maybe in the user-spec case the user should call it `no_return()` (so it would have a special meaning, not just an alias for `none()`, similarly as `term()` vs `any()` distinction. I think dialyzer makes this distinction too). And `none()` would be always a type error.

josefs: Right. For the sake of this PR I think it makes sense to leave the typing rules as is. 

@gomoripeti, your idea of differentiating `no_return()` and `none()` is interesting. But what if the programmer actually writes `none()` in a spec. What should the Gradualizer do then? I don't like the idea of rejecting the spec, there is legacy code which uses `none()`.

My sense is that it'd be good to get user feedback on this particular issue. It is not clear to me what the right type rule should be in the end. Leaving things as they are and erring on the side of being too lenient is the right thing to do right now.

PR #103, opened 2018-11-27
Maps are not contravariant in keys

UlfNorell:


OvermindDL1:
Maps can and probably should be thought of as row-typed records (unlike tagged tuples, which are like static records), when that is done and the typing follows standard HM-extended row typing resolvers then all the test cases should work properly, pass should pass and fail should fail.

UlfNorell:
> You've already mentioned that after errors are terms (+the test suite is updated), splitting should_pass/should_fail tests into multiple files should not be needed so I'm OK with going as is.

In this case there's only one type error in the file though, so nothing to split.

PR #109, opened 2018-12-11
 Basic type refinement and infer types inside patterns

### LINKED ISSUE #32: Handling aliases correctly
Currently the following code
```erlang
-spec test() -> {atom1, atom2} | atom3.
test() ->
    case f() of
        {atom1, _} = X -> X;
        _ -> atom3
    end.

-spec f() -> {atom1, atom2} | atom4.
f() ->
    {atom1, atom2}.
```
gives me the error
```
The variable 'X' on line 7 has type {atom1, atom2} | atom4 but is expected to have type {atom1, atom2} | atom3
```

### LINKED ISSUE #91: Basic type refinement
Given `getenv/1` from OTP's `os.erl`:

```erlang
-spec getenv(VarName) -> Value | false when
      VarName :: string(),
      Value :: string().

getenv(_) ->
    erlang:nif_error(undef).
```

The following code:

```erlang
-spec get_so_path() -> binary() | string().
get_so_path() ->
    case os:getenv("EJABBERD_SO_PATH") of
        false ->
            case code:priv_dir(mongooseim) of
                {error, _} ->
                    ".";
                Path ->
                    filename:join([Path, "lib"])
            end;
        Path ->
            Path %% this is line 77
    end.
```

Results in:

```
The variable 'Path' on line 77 has type Value | false but is expected to have type binary() | string()
```

Which seems to be incorrect, since the first clause of the `case` rules out `false`, therefore `Path` must be a `string()`. I have a hunch this might be connected with [this comment on type refinements](https://github.com/josefs/Gradualizer/issues/32#issuecomment-424234907), but my understanding of the theory is vague, to say the least :)

Surprisingly, the error persists in this case:

```erlang
-spec get_so_path() -> binary() | string().
get_so_path() ->
    case os:getenv("APP_SO_PATH") of
        false ->
            "/fake/default/path";
        Path ->
            Path
    end.
```

~But disappears when a binary literal is provided in the `false` clause:~
It did disappear due to Vim not being able to parse a Gradualizer crash message.

```erlang
-spec get_so_path() -> binary() | string().
get_so_path() ->
    case os:getenv("APP_SO_PATH") of
        false ->
            <<"/fake/default/path">>;
        Path ->
            Path
    end.
```

zuiderkwast:
Singleton integers, atoms and nil are used for type refinement. Recursively
type refinement in tuples. Only when there are no guards. Annotated types
are also handled.

Also limit upper bounds on "aliases" i.e. variables bound in 'match' patterns
on the form `Pattern1 = Pattern2` by inferring types from the patterns.

Two extra types are returned by add_type_pat/4.

Fixes #91
Fixes #32


src/typechecker.erl
```diff
@@ -2909,116 +3025,154 @@ type_check_function(Env, {function,_, Name, NArgs, Clauses}) ->
...
+    {Ty, Ty, VEnv, constraints:empty()};
+add_type_pat({var, P, A}=Var, Ty, TEnv, VEnv) ->
+    %% TODO: In a fun clause, A is always free, but not in e.g. case clauses
+    case VEnv of
+        #{A := VarTy} ->
+            case subtype(VarTy, Ty, TEnv) of
+                {true, Cs} ->
+                    {type(none), VarTy, VEnv, Cs};
+                false ->
+                    throw({type_error, tyVar, P, Var, VarTy, Ty})
```
gomoripeti: is it a type error in a case clause when Var :: a | b and Ty is a | c?
```
-spec f(a | b, a | c) -> boolean().
f(B,C) ->
  case C of B -> true
    _ -> false
  end.
```

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+refine_ty({ann_type, _, [_, Ty1]}, Ty2, TEnv) ->
+    refine_ty(Ty1, Ty2, TEnv);
+refine_ty(Ty1, {ann_type, _, [_, Ty2]}, TEnv) ->
+    refine_ty(Ty1, Ty2, TEnv);
+refine_ty(?type(T, A), ?type(T, A), _) ->
+    type(none);
+refine_ty({atom, _, A}, {atom, _, A}, _) ->
+    type(none);
+refine_ty(?type(list, A), ?type(nil), _TEnv) ->
+    type(nonempty_list, A);
```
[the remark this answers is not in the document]
UlfNorell: That's not true is it? I realise I'm not absolutely sure about the semantics of the improper list types (and gradualizer seems similarly confused):
```erlang
-spec foo() -> maybe_improper_list(a, b).
foo() -> [a | b].  %% The atom b does not have type maybe_improper_list(a, b)
```
I believe the semantics (type as set of values) is:
```erlang
maybe_improper(a, b) = { b, [a | b], [a, a | b], [a, a, a | b], ... }
nonempty_improper_list(a, b) = { [a | b], [a, a | b], ... }
```
In any case I don't think removing `[]` from a `maybe_improper_list` gives you a `nonempty_improper_list`.

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+refine_ty({Tag1, _, M} = Ty, {Tag2, _, N}, _TEnv)
+    when Tag1 == integer orelse Tag1 == char,
+         Tag2 == integer orelse Tag2 == char ->
+    if M == N -> type(none);
+       M /= N -> Ty
+    end;
+refine_ty(Ty, {Tag, _, V}, TEnv) when ?is_int_type(Ty),
+                                      Tag == integer orelse Tag == char ->
+    case Ty of
+        ?type(non_neg_integer) when V == 0 -> type(pos_integer);
```
UlfNorell: Missing test case for this line:
```erlang
-spec int_literals_3(non_neg_integer()) -> pos_integer().
int_literals_3(0) -> 42;
int_literals_3(N) -> N.
```

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+                    %% It's in the range
+                    Tys = int_range_to_types({Lo, V - 1}) ++
+                          int_range_to_types({V + 1, Hi}),
+                    normalize(type(union, Tys), TEnv);
+                true ->
+                    Ty
+            end;
+        _ ->
+            %% Not possible to subtract from other integer types
+            throw(no_refinement)
```
UlfNorell: Return `Ty`?

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+    end;
+refine_ty(Ty1, Ty2, TEnv) ->
+    case subtype(Ty2, Ty1, TEnv) of
+        false     -> Ty1;                 %% disjoint
+        {true, _} -> throw(no_refinement) %% inexact
+    end.
```
UlfNorell: Return `Ty1` in both cases? See comment about imprecision in integer types and `any()` above.

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+    if M == N -> type(none);
+       M /= N -> Ty
+    end;
+refine_ty(Ty, {Tag, _, V}, TEnv) when ?is_int_type(Ty),
+                                      Tag == integer orelse Tag == char ->
+    case Ty of
+        ?type(non_neg_integer) when V == 0 -> type(pos_integer);
+        ?type(non_neg_integer) when V < 0  -> Ty;
+        ?type(pos_integer)     when V =< 0 -> Ty;
+        ?type(neg_integer)     when V < 0  -> Ty;
```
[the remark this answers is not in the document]
UlfNorell: My suggestion would be to drop these cases completely and return `Ty` instead of `no_refinement` in the catch-all below. I see no reason to abort refinement when there's some imprecision (see comment on `any()` case above).

src/typechecker.erl
UlfNorell: I think that makes sense. Compare
```erlang
-spec f(a | b, a | c) -> boolean().
f(B, C) ->
  case C of
    X when X == B -> true;  %% The operator '==' on line 31 requires arguments of compatible types.
    _ -> false              %% However, it has arguments of type a | c and a | b
  end.
```

src/typechecker.erl
```diff
@@ -2909,116 +3025,154 @@ type_check_function(Env, {function,_, Name, NArgs, Clauses}) ->
...
-            add_type_pat(PT, ListTy, TEnv, VEnv2);
+            TailTy = normalize(type(union, [ListTy, type(nil)]), TEnv),
+            {_TailPatTy, _TauUBound, VEnv3, Cs} = add_type_pat(PT, TailTy, TEnv, VEnv2),
+            {type(none), ListTy, VEnv3, Cs};
         {elem_ty, ElemTy, Cs1} ->
-            {VEnv2, Cs2} = add_type_pat(PH, ElemTy, TEnv, VEnv),
-            {VEnv3, Cs3} = add_type_pat(PT, ListTy, TEnv, VEnv2),
-            {VEnv3, constraints:combine([Cs1, Cs2, Cs3])};
+            {_PatTy1, _UBound1, VEnv2, Cs2} = add_type_pat(PH, ElemTy, TEnv, VEnv),
+            TailTy = normalize(type(union, [ListTy, type(nil)]), TEnv),
```
UlfNorell: This seems wrong for improper lists. Or maybe not, see above that I'm not entirely sure of the improper list semantics.

src/typechecker.erl
```diff
@@ -2909,116 +3025,154 @@ type_check_function(Env, {function,_, Name, NArgs, Clauses}) ->
...
+            {_PatTy2, _Ubound2, VEnv3, Cs3} = add_type_pat(PT, TailTy, TEnv, VEnv2),
+            {type(none), ListTy, VEnv3, constraints:combine([Cs1, Cs2, Cs3])};
         {elem_tys, _ElemTys, Cs1} ->
             %% TODO: As a hack, we treat a union type as any, just to
             %% allow the program to type check.
             VEnv2 = add_any_types_pat(PH, VEnv),
-            {VEnv3, Cs2} = add_type_pat(PT, ListTy, TEnv, VEnv2),
-            {VEnv3, constraints:combine(Cs1, Cs2)};
+            TailTy = normalize(type(union, [ListTy, type(nil)]), TEnv),
```
UlfNorell: Same as above re `type(nil)`

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+        RefTy  -> RefTy
+    end.
+
+%% May throw no_refinement.
+refine_ty(?type(union, Tys), Ty, TEnv) ->
+    normalize(type(union, [refine(U, Ty, TEnv) || U <- Tys]), TEnv);
+refine_ty(?type(tuple, Tys1), ?type(tuple, Tys2), TEnv)
+  when length(Tys1) > 0 ->
+    %% Non-empty tuple
+    RefTys = [refine(Ty1, Ty2, TEnv) || {Ty1, Ty2} <- lists:zip(Tys1, Tys2)],
```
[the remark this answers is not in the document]
UlfNorell: Actually, that's not true. For union types we refine the alternatives. Example:
```erlang
-spec foo({a, b, c} | {a, b}) -> ok.
foo({a, b, c}) -> ok;
foo({a, b}) -> ok.
```

```
** exception error: no function clause matching lists:zip([],[{atom,21,c}]) (lists.erl, line 387)
     in function  lists:zip/2 (lists.erl, line 387)
     in call from lists:zip/2 (lists.erl, line 387)
     in call from typechecker:refine_ty/3 (../src/typechecker.erl, line 2943)
     in call from typechecker:refine/3 (../src/typechecker.erl, line 2932)
     in call from typechecker:'-refine_ty/3-lc$^0/1-3-'/3 (../src/typechecker.erl, line 2939)
     in call from typechecker:'-refine_ty/3-lc$^0/1-3-'/3 (../src/typechecker.erl, line 2939)
     in call from typechecker:refine_ty/3 (../src/typechecker.erl, line 2939)
```

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+    PossibleTyss = [case RefTy of
+                        Ty          -> [Ty];
+                        ?type(none) -> [Ty];
+                        _           -> [Ty, RefTy]
+                    end || {Ty, RefTy} <- lists:zip(Tys1, RefTys)],
+    [_|TuplesElems] = pick_from_each(PossibleTyss),
+    Tuples = [type(tuple, TupleElems) || TupleElems <- TuplesElems],
+    normalize(type(union, Tuples), TEnv);
+refine_ty(?type(any), _Ty, _) ->
+    throw(no_refinement);
```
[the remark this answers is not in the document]
UlfNorell: You may be right. I have to think on this.

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+  when length(Tys1) > 0 ->
+    %% Non-empty tuple
+    RefTys = [refine(Ty1, Ty2, TEnv) || {Ty1, Ty2} <- lists:zip(Tys1, Tys2)],
+    %% {a|b, a|b} \ {a,a} => {a|b, b} | {b, a|b}
+    PossibleTyss = [case RefTy of
+                        Ty          -> [Ty];
+                        ?type(none) -> [Ty];
+                        _           -> [Ty, RefTy]
+                    end || {Ty, RefTy} <- lists:zip(Tys1, RefTys)],
+    [_|TuplesElems] = pick_from_each(PossibleTyss),
```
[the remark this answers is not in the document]
zuiderkwast: Oh, that's true. We don't need to generate `{b, b}` in the results. I was mixing it up by expanding it to `{a, b}, {b, a}, {b, b}`, but that's not what we're doing. Good catch!

src/typechecker.erl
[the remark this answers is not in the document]
zuiderkwast: More relevant: If we remove `b` from `maybe_improper_list(a, b)` we get a proper `list(a)`, but I don't think we have to care about improper lists. Refinement isn't ever going to be complete anyway.

src/typechecker.erl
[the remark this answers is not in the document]
zuiderkwast: It's a typo for sure Peter. Good catch. About Ulf's suggesion, I'm not sure yet.

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
...
+    case subtype(Ty2, Ty1, TEnv) of
+        false     -> Ty1;                 %% disjoint
+        {true, _} -> throw(no_refinement) %% inexact
+    end.
+
+%% [Xs, Ys, ...] -> [[X, Y, ...] || X <- Xs, Y <- Ys, ...]
+pick_from_each([]) -> [];
+pick_from_each([Xs]) ->
+    [[X] || X <- Xs];
+pick_from_each([Xs|Xss]) ->
```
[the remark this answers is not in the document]
zuiderkwast: Good points, both of them. I'll replace it with another function.

src/typechecker.erl
zuiderkwast: Good point. I'll add that. I'll look at the coverage report to see I've missed other cases as well.

src/typechecker.erl
zuiderkwast: Improper lists aren't that complicated. These improper list types exist (apart from aliases):

* `maybe_improper_list(Type1, Type2)`
* `nonempty_improper_list(Type1, Type2)`
* `nonempty_maybe_improper_list(Type1, Type2)`

There is no (possibly empty) `improper_list(Type1, Type2)` because "empty" doesn't make sense for this one. The tail of a `nonempty_improper_list(Type1, Type2)` is either a `nonempty_improper_list(Type1, Type2)` or a `Type2`.

zuiderkwast: Actually, I didn't consider improper lists. I should exclude them to be on the safe side.

src/typechecker.erl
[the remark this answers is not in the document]
UlfNorell: I have done a bit more thinking (but maybe not enough). I believe the problem is the `Ty` case in refinement of tuples:
```erlang
    PossibleTyss = [case RefTy of
                        Ty          -> [Ty];
                        ?type(none) -> [Ty];
                        _           -> [Ty, RefTy]
                    end || {Ty, RefTy} <- lists:zip(Tys1, RefTys)],
```
Note that this is the same as the case for `none()`. That is, no refinement is treated the same as complete covering, which surely can't be right. I think this case should simply be removed:
```erlang
    PossibleTyss = [case RefTy of
                        ?type(none) -> [Ty];
                        _           -> [Ty, RefTy]
                    end || {Ty, RefTy} <- lists:zip(Tys1, RefTys)],
```
There really shouldn't have to be a special case for a no-op refinement, and needing one is a huge red flag that there's something iffy (what if the refined type is semantically the same but not syntactically?). With this change, and the change to `pick_from_each`, I believe imprecise refinements can be handled by simply returning the original type.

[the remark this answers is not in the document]
zuiderkwast: That seems right. But if `Ty == RefTy` for one of the types, say for example `RefTy2==Ty2`, and in the result union exactly one Ty replaced by RefTy so that in this example, the tuple `{Ty1, RefTy2, Ty3, ...}` is included in the result union, the result is the original tuple. This means there is no refinement. Therefore, we can as well have the `throw(no_refinement)` in various places, to abort refinement.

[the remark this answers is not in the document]
zuiderkwast: (I wanted to simplify `[Ty, Ty]` to `[Ty]` just to limit the 2^N explosion when possible, but now it won't be 2^N anyway.)

src/typechecker.erl
UlfNorell: Using `subtype` here is probably fine, but for subtle reasons: `Ty2` will always be very precise, so that if `Ty2` is not a subtype of `Ty1` then they are disjoint. I think it would be more clear, and more future proof, to check if `glb(Ty1, Ty2)` returns `none()`.

src/typechecker.erl
UlfNorell: I filed #110 for my improper list thoughts.

src/typechecker.erl
zuiderkwast: Sure. I assumed `glb` isn't complete yet. I'll use it if it works; otherwise add a TODO comment.

UlfNorell: I believe it should work for the types you generate here, but please test 😄.

src/typechecker.erl
```diff
@@ -2890,6 +2906,106 @@ check_clause(_Env, _ArgsTy, _ResTy, Term) ->
     io:format("DEBUG: check_clause term: ~p~n", [Term]),
     throw(check_clause).
 
+%% Refine types by matching clause. MatchedTys are the types exhausted by
+%% each pattern in the previous clause.
+refine_clause_arg_tys(Tys, MatchedTys, [], TEnv) ->
```
josefs: Why do you insist on having no guards when refining a type? I assume you don't want to deal with refining types based on predicates like `is_integer/1` right now. But couldn't you just ignore these types of guards and refine the type in the pattern anyway?

zuiderkwast: To be able to refine the types for the next clause, we need to know exactly why the previous clause didn't match. If we have any guard, we can't be sure that we are at the next clause because of the pattern mismatched, because it may be that the clause mismatched because of the guard alone.

```Erlang
-spec foo(a|b, boolean()) -> b.
foo(A, Guard) ->
    case A of
        a when Guard -> b;
        X            -> X  %% <--- may still be a|b, when Guard is false
    end.
```

zuiderkwast:
@josefs 
> How would you characterize when type refinement happens? If a programmer came and ask you what situations type refinement happens and when it doesn't, what's your response? I think it's important to have an answer that's simple and easy to understand and I get the sense you might have just such an answer.

Short answer: *We can refine only if we know exactly why the previous clause mismatches.*

If there is any imprecision, we cannot refine. If there is guard with already bound variables, we cannot refine. See the test cases added under `should_fail` for some cases where we can't refine. See also my answer to your question about guards.

For guards, I am considering adding refinement if the only guard is `is_atom(X)` (and similar functions) where X is bound in the pattern. I think the guards have to be passed around in add_type_pat all the way to the place where X is first bound in order to be able to correctly tie X to exactly (not a subtype of) `atom()`.

src/typechecker.erl
josefs: Ah, yes. Indeed.

josefs:
> Short answer: We can refine only if we know exactly why the previous clause mismatches.

This is a start, but it begs the question: when do we know exactly why the previous clause mismatches?

The kind of characterization I'm looking for is something along the lines: we refine all patterns of type integer, atoms, tuples and unions.

Your plan for guards sounds good.

src/typechecker.erl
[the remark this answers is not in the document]
zuiderkwast: I think removing the clause which prevents the refinement `any() \ any() => none()` may solve your `fails` example above. The second `any()` doesn't actually mean *unknown* in this case. In this case, it's just `Ty1 == Ty2` (which means that the type is exhausted by the pattern).

zuiderkwast:
I definitely agree that an exact description is necessary. I think we should start to add documentation as a part of each new feature, or at least in directly afterwards, to a manual in the wiki I suggest. Pushing for this is necessary. Documentation doesn't come by itself.....

src/typechecker.erl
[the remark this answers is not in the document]
zuiderkwast: Yep, it works.

src/typechecker.erl
zuiderkwast: Using glb works.

src/typechecker.erl
```diff
@@ -2987,24 +2982,42 @@ refine_ty(Ty, {Tag, _, V}, TEnv) when ?is_int_type(Ty),
...
+%%
+%% If RefTyI==TyI or RefTyI==none() for any I, that list I is excluded.
+pick_one_refinement_each([], []) -> [];
+pick_one_refinement_each([Ty|Tys], [RefTy|RefTys]) ->
+    %% The lists (zero or one list) where we refine head and keep tail
+    %% unrefined, if head is possible to refine.
+    RefHeadCombinations =
+        case RefTy of
+            Ty          -> [];           %% pattern mismatches; no refinement
+            ?type(none) -> [];           %% pattern matches all; no refinement
```
UlfNorell: I'm still not happy with disjoint pattern and complete match being treated the same. With this implementation you get:
```erlang
refine({a, b | c}, {b, c}) -> {a, b}.
```
since the mismatch of the first component is treated as a complete match. Now `{b, c}` isn't a well-typed pattern of type `{a, b | c}`, but from this you get
```erlang
refine({a, b | c} | {b, c}, {b, c}) -> {a, b}.
```
which is clearly wrong.

Currently I don't think you can get there because of https://github.com/josefs/Gradualizer/blob/9fa405f1dd470c4f0effc269379e8b14c7dee6ef/src/typechecker.erl#L3135
But if that TODO is addressed we would get incorrect refinements.

I think the right thing to do is to skip this case, i.e. fall through to the `[[RefTy|Tys]]` case. That would give us
```erlang
refine({a, b | c}, {b, c}) -> {a, b | c} | {a, b}.
```
which is correct (although the `{a, b}` component looks a little weird).

For products of sets we have A₁ × B₁ - A₂ × B₂ = (A₁ - A₂) × B₁ ∪ A₁ × (B₁ - B₂), or visually:

```
           A₂
        +------+
        |      |
   +----|--+   | B₂ 
   |////|  |   |
B₁ |////+--+---+
   |XXXX\\\|
   +-------+
       A₁
```

In the code the `none()` case is simply an optimisation for when A₁ - A₂ = ∅, and it would be perfectly fine to always return `[[RefTy|Tys]]`.

zuiderkwast: Thanks @UlfNorell for the thorough thoughts! I'll try removing that clause and see if the tests still pass. I'll also add tests for the examples you mention here. I can do this tonight, otherwise in the weekend.

It would be very nice if someone (you?😉) wants to address the `%% TODO: This code approximates unions of tuples with tuples of unions`. It would be a relief to get rid of those.

@josefs I've added a wiki page about [Type refinement](https://github.com/josefs/Gradualizer/wiki/Type-refinement). I didn't put much effort in it, but it's something. Feel free to comment on it here or otherwise or edit it yourself.

josefs: The wiki page is great!

zuiderkwast: I've been processing the disjoint pattern case in my sleep.

If a pattern inside a (nested) tuple pattern can't match, it means the whole tuple pattern can't match. Therefore, I think the right thing to do is a non-local return to an outer union, where the refinement is applied. If there is no union, the exception can be caught in `type_check_clause` and a message "this clause can never match" can be reported.

zuiderkwast: Disjoint is now thrown upward (caught where a union is refined) so at least that case is gone. Tuple union refinement now behaves nicely, e.g. `refine({a, b | c}, {b, c}) -> {a, b | c}.`

If the whole thing is refined to `none()`, it means the next clause is unreachable, so we have unreachable clause detection implemented as well as a bonus.

zuiderkwast:
I'm ready to merge this unless there are more comments.

PR #112, opened 2018-12-18
Return error terms

Zalastax:
Part of an upcoming series of PR's replacing #59.
The discussion over there has been great but doing it all in one big bang is not within my reach.
By keeping working on #59 I believe that I'm hindering progress since others might be avoiding changing things in this area.

This PR contains the changes needed to return errors as terms - the change I first set out to do.
I've added a comment cautioning others from depending on the format of these terms - I hope that will be enough of a deterrent.

### DIFF
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -6,6 +6,10 @@
 %%%   in the given directory.
 %%% - `print_file': if `true' prefix error printouts with the file name the
 %%%   error is from.
+%%% - `crash_on_error`: if `true` crash on the first produced error
+%%% - `return_errors`: if `true`, turns of error printing and errors
+%%%   (in their internal format) are returned in a list instead of being
+%%%   condensed into a single ok | nok.
 -module(gradualizer).
 
 -export([type_check_file/1,
@@ -25,12 +29,12 @@
 %% API functions
 
 %% @doc Type check a source or beam file
--spec type_check_file(file:filename()) -> ok | nok.
+-spec type_check_file(file:filename()) -> ok | nok | [{file:filename(), any()}].
 type_check_file(File) ->
     type_check_file(File, []).
 
 %% @doc Type check a source or beam file
--spec type_check_file(file:filename(), options()) -> ok | nok.
+-spec type_check_file(file:filename(), options()) -> ok | nok | [{file:filename(), any()}].
 type_check_file(File, Opts) ->
     ParsedFile =
         case filename:extension(File) of
@@ -44,19 +48,29 @@ type_check_file(File, Opts) ->
     case ParsedFile of
         {ok, Forms} ->
             Opts2 = proplists:expand([{print_file, [{print_file, File}]}], Opts),
-            typechecker:type_check_forms(Forms, Opts2);
+            ReturnErrors = proplists:get_bool(return_errors, Opts),
+            Errors = typechecker:type_check_forms(Forms, Opts2),
+            case {ReturnErrors, Errors} of
+                {true, _ } ->
+                    lists:map(fun(Error) -> {File, Error} end, Errors);
+                {false, []} ->
+                    ok;
+                {false, [_|_]} ->
+                    nok
+            end;
         Error ->
             throw(Error)
     end.
 
 
 %% @doc Type check a module
--spec type_check_module(module()) -> ok | nok.
+-spec type_check_module(module()) -> ok | nok | [{file:filename(), any()}].
 type_check_module(Module) ->
     type_check_module(Module, []).
 
 %% @doc Type check a module
--spec type_check_module(module(), options()) -> ok | nok.
+-spec type_check_module(module(), options()) ->
+                               ok | nok | [{file:filename(), any()}].
 type_check_module(Module, Opts) when is_atom(Module) ->
     case code:which(Module) of
         File when is_list(File) ->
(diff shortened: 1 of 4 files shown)


src/gradualizer.erl
```diff
@@ -25,12 +31,12 @@
 %% API functions
 
 %% @doc Type check a source or beam file
--spec type_check_file(file:filename()) -> ok | nok.
+-spec type_check_file(file:filename()) -> [{file:filename(), any()}].
```
zuiderkwast: Is there no way of *not* returning the errors? If they are printed, they don't have to be returned, do they? The terminal will be flooded when running in an Erlang shell. (I usually do that.)

Zalastax:
This is the only part that contains code that I can salvage from #59. Merge this later if you want.

src/gradualizer.erl
Zalastax: Would you be ok with a wrapper,`gradualizer:checked_ok/1`?

```Erlang
checked_ok(Errors) -> length(Errors) =:= 0.
```

I'd prefer to not add an option since that makes for strange types and most likely a more complicated implementation.

src/gradualizer.erl
```diff
@@ -6,6 +6,12 @@
 %%%   in the given directory.
 %%% - `print_file': if `true' prefix error printouts with the file name the
 %%%   error is from.
+%%% - `crash_on_error`: if `true` crash on the first produced error
```
josefs: Perhaps this option should be called `crash_on_first_error`.
Also, I think the word "crash" might be a bit imprecise. A suggestion would be "exception" which is more accurate.

Zalastax: I'm not introducing this option, it was just not documented. I can split this change into a separate PR that also includes renaming it.

zuiderkwast: I added it for debugging purposes some months ago. I wasn't sure if it would be better to leave it undocumented. I didn't include the word `first` because there is obviously only one crash... By crash, I meant the same behaviour as for an unexpected programming error (case_clause, etc.). I needed the stack trace.

I don't know if exception is a better name. Less precise perhaps. An exception (throw/error/exit) can be caught and the checking can continue, as we normally do with type errors.

Perhaps we can delete `crash_on_error` when the typechecker is becoming more stable. I don't think it is stable enough yet though. I have seen gradualizer hang, so yesterday (unpushed) I added a `debug` option for printing progress stuff like "Checking X :: Y" while debugging, to be able to locate where it hangs.

The point is that it might be good to have options for debugging the gradualizer and these should probably be either undocumented or marked in the documentation as "you probably don't want to use this; it's for debugging the Gradualizer".

Zalastax: Should I just remove this comment then and instead add a comment in `typechecker`  about removing the option when it's no longer needed?

zuiderkwast: I don't care.

(It's good to avoid unrelated changes in the same PR. People tend to comment on them. :-))

josefs: Feel free to resolve this conversation. I agree that changing the option would be an unrelated change and we shouldn't do it as part of this PR.

src/gradualizer.erl
```diff
@@ -6,6 +6,10 @@
...
+%%% - `no_report_errors`: if `true` don't print type checking error messages
+%%% - `return_errors`: if `true`, errors (in their internal format) are returned
```
gomoripeti: what if it would not be possible to both print and return errors? (or does anyone want to do that?) just a single option `return_errors`, when false, errors are printed, as by default. (dialyzer has a similar option `--raw`)

zuiderkwast: I agree with Péter that *report* can just be the negation of *return*. No separate option needed IMO.

If we do have both options, I think the non-negated `report_errors` is better than `no_report_errors`. The default value for *report* can be the negation of *return*, so that you don't get both unless you specify both.

zuiderkwast:
Thanks for adding the possibility to not return errors. :-) I'm almost happy now.

(Whether or not `nok` is a good atom is a future topic. The erlang compiler returns `error` or  `{error, Errors, Warnings}` depending on whether they should be reported or not. I somehow like `nok` though. It's funny.)

Zalastax:
@zuiderkwast what's left to make you really happy?
>  I'm almost happy now.

src/typechecker.erl
```diff
@@ -3510,7 +3510,7 @@ type_check_forms(Forms, Opts) ->
     StopOnFirstError = proplists:get_bool(stop_on_first_error, Opts),
     CrashOnError = proplists:get_bool(crash_on_error, Opts),
     File = proplists:get_value(print_file, Opts),
-    NoReportErrors = proplists:get_bool(no_report_errors, Opts),
+    NoReportErrors = not(proplists:get_bool(return_errors, Opts)),
```
zuiderkwast: This is double negation, isn't it? It's the same as ReportErrors = ReturnErrors.

zuiderkwast:
> what's left to make you really happy?

@Zalastax It was the same as Péter's comment. I was going to say I'm happy now, but I think the logic is double negated for report and return, so you get both or nothing.

src/typechecker.erl
Zalastax: Haha, yes you're right!

PR #114, opened 2018-12-19
Unify errors 1

Zalastax:


src/typechecker.erl
```diff
@@ -1951,7 +1951,7 @@ do_type_check_expr_in(Env, Ty, Char = {char, LINE, _}) ->
         {true, Cs} ->
            {#{}, Cs};
        false ->
-           throw({type_error, char_expr, LINE, Char, Ty})
+           throw({type_error, Char, {type, erl_anno:new(0), char, []}, Ty})
```
gomoripeti: nitpicking: could use `type(char)` (and `type(list)` below) for brevity and consistency.
Also `LINE` in the function head seems to be unused

Zalastax: Absolutely, thanks!

src/typechecker.erl
```diff
@@ -1968,7 +1968,7 @@ do_type_check_expr_in(Env, Ty, Cons = {cons, LINE, H, T}) ->
             {     VB2, Cs2} = type_check_expr_in(Env, allow_empty_list(Ty), T),
             {union_var_binds(VB1, VB2, Env#env.tenv), constraints:combine(Cs1, Cs2)};
         {type_error, _} ->
-            throw({type_error, cons, LINE, Cons, Ty})
+            throw({type_error, Cons, type(list), Ty})
     end;
```
zuiderkwast: The actual type is nonempty_list I think. Otherwise, checking a cons against the type [] will give a strange error message.

Zalastax: Oh, indeed, I'll fix that later today!

PR #116, opened 2018-12-20
Unify errors for char nil, string, bin, tuple, and map

Zalastax:

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1964,12 +1964,12 @@ do_type_check_expr_in(Env, Ty, Cons = {cons, _, H, T}) ->
         {type_error, _} ->
             throw({type_error, Cons, type(nonempty_list), Ty})
     end;
-do_type_check_expr_in(Env, Ty, {nil, LINE}) ->
-    case subtype({type, LINE, nil, []}, Ty, Env#env.tenv) of
+do_type_check_expr_in(Env, Ty, {nil, _} = Nil) ->
+    case subtype(type(nil), Ty, Env#env.tenv) of
         {true, Cs} ->
             {#{}, Cs};
         false ->
-            throw({type_error, nil, LINE, Ty})
+            throw({type_error, Nil, type(nil), Ty})
     end;
 do_type_check_expr_in(Env, Ty, {string, LINE, String}) ->
     case subtype({type, LINE, string, []}, Ty, Env#env.tenv) of
@@ -3761,9 +3761,13 @@ handle_type_error({type_error, {cons, Anno, _, _} = Cons, ActualTy, ExpectedTy})
                erl_anno:line(Anno),
                typelib:pp_type(ExpectedTy),
                typelib:pp_type(ActualTy)]);
-handle_type_error({type_error, nil, LINE, Ty}) ->
-    io:format("The empty list on line ~p does not have type ~s~n",
-              [LINE, typelib:pp_type(Ty)]);
+handle_type_error({type_error, {nil, Anno} = Nil, ActualTy, ExpectedTy}) ->
+    io:format("The empty list ~s on line ~p is expected "
+              "to have type ~s but it has type ~s~n",
+              [erl_pp:expr(Nil),
+               erl_anno:line(Anno),
+               typelib:pp_type(ExpectedTy),
+               typelib:pp_type(ActualTy)]);
 handle_type_error({argument_length_mismatch, P, LenTy, LenArgs}) ->
     io:format("The clause on line ~p is expected to have ~p argument(s) "
               "but it has ~p~n ",
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -2010,16 +2010,16 @@ do_type_check_expr_in(Env, ResTy, {tuple, LINE, TS}) ->
     end;
 
 %% Maps
-do_type_check_expr_in(Env, ResTy, {map, LINE, Assocs}) ->
-    case subtype(ResTy, {type, LINE, map, any}, Env#env.tenv) of
+do_type_check_expr_in(Env, ResTy, {map, _, Assocs} = Map) ->
+    case subtype(ResTy, type(map, any), Env#env.tenv) of
         {true, Cs1} ->
             %% TODO: check the type of the map fields
             {_AssocTy, VBs, Cs2} = type_check_assocs(Env, Assocs),
             {VBs, constraints:combine(Cs1, Cs2)};
         false ->
-            throw({type_error, map, LINE, ResTy})
+            throw({type_error, Map, type(map, any), ResTy})
     end;
-do_type_check_expr_in(Env, ResTy, {map, LINE, Expr, Assocs}) ->
+do_type_check_expr_in(Env, ResTy, {map, _, Expr, Assocs} = Map) ->
     {Ty, VBExpr,   Cs1} = type_check_expr(Env, Expr),
     {_AssocTy, VBAssocs, Cs2} = type_check_assocs(Env, Assocs),
     % TODO: Update the type of the map.
@@ -2029,7 +2029,7 @@ do_type_check_expr_in(Env, ResTy, {map, LINE, Expr, Assocs}) ->
             {union_var_binds(VBExpr, VBAssocs, Env#env.tenv),
              constraints:combine([Cs1, Cs2, Cs3])};
         false ->
-            throw({type_error, map, LINE, ResTy})
+            throw({type_error, Map, Ty, ResTy})
     end;
 
 %% Records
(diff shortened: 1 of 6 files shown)


src/typechecker.erl
```diff
@@ -2003,7 +2003,7 @@ do_type_check_expr_in(Env, ResTy, {tuple, LINE, TS}) ->
         {elem_tys, Tyss, Cs} ->
             case type_check_tuple_union_in(Env, Tyss, TS) of
                 none ->
-                    throw({type_error, tuple, LINE, ResTy});
+                    throw({type_error, Tuple, type(tuple, TS), ResTy});
```
gomoripeti: I think `TS` is a list of expressions, not a list of types!

src/typechecker.erl
```diff
@@ -1970,31 +1970,31 @@ do_type_check_expr_in(Env, Ty, Cons = {cons, _, H, T}) ->
...
-    case subtype({type, LINE, string, []}, Ty, Env#env.tenv) of
+do_type_check_expr_in(Env, Ty, {string, _, _} = String) ->
+    case subtype(type(string), Ty, Env#env.tenv) of
       {true, Cs} ->
         {#{}, Cs};
       false ->
-        throw({type_error, string, LINE, String, Ty})
+        throw({type_error, String, type(string), Ty})
     end;
 do_type_check_expr_in(Env, Ty, {bin, LINE, _BinElements} = Bin) ->
```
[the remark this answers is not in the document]
Zalastax: Aaaah, thanks!

src/typechecker.erl
Zalastax: That's very possible - I think I tested with atoms and they share type and expression representation, right? Is it `Tyss?` that I should use instead?

zuiderkwast: Tyss is a `[[type()]]`.

gomoripeti: I was not sure either but maybe it should be `type(union, [type(tuple, Tys)||Tys <- Tyss])`?

josefs: Indeed, Tyss is a list representing a union of all the possible element types of the tuple. For example, `[[a,integer()],[b,string()]]` would mean that we have a two element tuple which can either have the type `{a, integer()}` or `{b, string()}`.

zuiderkwast: Right, but what should reported as the 3rd element in the error is the actual type of Tuple. Péter's type (Tyss as a union of tuples) is a subset the ResTy (the union elements that are tuples of the same size as Tuple) is not what should be reported here.

We have to go further into `type_check_tuple_union_in` to see why none of the Tyss are matching.

zuiderkwast: What this means is that none of the types in the union are matching. In `type_check_tuple_union_in`, none is returned if type checking against each of the tuple types give a type error. We could catch one of these, extract the actual type if it's present and then use that here. This requires that the actual type is part of all possible type errors, in the general case.

Perhaps we don't have to include an actual type in this case?

Zalastax: Perhaps it's modus operandi, but to me it feels strange that we mix type inference with type checking (that is what we're doing here, right?).
My intuition is that one should first infer the type (possibly lazily) and secondly check for compatibility. In that case, typechecking would always have an actual type and an expected type.
I suspect that there's a very good reason why that's not actually a good idea. Nevertheless, in this case shouldn't we be able to report the actual type?

zuiderkwast: In this part of the code, it's not mixing type inference with checking. It is checking only.

Checking against the union type is done by checking against each of the types in the union. No type is returned, just exceptions if there is a type error. A tuple may mismatch each of the types in the union for different reasons.

Here is what happens in, approximately:

* Check that `{a, 1} :: {b, 1} | {a, 2}`. (`Tyss = [[b, 1], [a, 2]]` here.)
    * Check `{a, 1} :: {b, 1}`
        * Check that `a :: b` --> Type error "a is expected to have type b but it has type a"
    * Catch type error and try the next type in the union: Check that `{a, 1} :: {a, 2}`
        * Check that `a :: a` -- OK, check next type element
        * Check that `1 :: 2` --> Type error "1 is expected to have type 2 but it has type 1"
    * Catch the type error. All of the types in the union gives type errors, so the function returns `none`.

So, the problem is that we don't have an actual type of the tuple.

It would be weird to use inference only to get something to include in the type error.

If all the type errors would contain the actual type, we could take it from any of the caught type errors, but as long as they don't we can't be sure to find it.

Are you following?

gomoripeti: so strictly speaking during type checking there shouldn't be an actual type ever (so that field of the tuple can be undefined). So we can say that `the tuple {a, 1} is expected to have type {b, 1} | {a, 2}` and leave it to the user. (Dialyzer also doesn't provide more info) (It does not give any more info anyway that the tuple expression has the type tuple. Same for char, string etc) 

Or we can try to refine the error (I don't even know how to put the sentence to be more helpful `the tuple {a, 1} is expected to have either its first element of type b or its second element 2`) but this can get very complicated soon.
However there is still some kind of "inference" if there is a variable or function call involved in the expression, in these cases it would be still useful to print what Gradualizer thinks are the types of these sub-exrpessions (because it may be not obvious for the user). Maybe (future improvement idea) the type of involved variables and function calls could be added to the error (eg: `the tuple {A, 1, f()} is expected to have type {b, 1, 1} | {a, 2, 2} where A has type a and f() has type 1 | 2`).

zuiderkwast: How about allowing two variants of expression type error `{type_error, Expr, ActualType, ExpectedType}` and `{type_error, Expr, ExpectedType}` for now? It's a simple way forward.

Zalastax: The simple way forward is indeed to not report an actual type wherever the current code structure makes doing so difficult. As @gomoripeti points out, there is an "inference" going on which translates an expression to it's most specific type. I'd like that translation to happen in a step that is not mixed with checking compatibility so that the actual type can be reported.

E.g. if `B :: boolean()` and we're checking `{a, B} :: {b, boolean()} | {a, 2}`, we'd start by "infering" `{a, B} :: {a, boolean()}` and then check if `{a, boolean()} <= {b, boolean()}` or `{a, boolean()} <= {a, 2}`.

Can all of our type checking be structured as
1. Infer most specific type `ActualType`
2. Compare against `ExpectedType`
?

zuiderkwast: This is about bidirectional type checking. Normally, we prefer to push a type inward rather than to infer, because it is supposed to give better precision in the type errors. I'm not sure this is true here though when we loop over types to find one that doesn't give a type error.

gomoripeti: how about going forward with this PR like this:
- put `undefined` in the actual type in case of type checking (I think it's better than the 3-tuple, because `handle_type_error` can match on the 4 tuple (and it can decide to not print the actual type if it is undefined - utility function candidate))
(one potential exception: in case of binaries `BinTy = gradualizer_bin:compute_type(Bin),` really looks like type inference. Also in case of map update)

move further improvements to follow-up PRs:
  - how to report error in failed checking of union
  - where do we use bidirectional type checking and how to report errors there
  - when error term will be a map we can consider adding optional contextual fields (var types, specs)

zuiderkwast: I think that tuple unions, list unions and similar union handling in the `expect_*_type` functions are not needed. Instead, we can handle all unions in the same way. We already do that in `add_type_pat/4`. It will be less code and also it will be more obvious that this kind of type error *"the expression E doesn't have any of the types in the union T1|...|Tn"* is union specific. It is not tuple specific.

I would leave this one out of this PR if I were you, to be able to merge it quickly.

src/typechecker.erl
```diff
@@ -2012,20 +2012,20 @@ do_type_check_expr_in(Env, ResTy, {tuple, LINE, TS}) ->
...
         {type_error, _} ->
-            throw({type_error, tuple, LINE, ResTy})
+            throw({type_error, Tuple, type(tuple, any), ResTy})
     end;
 
 %% Maps
-do_type_check_expr_in(Env, ResTy, {map, LINE, Assocs}) ->
-    case subtype(ResTy, {type, LINE, map, any}, Env#env.tenv) of
+do_type_check_expr_in(Env, ResTy, {map, _, Assocs} = Map) ->
+    case subtype(ResTy, type(map, any), Env#env.tenv) of
```
gomoripeti: this was wrong in the original code: should be `map() :: ResTy` instead of `ResTy :: map()` (I think `record_field` and `record_index` has similar flip issue)

Zalastax: I don't quite follow, could you spell out what the code should say?

gomoripeti: What I meant is `subtype(ResTy, type(map, any),...` checks whether `ResTy` is a subtype of `map()`, but we should rather check that `map()` is a subtype of `ResTy` (ie `subtype(, type(map, any), ResTy,...`)

UlfNorell: I'm afraid neither is correct. Consider
```erlang
-spec f() -> #{ a := b } | undefined.
f() -> #{ a => b }
```
We have neither `map() :: #{a := b} | undefined` or the other way around. Instead we need a `expect_map_type` or similar (perhaps taking the `glb` of `map()` and `ResTy` first?) to dig out the expected types for the assocs. But this isn't really the PR for that... I've filed #131.

Zalastax:
Since we're not fixing the map error, is this ready to be merged?

zuiderkwast:
> Since we're not fixing the map error, is this ready to be merged?

Ah, so what's changed now? Have you skipped tuples as well? Sure then, merge it!

src/typechecker.erl
zuiderkwast: Hey, we do have `map() :: #{a := b} | undefined`. Of course it's not complete as we don't check the keys and values, but `map()` is the any-map so it's a compatible subtype to `#{a := b}`.

UlfNorell: Oh, you're right. I still get an error on the example though... mysterious.

PR #117, opened 2018-12-21
Various bug smashing

zuiderkwast:
These commits are really independent of each other so I recommend reviewing them one by one.

The `debug` option was added just for debugging. We don't need it, but it was really useful for localizing where the program hanged. (Turned out to be in `union_var_binds`.)

These bugs were found by running the gradualizer on itself.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3503,11 +3503,20 @@ is_power_of_two(_) -> false.
 union_var_binds(VB1, VB2, TEnv) ->
     union_var_binds([VB1, VB2], TEnv).
 
+%% This function has been identified as a bottleneck.
+%% Without tail recursion, the gradualizer would hang when self-gradualizing
+%% when called from add_type_pat/4, the clause where the type is a union type.
 union_var_binds([], _) ->
     #{};
-union_var_binds([ VarBinds | VarBindsList ], TEnv) ->
+union_var_binds(VarBindsList, TEnv) ->
     Glb = fun(_K, S, T) -> glb(S, T, TEnv) end,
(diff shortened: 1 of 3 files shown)


src/typechecker.erl
```diff
@@ -1905,6 +1906,9 @@ type_check_comprehension(Env, Compr, Expr, [Guard | Quals]) ->
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 type_check_expr_in(Env, ResTy, Expr) ->
+    Env#env.debug andalso
+        io:format("~p: Checking that ~s :: ~s~n",
+                  [element(2, Expr), erl_pp:expr(Expr), typelib:pp_type(ResTy)]),
```
UlfNorell:
```erlang
debug(Env, Fmt, Args) ->
   Env#env.debug andalso io:format(Fmt, Args).
```
Please.

src/typechecker.erl
```diff
@@ -2948,7 +2960,14 @@ refine_clause_arg_tys(Tys, MatchedTys, [], TEnv) ->
...
     catch
-        no_refinement -> Tys
+        no_refinement ->
+            %% Imprecision prohibits refinement
+            Tys;
+        disjoint ->
+            %% This can currently happen due to unhandled type variables, e.g.
+            %% Elem :: T \ {attribute, _TyVar-54982374928, compile, export_all}
+            %% No refinement.
+            Tys
```
UlfNorell: Can you add a test case for this?

zuiderkwast: Sure, I will do that.

src/typechecker.erl
zuiderkwast: Sure! Good point.

PR #124, opened 2018-12-27
Support char as singleton integer type

### LINKED ISSUE #115: Support char as singleton integer type
Character literals are allowed everywhere where integers are allowed in types (at least in more recent Erlang versions). But Gradualizer does not support this yet (not in `?is_int_type/1` macro, or in `int_type_to_range/1`, etc.).

eg. this is a valid spec which should pass
```
-spec char_range() -> $a..$z.
char_range() ->
    $c.
```

gomoripeti:
(The added test is not too exhaustive, but it covers a good amount of
the changed lines)

Fixes #115

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -18,7 +18,10 @@
 
 -export_type([typed_record_field/0]).
 
--type type() :: erl_parse:abstract_type().
+-type type() :: erl_parse:abstract_type()
+              | af_character(). %% Workaround: char was not part of type() until after OTP 21.2.1
+                                %% (although it is supported since OTP 19.3)
+-type af_character() :: {'char', erl_anno:anno(), char()}.
 
 %% Pattern macros
 -define(type(T), {type, _, T, []}).
@@ -672,6 +675,7 @@ is_int_type({type, _, T, _})
   when T == pos_integer; T == non_neg_integer; T == neg_integer;
        T == integer; T == range -> true;
 is_int_type({integer, _, _}) -> true;
+is_int_type({char, _, _}) -> true;
 is_int_type(_) -> false.
 
 %% A type used while normalizing integer types. The ranges that are possible to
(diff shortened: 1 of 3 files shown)


src/typechecker.erl
```diff
@@ -18,7 +18,10 @@
 
 -export_type([typed_record_field/0]).
 
--type type() :: erl_parse:abstract_type().
+-type type() :: erl_parse:abstract_type()
+              | af_character(). %% Workaround: char was not part of type() until after OTP 21.2.1
```
gomoripeti: see https://github.com/erlang/otp/pull/2074

src/typechecker.erl
```diff
@@ -747,9 +751,11 @@ int_type_to_range({type, _, integer, []})              -> {neg_inf, pos_inf};
 int_type_to_range({type, _, neg_integer, []})          -> {neg_inf, -1};
 int_type_to_range({type, _, non_neg_integer, []})      -> {0, pos_inf};
 int_type_to_range({type, _, pos_integer, []})          -> {1, pos_inf};
-int_type_to_range({type, _, range, [{integer, _, I1},
-                                    {integer, _, I2}]})
-                                        when I1 =< I2  -> {I1, I2};
+int_type_to_range({type, _, range, [{Tag1, _, I1}, {Tag2, _, I2}]})
+  when Tag1 =:= integer orelse Tag1 =:= char,
```
gomoripeti: it would be a bit weird from the user to mix integer and char within a range but mixed tags are allowed nevertheless

src/typechecker.erl
```diff
@@ -1933,11 +1939,11 @@ do_type_check_expr_in(Env, Ty, Atom = {atom, LINE, _}) ->
             throw({type_error, Atom, LINE, Ty})
     end;
 do_type_check_expr_in(Env, Ty, Char = {char, _, _}) ->
-    case subtype(type(char), Ty, Env#env.tenv) of
+    case subtype(Char, Ty, Env#env.tenv) of
```
gomoripeti: this is the main fix: although the whole `char()` range might not be a subtype of `Ty`, but the singleton `Char` type can still be.

src/typechecker.erl
```diff
@@ -2295,7 +2301,7 @@ arith_op_arg_types(Op, Ty = {type, _, float, []}) ->
     end;
 
 %% Singleton types are not closed under any operations
-arith_op_arg_types(_, {T, _, _}) when T == integer; T == float ->
```
gomoripeti: (there are no singleton float types)

zuiderkwast:
The last commit looks good.

At some point, I think we need to have our own copy of the `abstract_type()` type. Even if things like this one can be fixed in OTP (we have both created PRs at OTP), we may want to backport them to work on all OTP versions. Apart from this, we can include our own extension with bounded funs and intersection types.

gomoripeti:
@zuiderkwast After seeing so many small bugs in `abstract_type()` type, I also changed my mind and don't think it is overkill to duplicate it. Also I am big proponent of extending it for Gradualizer...so to sum up I totally agree :)

PR #132, opened 2018-12-31
Add gradualizer:type_check_forms/2

### LINKED ISSUE #76: Adding more input consumers
Currently `gradualizer:type_check_file` accepts files ending in `.beam` or `.erl`, however it would be nice to be able to give a BEAM forms term directly (`type_check_forms`?), perhaps other things like accepting an erlang string would be useful as well, however being able to pass in in-memory forms or a binary of a beam file (that may not exist on disk) would be very useful for tools built on this.

zuiderkwast:
This provides a way to type check an AST stored in memory.

Fixes #76

### DIFF
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -19,7 +19,8 @@
          type_check_dir/1,
          type_check_dir/2,
          type_check_files/1,
-         type_check_files/2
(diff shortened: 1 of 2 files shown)


zuiderkwast:
@OvermindDL1

PR #136, opened 2019-01-09
Fix function_clause on improper type_error thrown from add_type_pat/4

erszcz:
Fix errors of the following type:

```erlang
$ gradualizer src/mam/mod_mam_utils.erl
src/mam/mod_mam_utils.erl: escript: exception error: no function clause matching
                 typechecker:handle_type_error({type_error,bin,154,
                                                {type,0,binary,
                                                 [{integer,0,0},
                                                  {integer,0,0}]},
                                                {user_type,152,
                                                 iso8601_datetime_binary,[]}}) (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/typechecker.erl, line 3711)
  in function  typechecker:'-type_check_forms/2-fun-0-'/7 (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/typechecker.erl, line 3598)
  in call from lists:foldr/3 (lists.erl, line 1276)
  in call from gradualizer:type_check_forms/3 (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/gradualizer.erl, line 147)
  in call from gradualizer:'-type_check_files/2-fun-1-'/4 (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/gradualizer.erl, line 118)
  in call from lists:foldl/3 (lists.erl, line 1263)
  in call from gradualizer_cli:handle_args/1 (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/gradualizer_cli.erl, line 17)
```

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3246,14 +3246,14 @@ add_type_pat(String = {string, P, _}, Ty, _TEnv, VEnv) ->
         false ->
             throw({type_error, pattern, P, String, Ty})
     end;
-add_type_pat({bin, P, BinElements} = Bin, Ty, TEnv, VEnv) ->
+add_type_pat({bin, _P, BinElements} = Bin, Ty, TEnv, VEnv) ->
     %% Check the size parameters of the bit pattern
     BinTy = gradualizer_bin:compute_type(Bin),
     Cs1 = case subtype(BinTy, Ty, TEnv) of
               {true, Cs0} ->
                   Cs0;
               false ->
-                  throw({type_error, bin, P, BinTy, Ty})
+                  throw({type_error, Bin, BinTy, Ty})
           end,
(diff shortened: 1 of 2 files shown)


zuiderkwast:
Code change looks good!

> It's strange that codecov says the change has full coverage - it clearly can't have full coverage since the tests we have don't catch this issue.

Yes, it's strange. Codecov doesn't seem to count the coverage on the lines where the `throw` is. See https://codecov.io/gh/josefs/Gradualizer/pull/136/diff . Perhaps it's a bug in the OTP `cover` module.

erszcz:
Here's the test. The case with `cover` is strange, indeed o_O

PR #137, opened 2019-01-10
Unify errors for fun, named_fun, and var

Zalastax:

Still looking for a test case that covers `add_type_pat({var, _, A} = Var, Ty, TEnv, VEnv)`.
I tried exploring different expressions using `=`, generators, and `case` but I didn't manage to find a case where the check fails (and should fail). Help wanted!

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3740,23 +3740,13 @@ handle_type_error({type_error, tyVar, LINE, Var, VarTy, Ty}) ->
     io:format("The variable ~p on line ~p has type ~s "
               "but is expected to have type ~s~n",
               [Var, LINE, typelib:pp_type(VarTy), typelib:pp_type(Ty)]);
-handle_type_error({type_error, {char, Anno, _} = Char, ActualTy, ExpectedTy}) ->
-    io:format("The character ~s on line ~p is expected "
-              "to have type ~s but it has type ~s~n",
-              [erl_pp:expr(Char),
-               erl_anno:line(Anno),
-               typelib:pp_type(ExpectedTy),
-               typelib:pp_type(ActualTy)]);
+handle_type_error({type_error, {char, _, _} = Char, ActualType, ExpectedType}) ->
+    print_type_error("character", Char, ActualType, ExpectedType);
 handle_type_error({type_error, {atom, _, A}, LINE, Ty}) ->
     io:format("The atom ~p on line ~p does not have type ~s~n",
               [A, LINE, typelib:pp_type(Ty)]);
-handle_type_error({type_error, {string, Anno, _} = String, ActualTy, ExpectedTy}) ->
-    io:format("The string ~s on line ~p is expected "
-              "to have type ~s but it has type ~s~n",
-              [erl_pp:expr(String),
-               erl_anno:line(Anno),
-               typelib:pp_type(ExpectedTy),
-               typelib:pp_type(ActualTy)]);
+handle_type_error({type_error, {string, _, _} = String, ActualType, ExpectedType}) ->
+    print_type_error("string", String, ActualType, ExpectedType);
 handle_type_error({type_error, int, I, LINE, Ty}) ->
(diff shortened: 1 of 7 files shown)


UlfNorell:
Here's what QuickCheck came up with to hit that line in the code (don't ask me to explain it though):

```erlang
bang() -> (_ = X) = X = type and error.
```

But it seems we catch the type error in the union case below. I can't find an example where the type error is actually reported.

UlfNorell:
After some more testing I found one:
```erlang
-spec foo() -> integer().
foo() -> 0.

fail() -> X = [X = foo()].
```

zuiderkwast:
How about this one?

```Erlang
-spec f(integer(), atom()) -> ok.
f(X, Y) ->
    case Y of
        X -> ok
    end.
```

Btw, a pattern that can never match doesn't cause a runtime error. It is just dead code. We still consider it a type error, but it should be different IMO. A type error in a pattern should give the same error message as a type error in an expression.

zuiderkwast:
The key point is bound variables in a pattern. For those we can get type errors. For the free ones we can't.

This one could be worth testing as well:

```Erlang
spec f(integer(), atom()) -> ok.
f(X, Y) ->
    case {X, Y} of
        {Z, Z} -> ok
    end.
```

I think the first occurrence Z is free and the second is bound to `integer()`, according to how `add_type_pat/4` is implemented.

gomoripeti:
"It doesn't cause a runtime error. "
Well - your example will cause a `case_clause` runtime error. Also the below example will cause a `badmatch`:
```
-spec f(integer(), atom()) -> any().
f(X, Y) ->
    X = Y.
```
The problem is that the pattern checking code does not know about the context, whether this is just an unreachable clause (warning) or all clauses exhausted (error)

zuiderkwast:
"your example will cause a case_clause runtime error"
I was assuming we don't want to do exhastiveness checking. Non-exhaustive clauses are pretty common and adding catch-all clauses are not considered good style in the Erlang community AFAIK.

Anyway, I think it would be nice to get a message such as **`"The pattern Z on line 3 has type integer() but is expected to have type atom()"`** for type errors in patterns.

src/typechecker.erl
```diff
@@ -3135,15 +3135,15 @@ add_types_pats([Pat | Pats], [Ty | Tys], TEnv, VEnv, PatTysAcc, UBoundsAcc, CsAc
...
+add_type_pat({var, _, A} = Var, Ty, TEnv, VEnv) ->
     %% TODO: In a fun clause, A is always free, but not in e.g. case clauses
     case VEnv of
         #{A := VarTy} ->
             case subtype(VarTy, Ty, TEnv) of
                 {true, Cs} ->
                     {type(none), VarTy, VEnv, Cs};
                 false ->
-                    throw({type_error, tyVar, P, A, VarTy, Ty})
+                    throw({type_error, Var, VarTy, Ty})
```
zuiderkwast: Just to be clear, I don't like the type error for the pattern to be the same as for the expression. (I know this may be out of the scope of this change, but then I think we should add a new task for *Unify type errors for patterns*.)

src/typelib.erl
```diff
@@ -4,6 +4,7 @@
 -export([remove_pos/1, annotate_user_types/2, get_module_from_annotation/1,
          substitute_type_vars/2,
          pp_type/1, debug_type/3, parse_type/1]).
+-export_type([constraint/0, function_type/0, type_et_cetera/0]).
```
zuiderkwast: The name "type_et_cetera" is not very good. I added it just very temporarily and locally. When exporting it, I think it would make sense to find a better name. How about "extended_type"? Is that better?

This type itself isn't very good IMO. Especially, I don't like that intersection is represented simply as a list. We should change the representation of intersection to `{type, _, intersection [our_type()]}` though the code, but that's another task.

Zalastax:
Anything left to do in this particular PR? Anything left that needs to be tracked in an issue?

Issue #129, opened 2018-12-29
Multiple should_fail per file


zuiderkwast:
It would be nice to have multiple functions that should fail in the same test module.

A simple solution for this would be to check that the test returns as many errors as the number of exported functions. (Helper functions are often added and these should not fail.) This can be done after #112.

zuiderkwast:
Exporting the test functions also gets rid of compilation warnings for the test files.

Zalastax:
How do we do this while also testing the error reporting? Passing in `return_errors` will lower the test coverage. Should `should_fail_test` run both with and without `return_errors`?

PR #141, opened 2019-01-14
WIP: Improve type checking records

gomoripeti:
This is a work in progress. I have two questions:

I started to improve handling undefined records and record fields. But then I started to think how Gradualizer is used. In case it will be used as a parse transform (that takes place after pre-processing but before any other compiler steps) then sure these errors can happen. However if Gradualizer is used as other steps like xref, eunit, dialyzer, they are run after successful compilation. If Gradualizer works from a beam file this is surely that case. So why duplicate checks that the compiler already does. Maybe Gradualizer could just run the erlang linter (after epp) in case it works from a source file and the rest of the code can assume there are no linting errors in the AST. What do you think? (There are eg unexported functions in the should_pass examples which result in compiler warnings but Gradualizer happily checks them. I introduced undefined records but compilation fails, that's how I started to think about this)

It is possible to override/refine the types provided in the record definitions in types. eg:
```
-record(rec, {f1 :: atom()}).
-spec f(#rec{f1 :: foo}, term()) -> #rec{f1 :: boolean()}.
f(R, T) ->
  R#rec{f1 = is_integer(T)}.
```
In this case we could make use somehow of the `boolean()` type. Either type-check `is_integer/1` both against `atom()` (from the record definition) and against `boolean()` (from the expected type) ie the intersection of `atom()` and `boolean()`. Or infer the type of the whole record update and check that against the whole `ResTy`.
So in general I think there are multiple flows how to verify the extra info in type-inference and type-checking flows.
Also I'm not sure if this syntax is a pure override or field type (in the above case `boolean()`) must strictly be a subtype of the type from the rec def (ie `atom()`) and whether this also needs to be checked?

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1127,7 +1127,7 @@ expect_fun_type_union(Env, [Ty|Tys]) ->
             [TyOut | expect_fun_type_union(Env, Tys)]
     end.
 
-expect_record_type(Record, {type, _, record, [{atom, _, Name}]}, _TEnv) ->
+expect_record_type(Record, {type, _, record, [{atom, _, Name}|_]}, _TEnv) ->
     if Record == Name ->
          {ok, constraints:empty()};
        true ->
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1409,9 +1409,19 @@ type_check_expr(Env, {record, Anno, Expr, Record, Fields}) ->
     {VB2, Cs2} = type_check_fields(Env, Record, Anno, Fields),
     {RecTy, union_var_binds(VB1, VB2, Env#env.tenv), constraints:combine(Cs1, Cs2)};
 type_check_expr(Env, {record, Anno, Record, Fields}) ->
-    RecTy    = type(record, [{atom, erl_anno:new(0), Record}]),
     {VB, Cs} = type_check_fields(Env, Record, Anno, Fields),
-    {RecTy, VB, Cs};
+    {FieldTypes, VarBindsList, Css}
+                = lists:unzip3(
+                    lists:map(
+                      fun({record_field, _, FieldNameAtom, Expr}) ->
+                              {FTy, FVB, FCs} = type_check_expr(Env, Expr),
+                              {type(field_type, [FieldNameAtom, FTy]), FVB, FCs}
+                      end
+                     ,Fields)),
+    RecTy = type(record, [{atom, erl_anno:new(0), Record} | FieldTypes]),
+    {RecTy
+    ,union_var_binds([VB|VarBindsList], Env#env.tenv)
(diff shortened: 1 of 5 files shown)


Zalastax:
It makes sense to me that we should not need to do the same checks as the compiler already does. With that said, warnings are not errors and Gradualizer should not reject typechecking a module just because there are unexported functions (it should even check unexported functions in my opinion).

src/typechecker.erl
```diff
@@ -1412,8 +1412,19 @@ type_check_expr(Env, {record, Anno, Record, Fields}) ->
     RecTy    = type(record, [{atom, erl_anno:new(0), Record}]),
     {VB, Cs} = type_check_fields(Env, Record, Anno, Fields),
     {RecTy, VB, Cs};
-type_check_expr(_Env, {record_index, _, _Record, _Field}) ->
-    {type(integer), #{}, constraints:empty()};
+type_check_expr(Env, {record_index, Anno, Name, FieldNameAtom}) ->
+    %% we still fetch record field index even if infer=false
+    %% to be able to report undefined record or record field.
+    %% ("we do our best for the customer")
```
zuiderkwast: In other words, we're checking things that we don't need, to be able to report an error that the compiler would also report.

In principle, I think it's the wrong approach, but in this case, I don't see that there might be any performance impact and the complexity of the code is the same either way, so keep it!

Reporting an exact type for record index when inference is on is very good!

test/should_fail/record_index.erl
```diff
@@ -1,6 +1,10 @@
 -module(record_index).
 
--export([f/0, g/0]).
+-export([f/0, g/0,
+         undefined_record/0,
+         undefined_record_infer/0,
```
zuiderkwast: These are not automatically tested as only one fail per module is enough for `should_fail`. We should really fix #129.

src/typechecker.erl
```diff
@@ -1127,7 +1127,7 @@ expect_fun_type_union(Env, [Ty|Tys]) ->
             [TyOut | expect_fun_type_union(Env, Tys)]
     end.
 
-expect_record_type(Record, {type, _, record, [{atom, _, Name}]}, _TEnv) ->
+expect_record_type(Record, {type, _, record, [{atom, _, Name}|_]}, _TEnv) ->
```
zuiderkwast: In general, please add a TODO comment when adding code to ignore unimplemented stuff.

gomoripeti:
thank you guys for the feedback. I split this into multiple PRs
- #146 to fix an error
- #147 cosmetic change in undefined rec/field. this should be the responsibility of the compiler.
- I will work more on more precise type inference and record field type refinement

gomoripeti:
closing this as usable parts are merged in separate PRs. There is work left in refined record types

PR #143, opened 2019-01-14
Multiple expected errors

Zalastax:
I'm not too happy about how `test.erl` now uses `parsedata` and `typechecker:collect_specs_types_opaques_and_functions` but I prefer that to duplicating the logic for traversing the forms.

All deleted files should have moved to a file with a very similar name so I hope the diff is not too annoying.


test/test.erl
```diff
@@ -44,3 +62,11 @@ known_problem_should_fail_test_() ->
 map_erl_files(Fun, Dir) ->
     Files = filelib:wildcard(filename:join(Dir, "*.erl")),
     lists:map(Fun, Files).
+
+number_of_exported_functions(File) ->
+    {ok, Forms} = gradualizer_file_utils:get_forms_from_erl(File),
+    ParseData = typechecker:collect_specs_types_opaques_and_functions(Forms),
+    case ParseData#parsedata.export_all of
```
gomoripeti: what if this `case` would be moved to `typechecker:number_of_exported_functions(ParseData)` (or into `typechecker:number_of_exported_functions(Forms)` including `typechecker:collect_specs_types_opaques_and_functions(Forms)`)? Then don't need to copy `#parsedata{}`.

Zalastax: Ah, that's a smart work-around!

PR #146, opened 2019-01-15
Fix subtype checking record_field and record_index

gomoripeti:


test/should_fail/record_field.erl
```diff
@@ -1,9 +1,13 @@
...
 
 -record(rec, { apa :: integer()}).
 
 -spec f(#rec{}) -> boolean().
 f(A) ->
     A#rec.apa.
+
+-spec g(#rec{}) -> pos_integer().
+g(A) ->
+    A#rec.apa.
```
zuiderkwast: Why should this fail? `apa` is the first record element, i.e. the 2nd tuple element, isn't it?

test/should_fail/record_index.erl
```diff
@@ -1,9 +1,13 @@
...
 
 -record(rec, { apa :: integer()}).
 
 -spec f() -> boolean().
 f() ->
     #rec.apa.
+
+-spec g() -> 2.
+g() ->
+    #rec.apa.
```
zuiderkwast: Why should this fail? `apa` is the first record element, i.e. the 2nd tuple element, isn't it?

gomoripeti: the inferred type if record index is currently `integer()`, so this check should fail because it is not a subtype of `2`. But you are right, I wanted to wright a testcase which will still fail when exact record index inference will be added... but I failed :)

Zalastax: Move this one too known_problems/should pass and add a similar one with the wrong number!

gomoripeti: right, instead of creating a known problem, I pushed another commit with "very precise record index type", not to increase tech debt.

Issue #148, opened 2019-01-16
typechecker:handle_type_error function_clause again


erszcz:
I've just found this:

```sh
$ gradualizer src/history_archiver.erl
...
src/history_archiver.erl: escript: exception error: no function clause matching
                 typechecker:handle_type_error({type_error,
                                                {map,201,
                                                 {var,201,'Ctx'},
                                                 [{map_field_exact,201,
                                                   {atom,201,archive_id},
                                                   {var,201,'ID'}}]},
                                                {type,153,map,
                                                 [{type,153,map_field_exact,
                                                   [{atom,153,'__struct__'},
                                                    {atom,153,mam_context}]},
                                                  {type,154,map_field_exact,
                                                   [{atom,154,from},
                                                    {type,154,any,[]}]},
                                                  {type,155,map_field_exact,
                                                   [{atom,155,to},
                                                    {type,155,any,[]}]},
                                                  {type,156,map_field_exact,
                                                   [{atom,156,message},
                                                    {remote_type,156,
                                                     [{atom,156,exml},
                                                      {atom,156,element},
                                                      []]}]},
                                                  {type,157,map_field_exact,
                                                   [{atom,157,archive_id},
                                                    {type,157,union,
                                                     [{atom,157,undefined},
                                                      {type,157,binary,
                                                       []}]}]}]},
                                                {user_type,0,
                                                 mam_context_with_id,[]}}) (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/typechecker.erl, line 3734)
  in function  typechecker:'-type_check_forms/2-fun-0-'/7 (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/typechecker.erl, line 3613)
  in call from lists:foldr/3 (lists.erl, line 1276)
  in call from gradualizer:type_check_forms/3 (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/gradualizer.erl, line 147)
  in call from gradualizer:'-type_check_files/2-fun-1-'/4 (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/gradualizer.erl, line 118)
  in call from lists:foldl/3 (lists.erl, line 1263)
  in call from gradualizer_cli:handle_args/1 (/Users/erszcz/work/josefs/gradualizer/_build/default/lib/gradualizer/src/gradualizer_cli.erl, line 17)
```

[...]

Issue #140, opened 2019-01-11
Non-exhaustive matches in case expressions


erszcz:
I'm preparing a short talk on Gradualizer for our local community. One of the basic arguments for static typing I wanted to show is that `case` expressions are checked for all alternatives of an ADT. For example:

```haskell
data Allergen = Eggs
              | Chocolate
              | Pollen
              | Cats

allergenScore :: Allergen -> Int
allergenScore al = case al of
  Eggs         ->   1
  Chocolate    ->  32
  Pollen       ->  64
```

GHC will warn us that `Cats` are not taken into account:

```
$ ghc -Wincomplete-patterns src/Allergies.hs
[1 of 1] Compiling Allergies        ( src/Allergies.hs, src/Allergies.o )

src/Allergies.hs:13:20: warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In a case alternative: Patterns not matched: Cats
   |
13 | allergenScore al = case al of
   |                    ^^^^^^^^^^...
```

As will the PureScript compiler (which is somewhat more interesting due to its [Erlang backend](https://github.com/purerl/purescript)):

```
$ purs compile src/Allergies.purs
Compiling Allergies
Error found:
in module Allergies
at src/Allergies.purs line 13, column 20 - line 16, column 20

  A case expression could not be determined to cover all inputs.
  The following additional cases are required to cover all inputs:

    Cats

  Alternatively, add a Partial constraint to the type of the enclosing value.
```

Gradualizer as of now doesn't seem to provide such a warning for a literal translation of the example:

```erlang
-type allergen() :: eggs
                  | chocolate
                  | pollen
                  | cats.

-spec allergen_score(allergen()) -> integer().
allergen_score(Al) ->
    case Al of
        eggs         ->  1;
        chocolate    -> 32;
        pollen       -> 64
    end.
```

Should we expect that it will in the future?

zuiderkwast:
1. The use of non-exhaustive cases is considered good style in the Erlang community according to the *let it crash* principle. Catch-all clauses are considered bad style. Perhaps we can implement it anyway, but I would suggest an option for this, perhaps per function, e.g. `-gradualizer([{exhastiveness_checking, [allergen_score/1]}])`.

[...]

PR #151, opened 2019-01-18
Map update type_error handling

erszcz:
This partially addresses #148 - partially, since a new clause for `af_map_update/1` variant of `erl_parse:abstract_expr/0` is added to `handle_type_error` to fix the crash, but there's also a new known problem uncovered:

```erlang
-type map_with_integer() :: #{a := integer()}.
-type map_with_binary()  :: #{a := binary()}.

-spec f(map_with_integer()) -> map_with_binary().
f(#{} = Ctx) ->
    Ctx#{a := <<"binary">>}.
```

The above does not typecheck.

Actually, this crash is a nice example of why exhaustiveness checking for all type constructors of a given type would be a cool feature (#140).


test/should_fail/map_field_invalid_update.erl
```diff
@@ -0,0 +1,10 @@
+-module(map_field_invalid_update).
+
+-export([f/1]).
+
+-type map_with_integer() :: #{a := integer()}.
+-type map_with_binary()  :: #{a := binary()}.
+
+-spec f(map_with_integer()) -> map_with_binary().
```
zuiderkwast: You could write `#{a := integer()}` and `#{a := binary()}` directly in the spec, to keep user-defined types out of the test case. Just to make sure the user-def types handling is not affecting anything.

test/known_problems/should_pass/map_field_valid_update.erl
```diff
@@ -0,0 +1,10 @@
+-module(map_field_valid_update).
+
+-export([f/1]).
+
+-type map_with_integer() :: #{a := integer()}.
+-type map_with_binary()  :: #{a := binary()}.
+
+-spec f(map_with_integer()) -> map_with_binary().
```
zuiderkwast: See next comment

src/typechecker.erl
```diff
@@ -3894,6 +3894,8 @@ handle_type_error({type_error, list_op_error, ListOp, P, Ty, _}) ->
               [ListOp, P, typelib:pp_type(Ty)]);
 handle_type_error({type_error, {map, _, _} = Map, ActualType, ExpectedType}) ->
     print_type_error("map", Map, ActualType, ExpectedType);
+handle_type_error({type_error, {map, _, _, _} = Map, ActualType, ExpectedType}) ->
+    print_type_error("map", Map, ActualType, ExpectedType);
```
zuiderkwast: I suggest "map update" here, to distinguish between the messages for `M#{a=>b}` and `#{a=>b}`.

zuiderkwast:
(@Zalastax I think a squash commit would have been appropriate in this case. Not that it matters..)

PR #152, opened 2019-01-18
Unify error printing and unify errors for float, int

Zalastax:
Part of the series of PR's replacing #59, see #49

After discussion with @zuiderkwast we realized that all unified type errors are unique in that they are four-tuples where the second element is_tuple. The goal is to get all expressions that we can to match this format. The place I was worried about is how to replace binary operators, e.g. int_error. Our conclusion is that printing the erroneous expression and its expected type should be enough. The biggest missing feature is info about the column of an expression. I guess it's in the OTP backlog but is there a way we could help speed that process up (contributing or stating its importance).

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1944,12 +1944,12 @@ do_type_check_expr_in(Env, Ty, {float, LINE, F}) ->
         false ->
             throw({type_error, float, F, LINE, Ty})
     end;
-do_type_check_expr_in(Env, Ty, Atom = {atom, LINE, _}) ->
+do_type_check_expr_in(Env, Ty, Atom = {atom, _, _}) ->
     case subtype(Atom, Ty, Env#env.tenv) of
         {true, Cs} ->
             {#{}, Cs};
         false ->
-            throw({type_error, Atom, LINE, Ty})
+            throw({type_error, Atom, Atom, Ty})
     end;
 do_type_check_expr_in(Env, Ty, Char = {char, _, _}) ->
     case subtype(Char, Ty, Env#env.tenv) of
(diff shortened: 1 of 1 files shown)


gomoripeti:
regarding columns: the tokenizer and parser already supports including columns if start location is `{1, 1}` instead of `1` given to `erl_scan`. There seems to be also `epp:open/5` (https://github.com/erlang/otp/blob/master/lib/stdlib/src/epp.erl#L112 - only used by `escript` to skip the non-erlang header lines) which has a `StartLocation` argument, so it seems to be an easy patch to add a `location` option to `epp:parse_file/2` and we would get column info in the AST annotations.

src/typechecker.erl
```diff
@@ -3980,14 +3952,31 @@ handle_type_error({type_error, mismatch, Ty, Expr}) ->
...
     io:format("TYPE ERROR~n").
 
--spec print_type_error(string(),
-                       erl_parse:abstract_expr(),
+-spec describe_expr(erl_parse:abstract_expr()) -> string().
+describe_expr({atom, _, _})               -> "atom";
+describe_expr({bin, _, _})                -> "bit expression";
+describe_expr({char, _, _})               -> "character";
+describe_expr({cons, _, _, _})            -> "list";
+describe_expr({'fun', _, _})              -> "function expression";
```
gomoripeti: maybe this can be called "anonymous function" and `named_fun` "named function"?
(but you are right: "fun expression" is mentioned in the abstract format documentation)

josefs:
@Zalastax I just want to say I admire the way you've taken on this massive task of perfecting the error message format. Good work, and keep those PR's coming!

src/typechecker.erl
Zalastax: Fixed in follow up commit, forgot to fix before merge.

PR #153, opened 2019-01-21
Print column numbers in errors if available

gomoripeti:
This will only work if the abstract forms are aquired from an erlang
source file. Debug info from a beam file won't contain column numbers.

Many places in `handle_type_error` assume the annotation is a single
integer representing the line number, whereas it can be a location
two-tuple now. But the printouts will be fixed gradually as we unify
errors.

(Inspired by Pierre's comment https://github.com/josefs/Gradualizer/pull/152#issue-245932385)

### DIFF
--- a/src/gradualizer_file_utils.erl
+++ b/src/gradualizer_file_utils.erl
@@ -21,13 +21,38 @@
 
 -spec get_forms_from_erl(file:filename()) -> parsed_file() | parsed_file_error().
 get_forms_from_erl(File) ->
-    case epp:parse_file(File, []) of
+    case epp_parse_file(File) of
         {ok, Forms} ->
             {ok, Forms};
         {error, enoent} ->
             {file_not_found, File};
         {error, Reason} ->
             {file_open_error, {Reason, File}}
+    end.
+
+%% @doc Preprocess and parse a file including column numbers in the result
+epp_parse_file(File) ->
+    case file:open(File, [read]) of
+        {ok, Fd} ->
+            try
+                StartLocation = {1, 1},
+                case epp:open(File, Fd, StartLocation, [], []) of
+                    {ok, Epp} ->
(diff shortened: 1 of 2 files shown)


src/gradualizer_file_utils.erl
```diff
@@ -30,6 +30,26 @@ get_forms_from_erl(File) ->
             {file_open_error, {Reason, File}}
     end.
 
+%% @doc Preprocess and parse a file including column numbers in the result
+epp_parse_file(File) ->
+    case file:open(File, [read]) of
```
zuiderkwast: The file is not closed afterwards.

Zalastax:
Absolutely awesome!
The text is getting a bit long to me, May any of these formats be more suitable:
The ~s ~s (<L>, <C>)
The ~s ~s (lin: <L>, col: <C>)
The ~s ~s at L<L> C<C>
The ~s ~s at <L>:<C>

Even better would be to strive for https://elm-lang.org/blog/compiler-errors-for-humans, but that's for the future 😺

src/gradualizer_file_utils.erl
gomoripeti: good catch

gomoripeti:
I agree with a bit long message. (btw inspiring blog post)
Until now I was on the side that location should just be appended to file name (`src/typechecker.erl:4023:58: ...`). This is definitely friendly for editors to parse (and jump to the error). But as I read this is not the easiest for humans. Maybe there could be different formatters for editors and one-liners and multiline, verbose and extra-user-friendly messages for humans.

(sidenote: to reconstruct the original text apart from start line+column we would also need end line+column (plus the plain text source) and that is unfortunately not supported by the erlang scanner/parser)

gomoripeti:
(fixed file close and rebased to resolve conflict)

Issue #122, opened 2018-12-23
Bounded quantification


zuiderkwast:
Picking up this topic from #93. I think we do already agree about these points:

* Type variables are often used where the a type could just have been used directly.
* For type variables appearing only once, bounded quantification doesn't make sense.
* For type variables appearing only in the result, bounded quantification doesn't make sense.

But I think there are other cases where bounded quantification does make sense. I don't agree about this point:

>>> [bounded quantification] is way too strict for variables occurring more than once

>>  How so?

> Would you really expect the function -spec foo([Y]) -> Y when Y :: number() to only type check when returning one of the numbers already in the list? No adding them up or multiplying them?

Yes, I do think it makes sense that `foo/1` above must return one of the elements in the list. For many of the lists functions, this is what you want. Here are some examples (with only-once used variables unfolded):

```Erlang
-spec lists:any(fun((T) -> boolean()), [T]) -> boolean().
-spec lists:filter(fun((T) -> boolean()), [T]) -> [T].
-spec lists:map(fun((A) -> B), [A]) -> [B].
-spec lists:sort([T]) -> [T].
```

These often have `when T :: term()` in the spec, which I think shouldn't make a difference.

If we just unfold, we get `[term()]` or `term()` as the return type of many of these functions, which gives rise to a type error where e.g. `integer()` is expected. Here are examples of applying these functions:

```Erlang
-spec integer_sort(integer()) -> integer().
integer_sort(Xs) -> lists:sort(Xs).

-spec filter_even([integer()]) -> [integer()].
filter_even(Xs) -> lists:filter(fun (X) -> X rem 2 == 0 end, Xs).
```

I have a feeling that many of the remaining errors reported when gradualizing the gradualizer are from these, the errors involving `term()`. (Some specs are just wrong though, e.g. that of `lists:foldl/3` which always returns `term()`. These we can override.)

So, I suggest we go for bounded quantification where it makes sense, in a pragmatic way. What do you think?

zuiderkwast:
True, dropping `T :: term()` is another approach. Then, I can't see any real world examples of bounded polymorphic variables. I can only come up with more unlikely ones like this one:

```Erlang
%% Sorts the points by their distance to origo
-spec numeric_pair_sort([T]) -> [T] when T :: {number(), number()}.
```

Is truly polymorphic variables is easier to handle?

How do you implement polymorphism? Do we simply store type variable bindings within the Env which is passed around?

UlfNorell:
[...]

PR #154, opened 2019-01-26
New specs for functions in the lists module

josefs:
I've started the work on creating alternative type specs for otp modules which have specs that don't work well with the Gradualizer. This PR handles the lists module.

Most of the changes are just removing `T :: term()` from the constraints. 

I'd appreciate feedback on:

- the type of `reverse/2`. I've given it a slightly more restrictive type than in otp. For reference, the otp type is
    ```
    -spec reverse(List1, Tail) -> List2 when
      List1 :: [T],
      Tail :: term(),
      List2 :: [T],
      T :: term().
    ```
- Naming convention of type variables. I've mostly followed the naming convention from otp, but it mixes `A`, `B`, `C` with `X`, `Y`, `Z` and simply `T`. Should we try to harmonize the names?

- I have added union types to the return type of `merge` and `umerge`. Is that appropriate?

EDIT: No more unions in the type of `filter`

### DIFF
--- a/src/gradualizer_prelude.erl
+++ b/src/gradualizer_prelude.erl
@@ -31,7 +31,7 @@
 
 -spec lists:filter(fun((T) -> boolean() | {true, Value}), [T]) -> [T | Value].
 
--spec lists:flatmap(fun((A) -> B), [A]) -> [B].
+-spec lists:flatmap(fun((A) -> [B]), [A]) -> [B].
 
 -type deep_list(A) :: [A | deep_list(A)].
 
--- a/src/gradualizer_prelude.erl
+++ b/src/gradualizer_prelude.erl
@@ -30,6 +30,8 @@
 -spec lists:duplicate(T, non_neg_integer()) -> [T].
 
 -spec lists:filter(fun((T) -> boolean()), [T]) -> [T].
+-spec lists:filtermap(fun((Elem) -> boolean() | {'true', Value}), [Elem])
+		     -> [Elem | Value].
 
 -spec lists:flatmap(fun((A) -> [B]), [A]) -> [B].
 
--- a/src/gradualizer_prelude.erl
+++ b/src/gradualizer_prelude.erl
@@ -38,7 +38,7 @@
 -spec lists:flatten(deep_list(A))      -> [A].
 -spec lists:flatten(deep_list(A), [A]) -> [A].
 
--spec lists:foldl(fun((T, Acc) -> Acc), [T], Acc) -> Acc.
+-spec lists:foldl(fun((T, Acc) -> Acc), Acc, [T]) -> Acc.
 -spec lists:foldr(fun((T, Acc) -> Acc), [T], Acc) -> Acc.
 -spec lists:join(T, [T]) -> [T].
 
--- a/src/gradualizer_prelude.erl
+++ b/src/gradualizer_prelude.erl
@@ -39,7 +39,7 @@
 -spec lists:flatten(deep_list(A), [A]) -> [A].
 
 -spec lists:foldl(fun((T, Acc) -> Acc), Acc, [T]) -> Acc.
--spec lists:foldr(fun((T, Acc) -> Acc), [T], Acc) -> Acc.
+-spec lists:foldr(fun((T, Acc) -> Acc), Acc, [T]) -> Acc.
 -spec lists:join(T, [T]) -> [T].
 
 -spec lists:foreach(fun((T) -> term()), [T]) -> ok.
--- a/src/gradualizer_prelude.erl
+++ b/src/gradualizer_prelude.erl
@@ -5,6 +5,7 @@
 -spec erlang:hd([A, ...]) -> A.
 -spec erlang:tl([A, ...]) -> [A].
 
+
 %% change return Val to any() from term()
 -spec application:get_env(Par) -> 'undefined' | {'ok', Val} when
       Par :: atom(),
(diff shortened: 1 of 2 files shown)


src/gradualizer_prelude.erl
```diff
@@ -18,3 +19,81 @@
...
+-spec lists:append([[T]])     -> [T].
+-spec lists:append([T], [T])  -> [T].
+-spec lists:delete(T, [T])    -> [T].
+-spec lists:droplast([T, ...]) -> [T].
+
+-spec lists:dropwhile(fun((T) -> boolean()), [T]) -> [T].
+
+-spec lists:duplicate(T, non_neg_integer()) -> [T].
+
+-spec lists:filter(fun((T) -> boolean() | {true, Value}), [T]) -> [T | Value].
```
UlfNorell: Is this right? I get
```
1> lists:filter(fun(X) -> {true, bla} end, [1, 2, 3]).   
** exception error: no case clause matching {true,bla}
     in function  lists:'-filter/2-lc$^0/1-0-'/2 (lists.erl, line 1286)
```

src/gradualizer_prelude.erl
```diff
@@ -18,3 +19,81 @@
...
+
+-spec lists:flatmap(fun((A) -> B), [A]) -> [B].
```
UlfNorell: Should be
```suggestion
-spec lists:flatmap(fun((A) -> [B]), [A]) -> [B].
```

src/gradualizer_prelude.erl
```diff
@@ -18,3 +19,81 @@
...
+
+-type deep_list(A) :: [A | deep_list(A)].
+
+-spec lists:flatten(deep_list(A))      -> [A].
+-spec lists:flatten(deep_list(A), [A]) -> [A].
+
+-spec lists:foldl(fun((T, Acc) -> Acc), [T], Acc) -> Acc.
```
UlfNorell:
```suggestion
-spec lists:foldl(fun((T, Acc) -> Acc), Acc, [T]) -> Acc.
```

src/gradualizer_prelude.erl
```diff
@@ -18,3 +19,81 @@
...
+-spec lists:foldr(fun((T, Acc) -> Acc), [T], Acc) -> Acc.
```
UlfNorell:
```suggestion
-spec lists:foldr(fun((T, Acc) -> Acc), Acc, [T]) -> Acc.
```

src/gradualizer_prelude.erl
```diff
@@ -18,3 +19,81 @@
...
+
+-spec lists:map(fun((A) -> B), [A]) -> [B].
+-spec lists:mapfoldl(fun((A, Acc) -> {B, Acc}), Acc, [A]) -> {[B], Acc}.
+-spec lists:mapfoldr(fun((A, Acc) -> {B, Acc}), Acc, [A]) -> {[B], Acc}.
+
+-spec lists:max([T, ...]) -> T.
+-spec lists:min([T, ...]) -> T.
+
+-spec lists:merge([[T]]) -> [T].
+-spec lists:merge([X], [Y]) -> [X | Y].
```
UlfNorell: Why not
```suggestion
-spec lists:merge([T], [T]) -> [T].
```
like `append`?

src/gradualizer_prelude.erl
```diff
@@ -18,3 +19,81 @@
...
+-spec lists:merge(fun((A, B) -> boolean()), [A], [B]) -> [A | B].
```
UlfNorell: Ah here it makes sense to distinguish the types!

src/gradualizer_prelude.erl
```diff
@@ -18,3 +19,81 @@
...
+-spec lists:partition(fun((T) -> boolean()), [T]) -> {[T], [T]}.
+
+-spec lists:reverse([T]) -> [T].
+
+-spec lists:reverse([T], [T]) -> [T].
+
+-spec lists:sort([T]) -> [T].
+-spec lists:sort(fun((T, T) -> boolean()), [T]) -> [T].
+
+-spec lists:split(integer(), [T]) -> {[T], [T]}.
```
UlfNorell: `non_neg_integer()`?

src/gradualizer_prelude.erl
josefs: Indeed. Thanks!

src/gradualizer_prelude.erl
josefs: I'm in two minds about this one. I agree that your suggested type is simpler and in some ways better. But the type I gave is similar to that of `merge/3` and there is something to be said about them having similar types for symmetry.

zuiderkwast:
> Most of the changes are just removing `T :: term()` from the constraints.

Why is this necessary?

josefs:
@zuiderkwast the reasons for removing `term()` from the types are discussed in #93 and #122. I can expand on that if needed.

josefs:
This is the PR that keeps on giving.

Travis CI fails because the dialyzer crashes. I've reported that to the Erlang bug tracker.
In order to make the build work again I tried to add the following line to `rebar.config`:
`{dialyzer, [exclude_mods, ["gradualizer_prelude"]]}.
`
That makes rebar crash. I'll make sure to report that too.

If anyone has a good idea for how to make this PR go through all our checks, I'm all ears.

zuiderkwast:
> Travis CI fails because the dialyzer crashes. [...] That makes rebar crash.

Gahhh! Good ideas though! Perhaps the module in `exclude_mods` should be an atom instead of a string?

```Erlang
{dialyzer, [exclude_mods, [gradualizer_prelude]]}.
```

zuiderkwast:
> @zuiderkwast the reasons for removing term() from the types are discussed in #93 and #122. I can expand on that if needed.

Yes, and I added a0241c6a1b2ba5d1a54b0c14bb52824ca88a353b to remove constraints on the form `T :: term()` when checking function definitions. The specs in this "prelude" is only used for checking function calls though. What's the harm in checking that an argument of a function is a subtype of `term()`? You can elaborate if you want, but I can also just accept that I don't completely follow. :-)

josefs:
> just an ignorant question: what is the difference between lists:append([A], [B]) -> [A | B] and lists:append([T], [T]) -> [T]? (And if nothing, is the later chosen because simpler is better?)

Now that I think of it, I don't think there is a practical difference between these types. Let's see what happens in the type checker if we have two lists `Xs :: [integer()]`and `Ys :: [boolean()]` and use the two different versions of append:

- For the first version, the type variables `A` and `B` are required to be supertypes of `integer()` and `boolean()` respectively. Assuming that these are the only constraints on `A` and `B` we get the final type `[integer() | boolean()]`.
- For the second version, the type variable `T` is required to be a supertype of both `integer()` and `boolean()`. The sensible type to give `T` is `integer() | boolean()`.

Things become slightly messier if we only know the type of the result of `append`. In the first type we would be required to somehow split up the known result type into a union. But there is really no sensible way of doing that. I don't even know what the Gradualizer does currently in such a case. The latter type for `append` is clearly simpler, as the known result type must match `[T]` and that forces the arguments to be of the same type.

Hmmm. You've almost convinced me to go with the simpler type. 

This is all assuming that constraint solving is implemented, which is currently (still) isn't.

josefs:
> What's the harm in checking that an argument of a function is a subtype of `term()`? You can elaborate if you want, but I can also just accept that I don't completely follow. :-)

You're absolutely right in that there is no harm if an argument to a function is `term()`. It's a lot worse when the result is of that type though. It makes the type rather useless as it means that the function can return literally anything.

So removing `term()` from arguments is not strictly necessary, but makes the types simpler and more symmetric once we've removed `term()` from result types.

josefs:
I've resolved the dialyzer/rebar crashes. The reason rebar crashed was that I had forgotten to use a tuple. Here's what caused rebar to crash:

`{dialyzer, [exclude_mods, ["gradualizer_prelude"]]}.
`

and here is what finally worked:

`{dialyzer, [{exclude_mods, [gradualizer_prelude]}]}.
`

PR #155, opened 2019-01-28
Unify and test first part of arith_op and int_op

Zalastax:
I'm leaving `type_check_arith_op_in` as is for now since this PR is big enough already.
I'd love if the [Wiki entry on bidirectional-type-checking](https://github.com/josefs/Gradualizer/wiki/Bidirectional-type-checking) was improved - I can get by, but it's a bit difficult since there are few comments in the codebase.

I've added some known problems that I think we ought to catch. We are currently quite inconsistent, e.g. changing the functions in `known_problems/should_pass/int_op.erl` to below makes them both fail.
 ```Erlang
-spec plusfloatany(float(), any()) -> integer().
plusfloatany(X, Y) ->
    X + Y.

-spec divfloatany(float(), any()) -> integer().
divfloatany(X, Y) -> X div Y.
```

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1689,33 +1689,48 @@ type_check_rel_op(Env, Op, P, Arg1, Arg2) ->
             end
     end.
 
-type_check_arith_op(Env, Op, P, Arg1, Arg2) ->
+type_check_arith_op(Env, _, _, Arg1, Arg2) ->
     {Ty1, VB1, Cs1} = type_check_expr(Env, Arg1),
     {Ty2, VB2, Cs2} = type_check_expr(Env, Arg2),
 
-    case compat_arith_type(Ty1,Ty2) of
-        false ->
-          throw({type_error, arith_error, Op, P, Ty1, Ty2});
-        Ty ->
-            {Ty
-            ,union_var_binds(VB1, VB2, Env#env.tenv)
-            ,constraints:combine(Cs1, Cs2)}
-    end.
+    TNumber = type(number),
+    TFloat = type(float),
+    expect_expr_subtype(Arg1, Ty1, TNumber, #tenv{}),
+    expect_expr_subtype(Arg2, Ty2, TNumber, #tenv{}),
 
-type_check_int_op(Env, Op, P, Arg1, Arg2) ->
+    Ty = case {Ty1, Ty2, Env#env.infer} of
+             {{type, _, any, _}, _, false} -> type(any);
+             {_, {type, _, any, _}, false} -> type(any);
+             {_, {type, _, any, _}, true} -> TFloat;
+             {{type, _, any, _}, _, true} -> TFloat;
+             _ ->
+                 case {subtype(Ty1, TFloat, #tenv{}),
+                       subtype(Ty2, TFloat, #tenv{})} of
+                     {{true, _}, _} -> TFloat;
+                     {_, {true, _}} -> TFloat;
+                     _              -> type(integer)
+                 end
+         end,
+    {Ty,
+     union_var_binds(VB1, VB2, Env#env.tenv),
+     constraints:combine(Cs1, Cs2)}.
+
+type_check_int_op(Env, _, _, Arg1, Arg2) ->
     {Ty1, VB1, Cs1} = type_check_expr(Env, Arg1),
     {Ty2, VB2, Cs2} = type_check_expr(Env, Arg2),
 
-    case compat_arith_type(Ty1,Ty2) of
-        false ->
-            throw({type_error, int_error, Op, P, Ty1, Ty2});
-        {type, _, Ty, []} when Ty == float orelse Ty == number ->
-            throw({type_error, int_error, Op, P, Ty1, Ty2});
-        Ty ->
-            {Ty
-            ,union_var_binds(VB1, VB2, Env#env.tenv)
-            ,constraints:combine(Cs1, Cs2)}
-    end.
+    TInteger = type(integer),
+    expect_expr_subtype(Arg1, Ty1, TInteger, #tenv{}),
+    expect_expr_subtype(Arg2, Ty2, TInteger, #tenv{}),
+
+    Ty = case {Ty1, Ty2, Env#env.infer} of
+        {{type, _, any, []}, _, false} -> type(any);
+        {_, {type, _, any, []}, false} -> type(any);
+        _                              -> TInteger
+    end,
+    {Ty,
+     union_var_binds(VB1, VB2, Env#env.tenv),
+     constraints:combine(Cs1, Cs2)}.
 
 type_check_list_op(Env, Op, P, Arg1, Arg2) ->
   {Ty1, VB1, Cs1} = type_check_expr(Env, Arg1),
@@ -1788,6 +1803,14 @@ type_check_call_ty_intersect(Env, [Ty | Tys], Args, E) ->
             type_check_call_ty_intersect(Env, Tys, Args, E)
     end.
 
+expect_expr_subtype(Expr, ExprTy, ExpectedTy, TEnv) ->
+    case subtype(ExprTy, ExpectedTy, TEnv) of
+        {true, Constraints} ->
+           Constraints;
+       false ->
+           throw({type_error, Expr, ExprTy, ExpectedTy})
+    end.
+
 type_check_call_ty_union(Env, Tys, Args, E) ->
     {ResTys, VBs, Css} =
         lists:unzip3([type_check_call_ty(Env, Ty, Args, E)
@@ -1795,46 +1818,6 @@ type_check_call_ty_union(Env, Tys, Args, E) ->
     {normalize(type(union, ResTys), Env#env.tenv),
      union_var_binds(VBs, Env#env.tenv),
      constraints:combine(Css)}.
-
-compat_arith_type(Any = {type, _, any, []}, {type, _, any, []}) ->
-    Any;
-compat_arith_type(Any = {type, _, any, []}, Ty) ->
-    case subtype(Ty, type(number), #tenv{}) of
-        false ->
-            false;
-        _ ->
-            Any
-    end;
-compat_arith_type(Ty, Any = {type, _, any, []}) ->
-    case subtype(Ty, type(number), #tenv{}) of
-        false ->
-            false;
-        _ ->
-            Any
-    end;
-compat_arith_type(Ty1, Ty2) ->
-    TInteger = type(integer),
-    case {subtype(Ty1, TInteger, #tenv{})
-         ,subtype(Ty2, TInteger, #tenv{})} of
-        {{true,_},{true,_}} ->
-            TInteger;
-        _ ->
-            TFloat = type(float),
-            case {subtype(Ty1, TFloat, #tenv{})
-                 ,subtype(Ty2, TFloat, #tenv{})} of
-                {{true,_},{true,_}} ->
-                    TFloat;
-                _ ->
-                    TNumber = type(number),
-                    case {subtype(Ty1, TNumber, #tenv{})
-                         ,subtype(Ty2, TNumber, #tenv{})} of
-                        {{true,_},{true,_}} ->
-                            TNumber;
-                        _ ->
-                            false
-                    end
-            end
-    end.
 
 type_check_comprehension(Env, lc, Expr, []) ->
     {Ty, _VB, Cs} = type_check_expr(Env, Expr),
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by zuiderkwast** [comment]
You're saying that there is inconsistency depending on in which direction the type-checking operates. The code in the description works as expected, but the code you added in the known problem is what this issue is about: the type inference/propagation mode.

```Erlang
%% From known_problems/int_op.erl
-spec plusfloatany(float(), any()) -> integer().
plusfloatany(X, Y) ->
    A = X + Y,
    A.
```

Here, are you saying that `+` should propagate the type of it's non-any argument, so that for `float() + any()` we return `float()`? And currently, we return `any()` I assume.

This is similar to what we do for e.g. tuples. For `{1, 2, list_to_integer("3")}` we propagate `{any(), any(), 3}`. I think it makes sense. `+` can only return a `float()` in this case.

What can we do for `integer() + any()`, we should return `number()`, I guess.

----

As I've been thinking about these binops after Ulf's work on this part, I'm thinking if it's possible to check these as if they were ordinary functions with intersection types. Then we can perhaps deduplicate some of the code. This of course requires that we check intersection types properly.

```Erlang
%% Example spec
-spec '+'(integer(), integer()) -> integer();
         (float(),   integer()) -> float();
         (integer(), float())   -> float();
         (float(),   float())   -> float().
```

----

@Zalastax I haven't looked much as your code, sorry!

gomoripeti:
I think handling ops as functions is a good idea, Elixir does the same eg: https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/kernel.ex#L1157

Zalastax:
If we agree on representing ops as functions, how would we proceed? It seems to me like a fairly big refactoring. Do we have/need a reserved namespace for these op specs? How can we fix intersections and avoid causing a type checking explosion?

zuiderkwast:
The syntax `erlang:'+'/2` already works. We can use it.

```Erlang
1> erlang:'+'(9, 2).
11
2> lists:foldl(fun erlang:'+'/2, 3, [2,4,7]).
16
```

If we first implement intersection-typed functions so that we can handle a function call like `erlang:'+'(A, B)` in both directions of type checking, add these types to `gradualizer_prelude`, then we can refactor the operators later. How about that?

Type checking explosion is hopefully acceptable. We only check one function at a time. If there's a problem, perhaps we can deal with that later.

Here's another question: Should these operators always be considered typed? Currently, they are considered typed in a typed context and untyped in an untyped context, I think. If they're always considered typed, their return types are propagated in an untyped context.

When propagating/inferring return types, I think we want `any() + any() :: number()`, i.e. a union of the return types of all the clauses...

Zalastax:
> Here's another question: Should these operators always be considered typed? Currently, they are considered typed in a typed context and untyped in an untyped context, I think. If they're always considered typed, their return types are propagated in an untyped context.

I think they should always be considered typed - just as if the operators were implemented as typed functions. What an untyped context is and where errors should be reported needs to be added to [the wiki](https://github.com/josefs/Gradualizer/wiki/Gradual-typing).

> When propagating/inferring return types, I think we want any() + any() :: number(), i.e. a union of the return types of all the clauses...

Indeed!

test/known_problems/should_fail/int_op.erl
```diff
@@ -0,0 +1,11 @@
+-module(int_op).
+-export([plusfloatany/2, divfloatany/2]).
+
+-spec plusfloatany(float(), any()) -> integer().
+plusfloatany(X, Y) ->
+    A = X + Y,
+    A.
+
```
josefs: The function above passes for the same reason as `lambda_wrong_args.erl`. The root of the problem is that we don't do a good job with variables and match statements. Integer operators has nothing to do with the problem.
Perhaps we should have a separate file in `known_problems/should_fail` which just isolates the problem with variables.

test/known_problems/should_pass/arith_op.erl
```diff
@@ -0,0 +1,8 @@
+-module(arith_op).
+-export([plusintfloat/2, plusfloatint/2]).
+
+-spec plusintfloat(integer(), float()) -> float().
+plusintfloat(X, Y) -> X + Y.
+
+-spec plusfloatint(float(), integer()) -> float().
+plusfloatint(X, Y) -> X + Y.
```
josefs: We don't have subtyping between `integer()` and `float()` because it is possible to detect whether a number is a float or an integer. E.g.:
`case 5 of 5.0 -> true; 5 -> false end.`

Given that we don't have subtyping I think it's debatable whether these examples should actually pass. It would be somewhat inconsistent to have subtyping only in the case of arithmetic operators and nowhere else.

josefs:
Can you explain to me what you mean by the operators being typed/untyped?

The way I think about it is this: if there is type information available for an operator, either because we know what result type we want, or we know the type of some argument, we make sure to propagate that information as best we can.

Sure, as Ulf has demonstrated, sometimes it can be difficult to propagate as much information as we would like. But I don't see the need to characterize operators as being either typed or untyped. They just propagate as much type information as is feasible.

test/known_problems/should_fail/int_op.erl
Zalastax: That seems like a good idea!

test/known_problems/should_pass/arith_op.erl
Zalastax: To me it's not debatable that `5.0 + 5` evaluates to `10.0` which has type `float()`. I still haven't read up on the relevant papers (please add to Wiki!) so I generally go by intuition. I get very suspicious when we want to throw away type information (in this case making the codomain wider than it needs to) or make restrictions that make Erlang patterns more similar to a statically typed language (restricting andalso + orelse to work only on booleans).

I really like the spec that @zuiderkwast suggested and it doesn't look to me like we need subtyping to implement it.

zuiderkwast:
@josefs 
> Can you explain to me what you mean by the operators being typed/untyped?

By untyped context, I mean an expression which is handled by `type_check_expr` and by typed I mean `type_check_expr_in`.

In many test cases, we have used a variable to make the typechecker use `type_check_expr`, because we know it works that way, e.g.

```Erlang
-spec f() -> atom().
f() ->
    A = 5.0 + 5,
    A.
```

If `+` was treated as a function with a spec like the one above (with return types `float()` and `integer()` in different clauses), `type_check_expr` would have returned one of these types, but for operators, `type_check_expr` returns `any()` if the operands are `any()`. (Correct me if I'm wrong!)

Perhaps the problem is how we handle variables, as you say.

test/known_problems/should_pass/arith_op.erl
josefs: Yeah, this is not about subtyping as you say. I apologize for the brainfart.

The spec that @zuiderkwast proposed would probably work. I see two risks:

- Performance. When a function with an intersection type is called we go through every type and checks all its arguments with all those types. If we have complex arithmetic expressions we get an exponential blowup here. That makes me nervous.
- Quality of error messages. I haven't checked lately what kind of errors we get from intersection types but it'd be good to make sure that whatever error messages we get are good enough.

The performance problem might be the killer for using the specs. If we implement specialized handling of the types of the arithmetic operators we can make it considerably more efficient.

josefs:
> In many test cases, we have used a variable to make the typechecker use type_check_expr, because we know it works that way

It's better to not have a spec, or a spec which returns `any()` in order to get the typechecker to use `type_check_expr`.

> If + was treated as a function with a spec like the one above (with return types float() and integer() in different clauses), type_check_expr would have returned one of these types, but for operators, type_check_expr returns any() if the operands are any(). (Correct me if I'm wrong!)

Good point. We're overly eager to return `any()` for operators. We return `any()` if ANY of the arguments are `any()` when we should only return it if ALL the arguments are `any()`. That's clearly a bug.

zuiderkwast:
> Good point. We're overly eager to return any() for operators. We return any() if ANY of the arguments are any() when we should only return it if ALL the arguments are any(). That's clearly a bug.

Seems reasonable! Perhaps we should do the same for spec'ed functions, i.e. return `any()` if all operands are `any()`, even if the function has a spec! Then I guess we can get rid of the unwanted type errors in untyped code, e.g. the error we currently get when checking an spec'less function like this one:

```Erlang
f() -> integer_to_list(foo) + 42.
```

test/known_problems/should_pass/arith_op.erl
Zalastax: > When a function with an intersection type is called we go through every type and checks all its arguments with all those types. If we have complex arithmetic expressions we get an exponential blowup here. That makes me nervous

Shouldn't we at most reach the type `integer() | float()`? If we can't handle intersection types as simple as these, how can we expect to handle any intersection types? In any case, I'd like us to try this out and see if performance suffers (let's benchmark!).

Zalastax:
I'd like to try implementing intersection-typed functions; any pointers?

I know that one issue is the representation since intersection types don't exist, only intersection of specs, right?
Any papers you know of for doing this effectively?
I haven't looked at where in the code this should be implemented so that info would give me a good base as well (I'm guessing at least compat_type, glb, the two directions for functions).

josefs:
@Zalastax what about you give specs just as usual for these opertors, like so:

```
-spec erlang:'+'(integer(), integer()) -> integer();
                (float(), float()) -> float().
```

The you'll have to make sure that these specs are picked up when typechecking the operator.

Zalastax:
@josefs we could go for those temporarily while intersection types are buggy! I do however eventually want the more detailed specs (also for andalso etc.) since I've written code that would fail with the simpler specs.

josefs:
@Zalastax sorry I don't follow. You talk about simpler and detailed specs. What are you referring to?

Zalastax:
Simple:
```Erlang
-spec erlang:'+'(integer(), integer()) -> integer();
                (float(), float()) -> float().
```
Detailed:
```Erlang
-spec '+'(integer(), integer()) -> integer();
         (float(),   integer()) -> float();
         (integer(), float())   -> float();
         (float(),   float())   -> float().
```

Thinking about it a bit more, I don't think we can go for the simple specs since it will give type errors for quite common constructions. We **need** mixing of float and integer. I suggest we wait: close this PR, fix intersection types, and then reconsider using the detailed specs.

src/typechecker.erl
```diff
@@ -1689,33 +1689,48 @@ type_check_rel_op(Env, Op, P, Arg1, Arg2) ->
...
-            {Ty
-            ,union_var_binds(VB1, VB2, Env#env.tenv)
-            ,constraints:combine(Cs1, Cs2)}
-    end.
-
-type_check_int_op(Env, Op, P, Arg1, Arg2) ->
+    TNumber = type(number),
+    TFloat = type(float),
+    expect_expr_subtype(Arg1, Ty1, TNumber, #tenv{}),
+    expect_expr_subtype(Arg2, Ty2, TNumber, #tenv{}),
```
josefs: You're dropping the constraints from `expect_expr_subtype`. They should be collected and combined with the other constraints.

src/typechecker.erl
```diff
@@ -1689,33 +1689,48 @@ type_check_rel_op(Env, Op, P, Arg1, Arg2) ->
...
-        {type, _, Ty, []} when Ty == float orelse Ty == number ->
-            throw({type_error, int_error, Op, P, Ty1, Ty2});
-        Ty ->
-            {Ty
-            ,union_var_binds(VB1, VB2, Env#env.tenv)
-            ,constraints:combine(Cs1, Cs2)}
-    end.
+    TInteger = type(integer),
+    expect_expr_subtype(Arg1, Ty1, TInteger, #tenv{}),
+    expect_expr_subtype(Arg2, Ty2, TInteger, #tenv{}),
```
josefs: See above comment about dropping constraints.

josefs:
Oh, I see. I didn't mean that the type I specified to be taken seriously :-) I always intended that we should aim for the detailed specs.

I agree that it'd be good to close this PR as a first step. 

What is it about intersection types that needs fixing? As far as I remember they should work just fine. The only problem is that they can cause performance problems: a call to a function with an intersection type must be checked against all of the specified types. That can be expensive, especially if we have nested function calls.

Zalastax:
@josefs, I just tried it out and intersection types don't need fixing for call sites apparently. Defining functions with intersections don't work (pattern matching doesn't yet subtract from the intersection), which might be why @zuiderkwast thought intersections need fixing. @UlfNorell [thought that  we can't check the intersection type needed for andalso](https://github.com/josefs/Gradualizer/issues/23#issuecomment-440156237), but I tried it out just now and it works.
```Erlang
-spec andalso(true, T) -> T;
             (false, T) -> false.
```

I'll go ahead and replace some custom operator code with specs!

PR #157, opened 2019-02-06
Fix badmatch in union var binds

gomoripeti:
The resulting type of glb is not necessarily the same as the second type.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3646,7 +3646,7 @@ union_var_binds([], _) ->
     #{};
 union_var_binds(VarBindsList, TEnv) ->
     % TODO: Don't drop the constraints
-    Glb = fun(_K, S, T) -> {T, _Cs} = glb(S, T, TEnv), T end,
+    Glb = fun(_K, Ty1, Ty2) -> {Ty, _Cs} = glb(Ty1, Ty2, TEnv), Ty end,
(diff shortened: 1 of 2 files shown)


src/typechecker.erl
```diff
@@ -3646,7 +3646,7 @@ union_var_binds([], _) ->
     #{};
 union_var_binds(VarBindsList, TEnv) ->
     % TODO: Don't drop the constraints
-    Glb = fun(_K, S, T) -> {T, _Cs} = glb(S, T, TEnv), T end,
+    Glb = fun(_K, Ty1, Ty2) -> {Ty, _Cs} = glb(Ty1, Ty2, TEnv), Ty end,
```
gomoripeti: actually why do we use GLB here?
GLB is some kind of intersection and would make sense in code like this:
```
V = Expr1 :: Ty1,
V = Expr2 :: Ty2
```
but in an example like below we should rather take the union of the types (hence the name `union_var_binds`)
```
    case ... of
        true ->  V = Expr1;
        false -> V = Expr2
    end
```
All in all in most cases I think we need the intersection, in sequential code. We need to take the union on branching only. I think we should create two separate helper functions for these.

Is this the right thinking?

gomoripeti: thx for the confirming feedback :) I will address this in a later PR or just create an issue if I wont have time soon.

PR #158, opened 2019-02-09
Override specs for proplists

zuiderkwast:
Another step in #39.

This silences some errors especially often reported for `proplists:get_value/2,3` returning `term()` rather than `any()`.

For reference:

* https://github.com/erlang/otp/blob/master/lib/stdlib/src/proplists.erl
* http://erlang.org/doc/man/proplists.html

### DIFF
--- a/src/gradualizer_prelude.erl
+++ b/src/gradualizer_prelude.erl
@@ -99,3 +99,14 @@
 -spec lists:zipwith (fun((X, Y)    -> T), [X], [Y])      -> [T].
 -spec lists:zipwith3(fun((X, Y, Z) -> T), [X], [Y], [Z]) -> [T].
 
+%% Proplists: Changing term() to any() in the most frequently used functions.
+-spec proplists:delete(any(), list()) -> list().
+-spec proplists:get_all_values(any(), list()) -> list().
+-spec proplists:get_value(any(), list()) -> any().
+-spec proplists:get_value(any(), list(), any()) -> any().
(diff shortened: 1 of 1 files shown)


Zalastax:
Shouldn't we use a type variable instead? The original specs don't, but we can be better than that!

zuiderkwast:
> Shouldn't we use a type variable instead?

Where exactly do you want a type variable?

Zalastax:
The list should have a variable that is a subtype of list, I guess `list<T>` and the value type `T`. This would preserve the type of the list argument + force list and term to be compatible (why would you delete with a value of an incompatible type?)

zuiderkwast:
That's not an anwer to my "where exactly" question. Are you talking about delete/2? A proplist can contain atoms and tuples of any size. Please suggest better specs. I can't see them.

Zalastax:
In all of them!
E.g.
```Erlang
-spec proplists:delete(T, [T]) -> [T].
-spec proplists:get_all_values(T, [T]) -> [T].
-spec proplists:get_value(T, [T]) -> T.
```

zuiderkwast:
Those are clearly incorrect types. A proplist is not just a list of keys.

Zalastax:
Oh, of course, sorry! And looking at the code, the tuples can have any size (seems like they really should be lists instead). I believe the types you suggest are as good as we can do without making changes to the type system (e.g. adding variable sized tuples)

zuiderkwast:
Yes, they are not made to be typed. New code is typically using maps instead of proplists.

PR #159, opened 2019-02-11
Fix crashes

gomoripeti:
and some (maybe unnecessary) improvements to `verbose`

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1244,7 +1244,7 @@ solve_bounds(TEnv, Defs, [{acyclic, X} | SCCs], Acc) ->
 			  end,
 			  {type(term), constraints:empty()}, Tys1);
 	  _NoBoundsForX ->
-	      type(any) %% or should we return term()?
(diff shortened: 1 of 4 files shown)


josefs:
Good stuff. It'd be good to fix the dialyzer error though.

src/typechecker.erl
```diff
@@ -1288,14 +1294,20 @@ subst_ty(_, Ty) -> Ty.
 %% Returns the type of the expression, a collection of variables bound in
 %% the expression together with their type and constraints.
 %-spec type_check_expr(#env, any()) -> { any(), #{ any() => any()}, #{ any() => any()} }.
-type_check_expr(Env, {var, P, Var}) ->
+type_check_expr(Env, Expr) ->
+    Res = {Ty, _VarBinds, _Cs} = do_type_check_expr(Env, Expr),
+    ?verbose(Env, "~s: Inferred type of ~s :: ~s~n",
```
zuiderkwast: We only "infer" when the `infer` option is on. Otherwise, I think we should use the word "propagate" here.

PR #160, opened 2019-02-12
Add specs for functions in the erlang module

josefs:
The erlang module has plenty of functions which need to have their
specs fixed.

I've left out two of them because they have large specs and I didn't
have the energy to go through and fix them. It will probably not be a
big issue since these functions are very rarely used.

### DIFF
--- a/src/gradualizer_prelude.erl
+++ b/src/gradualizer_prelude.erl
@@ -2,12 +2,10 @@
 
 %% This module contains specs to replace incorrect or inexact specs in OTP.
 
--type ext_binary() :: binary().
-
 -spec erlang:apply(function(), [any()]) -> any().
--spec erlang:apply(atom(), function(), [any()]) -> any().
--spec erlang:binary_to_term(ext_binary()) -> any().
--spec erlang:binary_to_term(ext_binary(), [safe | unsafe]) -> any() | {any(), pos_integer()}.
+-spec erlang:apply(module(), atom(), [any()]) -> any().
+-spec erlang:binary_to_term(binary()) -> any().
+-spec erlang:binary_to_term(binary(), [safe | used]) -> any() | {any(), pos_integer()}.
 -spec erlang:element(pos_integer(), tuple()) -> any().
 -spec erlang:erase() -> [{any(), any()}].
 -spec erlang:erase(any()) -> any() | undefined.
@@ -41,8 +39,8 @@
 				  , arity() | [any()]
 				  , [{file, string()}|{line, pos_integer()}]}].
 -spec erlang:hd([A, ...]) -> A.
--spec erlang:max(any(), any()) -> any().
--spec erlang:min(any(), any()) -> any().
+-spec erlang:max(A, B) -> A | B.
+-spec erlang:min(A, B) -> A | B.
 -spec erlang:port_call(port() | atom(), integer(), any()) -> any().
 %% TODO: erlang:process_info
 -spec erlang:put(any(), any()) -> any().
@@ -50,13 +48,13 @@
       Class      :: error | exit | throw,
       Reason     :: any(),
       Stacktrace :: [{module(), atom(), arity() | [any()]} |
-		     {function(), [any()]}] |
-		    [{module(), atom(), arity() | [any()], [{atom(), any()}]} |
-		     {function(), [any()], [{atom(), any()}]}].
+                     {function(), [any()]}] |
+                    [{module(), atom(), arity() | [any()], [{atom(), any()}]} |
+                     {function(), [any()], [{atom(), any()}]}].
 -spec erlang:send(pid() | port() | atom() | {atom(), node()}, any()) -> any().
 -spec erlang:send(pid() | port() | atom() | {atom(), node()}, any()
-		 , [nosuspend | noconnect])
-		 -> ok | nosuspend | noconnect.
+                 ,[nosuspend | noconnect])
+                -> ok | nosuspend | noconnect.
 %% TODO: erlang:system_info({allocator, atom()) -> [term()]
 %% TODO: erlang:system_info({allocator_sizes, atom()) -> [term()]
 %% TODO: erlang:system_info(os_monotonic_time_source) -> [{atom(), term()}].
--- a/src/gradualizer_prelude.erl
+++ b/src/gradualizer_prelude.erl
@@ -2,8 +2,70 @@
 
 %% This module contains specs to replace incorrect or inexact specs in OTP.
 
+-type ext_binary() :: binary().
+
+-spec erlang:apply(function(), [any()]) -> any().
+-spec erlang:apply(atom(), function(), [any()]) -> any().
+-spec erlang:binary_to_term(ext_binary()) -> any().
+-spec erlang:binary_to_term(ext_binary(), [safe | unsafe]) -> any() | {any(), pos_integer()}.
+-spec erlang:element(pos_integer(), tuple()) -> any().
+-spec erlang:erase() -> [{any(), any()}].
+-spec erlang:erase(any()) -> any() | undefined.
+-spec erlang:fun_info(function()) -> [{Item, any()}] when
+      Item :: arity
+	    | env
+	    | index
+	    | name
+	    | module
+	    | new_index
+	    | new_uniq
+	    | pid
+	    | type
+	    | uniq.
+-spec erlang:fun_info(function(), Item) -> {Item, any()} when
+      Item :: arity
+	    | env
+	    | index
+	    | name
+	    | module
+	    | new_index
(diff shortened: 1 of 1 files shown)


src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
...
+-spec erlang:send(pid() | port() | atom() | {atom(), node()}, any()) -> any().
+-spec erlang:send(pid() | port() | atom() | {atom(), node()}, any()
+		 , [nosuspend | noconnect])
+		 -> ok | nosuspend | noconnect.
+%% TODO: erlang:system_info({allocator, atom()) -> [term()]
+%% TODO: erlang:system_info({allocator_sizes, atom()) -> [term()]
+%% TODO: erlang:system_info(os_monotonic_time_source) -> [{atom(), term()}].
+%% TODO: erlang:system_info(os_system_time_source) -> [{atom(), term()}].
+%% TODO: erlang:system_info(Item :: c_compiler_used) -> {atom(), term()}.
+%% TODO: erlang:system_info(Item :: check_io) -> [term()]
```
zuiderkwast: Intersection type for this one? If it's incomplete, just add a catch-any clause `(any()) -> any()` in the end.

src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
...
+-spec erlang:raise(Class, Reason, Stacktrace) -> no_return() when
+      Class      :: error | exit | throw,
+      Reason     :: any(),
+      Stacktrace :: [{module(), atom(), arity() | [any()]} |
+		     {function(), [any()]}] |
+		    [{module(), atom(), arity() | [any()], [{atom(), any()}]} |
+		     {function(), [any()], [{atom(), any()}]}].
+-spec erlang:send(pid() | port() | atom() | {atom(), node()}, any()) -> any().
+-spec erlang:send(pid() | port() | atom() | {atom(), node()}, any()
+		 , [nosuspend | noconnect])
```
zuiderkwast: Indentation looks odd here. Tabs?

src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
...
+-spec erlang:get() -> [{any(), any()}].
+-spec erlang:get(any()) -> any() | undefined.
+-spec erlang:get_keys() -> [any()].
+-spec erlang:get_stacktrace() -> [{ module()
+				  , atom()
+				  , arity() | [any()]
+				  , [{file, string()}|{line, pos_integer()}]}].
 -spec erlang:hd([A, ...]) -> A.
+-spec erlang:max(any(), any()) -> any().
+-spec erlang:min(any(), any()) -> any().
```
zuiderkwast: Can't we have `(A, B) -> A|B` for min and max?

src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
 
 %% This module contains specs to replace incorrect or inexact specs in OTP.
 
+-type ext_binary() :: binary().
+
+-spec erlang:apply(function(), [any()]) -> any().
+-spec erlang:apply(atom(), function(), [any()]) -> any().
+-spec erlang:binary_to_term(ext_binary()) -> any().
+-spec erlang:binary_to_term(ext_binary(), [safe | unsafe]) -> any() | {any(), pos_integer()}.
```
zuiderkwast: The options are `safe | used`, not ~~`unsafe`~~, according to http://erlang.org/doc/man/erlang.html#binary_to_term-2

src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
 
 %% This module contains specs to replace incorrect or inexact specs in OTP.
 
+-type ext_binary() :: binary().
+
+-spec erlang:apply(function(), [any()]) -> any().
+-spec erlang:apply(atom(), function(), [any()]) -> any().
```
gomoripeti: `function()` means a fun object, I think this should be `atom()` (but could use `module()` as first arg as that is an alias for `atom()`)

src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
...
+-spec erlang:max(any(), any()) -> any().
+-spec erlang:min(any(), any()) -> any().
+-spec erlang:port_call(port() | atom(), integer(), any()) -> any().
+%% TODO: erlang:process_info
+-spec erlang:put(any(), any()) -> any().
+-spec erlang:raise(Class, Reason, Stacktrace) -> no_return() when
+      Class      :: error | exit | throw,
+      Reason     :: any(),
+      Stacktrace :: [{module(), atom(), arity() | [any()]} |
+		     {function(), [any()]}] |
```
zuiderkwast: Indentation with tabs?

src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
...
+-spec erlang:get_keys() -> [any()].
+-spec erlang:get_stacktrace() -> [{ module()
+				  , atom()
+				  , arity() | [any()]
+				  , [{file, string()}|{line, pos_integer()}]}].
 -spec erlang:hd([A, ...]) -> A.
+-spec erlang:max(any(), any()) -> any().
+-spec erlang:min(any(), any()) -> any().
+-spec erlang:port_call(port() | atom(), integer(), any()) -> any().
+%% TODO: erlang:process_info
```
zuiderkwast: Why not temporarily spec it as `any() -> any()` and keep the todo?

src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
...
+	    | uniq.
+-spec erlang:get() -> [{any(), any()}].
+-spec erlang:get(any()) -> any() | undefined.
+-spec erlang:get_keys() -> [any()].
+-spec erlang:get_stacktrace() -> [{ module()
+				  , atom()
+				  , arity() | [any()]
+				  , [{file, string()}|{line, pos_integer()}]}].
 -spec erlang:hd([A, ...]) -> A.
+-spec erlang:max(any(), any()) -> any().
```
gomoripeti: can't these be made more precise like `[max|min](A, B) -> A | B.` Or the intention of the PR is to eliminate `term()` and the specs can be pimped later?

src/gradualizer_prelude.erl
```diff
@@ -2,8 +2,70 @@
 
 %% This module contains specs to replace incorrect or inexact specs in OTP.
 
+-type ext_binary() :: binary().
```
zuiderkwast: Why not just write `binary()` where this one is used?

src/gradualizer_prelude.erl
gomoripeti: I'm surprised that a fun can be part of a stacktrace but that's what the original doc also says

src/gradualizer_prelude.erl
josefs: Yes, it's defined as a massive intersection type in otp right now. I've listed the types which I found that were problematic but there are many, many more cases. I'm just going to leave this function for another diff.

src/gradualizer_prelude.erl
josefs: Thanks. I struggle a little bit with how to indent here but I've fixed the problem you pointed out at least

src/gradualizer_prelude.erl
josefs: Sure.

src/gradualizer_prelude.erl
josefs: Thanks. Fixed.

src/gradualizer_prelude.erl
josefs: Thanks

src/gradualizer_prelude.erl
josefs: The reason I kept the `ext_binary` name is that it is used in the documentation to indicate a special form of binary. But for Gradualizer it doesn't really matter. I'll change it.

src/gradualizer_prelude.erl
josefs: Indeed. Thanks!

src/gradualizer_prelude.erl
josefs: Yes, @zuiderkwast made the same comment. I'll change it.

src/gradualizer_prelude.erl
josefs: Yeah, it's weird. I suppose it could be a typo in the otp spec, but I'm not going to investigate it right now.

src/gradualizer_prelude.erl
josefs: I prefer to keep the otp type for the time being.

PR #162, opened 2019-02-17
Fix handling iolist/iodata

gomoripeti:
There was a typo in the alias of iolist.

Not all strings are iolists but in case of a literal string we are able
to decide, so we can propagate a more precise type.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -633,7 +633,10 @@ expand_builtin_aliases({type, Ann, iolist, []}) ->
     Union = [{type, Ann, byte, []},
              {type, Ann, binary, []},
              {type, Ann, iolist, []}],
-    {type, Ann, maybe_improper_list, [{type, Ann, union, Union}]};
+    Tail = [{type, Ann, nil, []},
+            {type, Ann, binary, []}],
+    {type, Ann, maybe_improper_list, [{type, Ann, union, Union},
+                                      {type, Ann, union, Tail}]};
 expand_builtin_aliases({type, Ann, function, []}) ->
     {type, Ann, 'fun', []};
 expand_builtin_aliases({type, Ann, module, []}) ->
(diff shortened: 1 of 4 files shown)


src/typechecker.erl
```diff
@@ -953,45 +957,65 @@ expect_list_type(Union = {type, _, union, UnionTys}, N) ->
...
+infer_literal_string("") ->
+    type(nil);
+infer_literal_string(Str) ->
+    SortedChars = lists:usort(Str),
+    case lists:all(fun(C) -> C =< 255 end, SortedChars) of
+        _ when length(SortedChars) =< 10 ->
+            %% heuristics: if there are not more than 10 different charactes
+            %% list them explicitely as singleton types
+            CharTypes = [{char, erl_anno:new(0), C} || C <- SortedChars],
+            type(nonempty_list, [type(union, CharTypes)]);
```
zuiderkwast: Normalize the union to so ranges become ranges?

src/typechecker.erl
```diff
@@ -953,45 +957,65 @@ expect_list_type(Union = {type, _, union, UnionTys}, N) ->
...
     {[type(any) | AccTy], AccCs};
-expect_list_union([], AccTy, AccCs, _NoAny, _N) ->
+expect_list_union([], AccTy, AccCs, _NoAny, _N, _TEnv) ->
     {AccTy, AccCs}.
 
+infer_literal_string("") ->
+    type(nil);
+infer_literal_string(Str) ->
+    SortedChars = lists:usort(Str),
+    case lists:all(fun(C) -> C =< 255 end, SortedChars) of
```
zuiderkwast: If we return a range from lowest to highest, we don't have to care about 255...

test/should_fail/iodata_fail.erl
```diff
@@ -0,0 +1,13 @@
+-module(iodata_fail).
+
+-export([utf/0]).
+
+%% an iodata can only contain bytes, no utf characters
+utf() ->
+    %% the last character is hexadecimal 256
```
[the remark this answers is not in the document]
gomoripeti: you are right, utf is the encoding, I meant code points above 255.

src/typechecker.erl
gomoripeti: good suggestion

src/typechecker.erl
gomoripeti: I dont think I get what you mean. we care about 255 as upper limit so that the `char()` fits in a `byte()` and the string is an `iolist()`
maybe you meant the arbitrary limit of 10 different characters?

zuiderkwast: No, it's about 255. After sorting, if it's more than 10, we can return

```Erlang
type(range, [hd(SortedChars), lists:last(SortedChars)]})
```

If all elements are 0..255, this will result in a subtype of `byte()`, such as `$b..$y` (for some string). This is more specific type than byte(), although not as complex as listing each value.

gomoripeti: gotcha, good idea!

PR #163, opened 2019-02-17
Replace some custom operator code with specs

Zalastax:
As a follow up to #155 I have tried replacing custom operators with specs.
My findings are:
 * Some checks in `test/should_pass/operator_subtypes.erl` pass just because the expression gets typed as `any()`
 * Intersection types *are* bugged, see `test/known_problems/should_pass/intersection.erl`
* Specs for operators need to be added on the side since they are not functions (hence `?BUILT_IN_SPECS`)
* Using specs made it very easy for me to resolve the issues I pointed out in #155
* `arith_op_arg_types` is really nice to have for integer ranges

I suspect that there are several test cases that don't actually check what we expect them to, how can we deal with that?
Could I use an integer subtype constraint to work with integer ranges, e.g. `band(A, A) -> A when A :: integer()`?
Is there anything in `arith_op_arg_types` we can't replace with specs?


src/gradualizer_db.hrl
```diff
@@ -0,0 +1,33 @@
+-define(BUILT_IN_SPECS, #{
```
gomoripeti: why not use `gradualizer_prelude.erl` so they can be in plain string format instead of AST?

src/typechecker.erl
```diff
@@ -1582,13 +1582,18 @@ do_type_check_expr(Env, {op, P, RelOp, Arg1, Arg2}) when
...
-      Op == '+' orelse Op == '-' orelse Op == '*' orelse Op == '/' ->
-    type_check_arith_op(Env, Op, P, Arg1, Arg2);
-do_type_check_expr(Env, {op, P, Op, Arg1, Arg2}) when
+      Op == '+' orelse Op == '-' orelse Op == '*' orelse Op == '/' orelse
       Op == 'div'  orelse Op == 'rem' orelse
       Op == 'band' orelse Op == 'bor' orelse Op == 'bxor' orelse
-      Op == 'bsl'  orelse Op == 'bsr' ->
-    type_check_int_op(Env, Op, P, Arg1, Arg2);
+      Op == 'bsl'  orelse Op == 'bsr'->
+    FunTy = case gradualizer_db:get_spec(erlang, Op, 2) of
```
gomoripeti: superfluous case? could be just matching?

test/known_problems/should_pass/intersection.erl
```diff
@@ -0,0 +1,5 @@
+-module(intersection).
+-export([inc/1]).
+
+-spec inc(number()) -> number().
+inc(N) -> N + 1.
```
gomoripeti: good catch

src/gradualizer_db.hrl
Zalastax: I tried that initially but got errors, e.g. spec for undefined function `erlang:'+'`.

src/typechecker.erl
Zalastax: Absolutely! Will fix while working on getting this to a mergeable state!

src/gradualizer_db.hrl
josefs: I think we should fix the error you got. We should be able to trick gradualizer into thinking that the function `erlang:'+'` exists.
I think it would be a good idea to have these specs written out using their surface level syntax rather than using the internal representation as it is now. It would help  with readability quite a lot!

Zalastax: I'll try to do so if we go for this solution!

src/gradualizer_prelude.erl
```diff
@@ -64,8 +64,6 @@
 -spec erlang:tl([A, ...]) -> [A].
 -spec erlang:tuple_to_list(tuple()) -> list().
 
-
-
 %% change return Val to any() from term()
```
josefs: Perhaps refrain from including whitespace changes to avoid merge problems.

Zalastax: That change was by accident - will re-add these lines in next commit!

src/gradualizer_db.hrl
zuiderkwast: > We should be able to trick gradualizer into thinking that the function `erlang:'+'` exists.

I think it is a compilation error though. Is it? `gradualizer_prelude` is compiled, although it doesn't contain any code.

gomoripeti: it's a vicious circle: we made `gradualizer_prelude` into a compiled erlang module in order to be included in the escript. Although we wouldn't need it to be compiled, just parse it. As a next level of workaround what if we would introduce two new attributes `-gradualizer_spec` and `gradualizer_type` (or similar with shorter names), which would behave exactly as type and spec but without the compiler checking them. So we could include things in there which pass parsing but not the compiler. Initially this would only be used in the preludes.

josefs: > I think it is a compilation error though. Is it? `gradualizer_prelude` is compiled, although it doesn't contain any code.

When I tried adding the typespec below to `gradualizer_prelude` I didn't get any compiler errors. So it seems to me that it should be a feasible approach. @Zalastax, can you confirm based on your experience?
```
-spec erlang:'+'(integer(), integer()) ->
			integer();
		(float(), float()) ->
			float();
		(float(), integer()) ->
			float();
		(integer(), float()) ->
			float().
```

Zalastax: I believed I got a compilation error for exactly that spec (when running the test suite). I'll double check when I have the time.

Zalastax:
The bin operators are the big trouble for this PR. Even if we manage to write specs for them, they will be too huge to check quickly. One trick could be to extend the spec language (only for internal use) and another would be to patch up the results before returning them (this seems hacky). I don't see much use in doing this change for + etc but not for band.

test/known_problems/should_pass/operator_subtypes.erl
```diff
@@ -0,0 +1,22 @@
+-module(operator_subtypes).
+-compile([pos_op1/2, nonneg_op1/2, neg_op1/2, word_op1/1, word_op2/2, word_op3/3]).
+
+-type word16() :: 0..65535.
+
+-spec pos_op1(pos_integer(), 1..10) -> pos_integer().
+pos_op1(X, Y) -> X * Y + 777 bor Y.
```
zuiderkwast: Clauses for `pos_integer()` and `non_neg_integer()` could be added to `+` and `*`, just so we don't break existing rules where it's possible to avoid it.

The special rules for bitwise operations closed under types on the form 0..2<sup>n</sup>-1 aren't possible to express in the type language syntax, so I suggest you keep Ulf's code for now.

Zalastax:
@zuiderkwast my goal is to get rid of the buggy functions `type_check_arith_op`, `type_check_arith_op_in`, `type_check_int_op` and `type_check_int_op_in` in order to reduce bugs and complexity. I don't think I'll be adding value to the codebase unless I manage to fix all affected operators. It's very unfortunate that the spec language can't express what we need for these operators 😭

src/gradualizer_db.hrl
erszcz: I do not get compile time errors with this spec and Erlang 19.3, 20.2, or 21.1.

gomoripeti:
I accidentally pushed "Ready for review", sorry

I just wanted to comment that I agree with @zuiderkwast that there is value in keeping band (and friends) with current implementation and converting the others to spec.
It is also a good idea to extend the type language internally to support band, although it might have the price that it must be supported everywhere in the code. In my view it is totally acceptable to have special erlang implementation for certain "clever" rules. Other candidates are `element/2` and `record_info/2`.

Zalastax:
Seems it can't be changed back either :open_mouth: 

I will agree with both of you once we have a path forward that we all agree on, then I can merge what's here with a calm mind. I would also like to extend the internal type language so I should put in an effort and suggest something.

test/known_problems/should_pass/operator_subtypes.erl
zuiderkwast: For the record, e.g. the type we have for `bxor` is equivallent to the inifite clause spec

```Erlang

-spec erlang:'bxor'(0..16#1, 0..16#1) -> 0..16#1; % 1 bit
-spec erlang:'bxor'(0..16#3, 0..16#3) -> 0..16#3; % 2 bits
-spec erlang:'bxor'(0..16#7, 0..16#7) -> 0..16#7; % 3 bits
-spec erlang:'bxor'(0..16#f, 0..16#f) -> 0..16#f; % 4 bits

-spec erlang:'bxor'(0..16#1f, 0..16#1f) -> 0..16#1f;
-spec erlang:'bxor'(0..16#3f, 0..16#3f) -> 0..16#3f;
-spec erlang:'bxor'(0..16#7f, 0..16#7f) -> 0..16#7f;
-spec erlang:'bxor'(0..16#ff, 0..16#ff) -> 0..16#ff; % 8 bits

-spec erlang:'bxor'(0..16#1ff, 0..16#1ff) -> 0..16#1ff;
-spec erlang:'bxor'(0..16#3ff, 0..16#3ff) -> 0..16#3ff;
-spec erlang:'bxor'(0..16#7ff, 0..16#7ff) -> 0..16#7ff;
-spec erlang:'bxor'(0..16#fff, 0..16#fff) -> 0..16#fff;

-spec erlang:'bxor'(0..16#1fff, 0..16#1fff) -> 0..16#1fff;
-spec erlang:'bxor'(0..16#3fff, 0..16#3fff) -> 0..16#3fff;
-spec erlang:'bxor'(0..16#7fff, 0..16#7fff) -> 0..16#7fff;
-spec erlang:'bxor'(0..16#ffff, 0..16#ffff) -> 0..16#ffff; % 16 bits

-spec erlang:'bxor'(0..16#1ffff, 0..16#1ffff) -> 0..16#1ffff;
-spec erlang:'bxor'(0..16#3ffff, 0..16#3ffff) -> 0..16#3ffff;
-spec erlang:'bxor'(0..16#6ffff, 0..16#7ffff) -> 0..16#7ffff;
-spec erlang:'bxor'(0..16#7ffff, 0..16#fffff) -> 0..16#fffff;

-spec erlang:'bxor'(0..16#1fffff, 0..16#1fffff) -> 0..16#1fffff;
-spec erlang:'bxor'(0..16#3fffff, 0..16#3fffff) -> 0..16#3fffff;
-spec erlang:'bxor'(0..16#6fffff, 0..16#7fffff) -> 0..16#7fffff;
-spec erlang:'bxor'(0..16#7fffff, 0..16#ffffff) -> 0..16#ffffff;

-spec erlang:'bxor'(0..16#1ffffff, 0..16#1ffffff) -> 0..16#1ffffff;
-spec erlang:'bxor'(0..16#3ffffff, 0..16#3ffffff) -> 0..16#3ffffff;
-spec erlang:'bxor'(0..16#6ffffff, 0..16#7ffffff) -> 0..16#7ffffff;
-spec erlang:'bxor'(0..16#7ffffff, 0..16#fffffff) -> 0..16#fffffff;

-spec erlang:'bxor'(0..16#1fffffff, 0..16#1fffffff) -> 0..16#1fffffff;
-spec erlang:'bxor'(0..16#3fffffff, 0..16#3fffffff) -> 0..16#3fffffff;
-spec erlang:'bxor'(0..16#6fffffff, 0..16#7fffffff) -> 0..16#7fffffff;
-spec erlang:'bxor'(0..16#7fffffff, 0..16#ffffffff) -> 0..16#ffffffff; % 32 bits

%% and so on....
```

zuiderkwast: Maybe, if we want to draw a line somewhere, it's enough if we keep only ~10 of these clauses for only the powers-of-two bits (1, 2, 4, 8, 16, 32, 64, 128, 256). That's good enough IMO, if it means we can stick to the -type/-spec notation in general.

For the record # 2: We mentioned that we'll solve constraint solving before changing this, to make sure these specs don't explode performance-wise.

src/gradualizer_db.hrl
erszcz: Also confirming the spec compiles with 24.2.1.

PR #164, opened 2019-02-19
Specs for the qlc module

josefs:


src/gradualizer_prelude.erl
```diff
@@ -170,3 +170,50 @@
...
+%% qlc
+
+-type(cache() :: ets | list | no).
+-type(key_pos() :: pos_integer() | [pos_integer()]).
+-type(max_list_size() :: non_neg_integer()).
+-type(tmp_file_usage() :: allowed | not_allowed | info_msg
+                        | warning_msg  | error_msg).
+-type(tmp_directory() :: [] | file:name()).
+
+-spec qlc:e(query_handle_or_list()) ->
```
gomoripeti: I don't see type `query_handle_or_list()` defined

src/gradualizer_prelude.erl
```diff
@@ -170,3 +170,50 @@
 -spec proplists:lookup_all(any(), list()) -> [tuple()].
 -spec proplists:append_values(any(), list()) -> list().
 -spec proplists:substitute_aliases([{any(), any()}], list()) -> list().
+
+%% qlc
+
+-type(cache() :: ets | list | no).
```
gomoripeti: I guess qlc uses this old style with the parens - we usually don't use it any more.

src/gradualizer_prelude.erl
josefs: Yeah, that was a typo. I've already fixed this

src/gradualizer_prelude.erl
josefs: Thanks! Fixed.

src/gradualizer_prelude.erl
```diff
@@ -170,3 +170,50 @@
...
+%% qlc
+
+-type cache() :: ets | list | no.
+-type key_pos() :: pos_integer() | [pos_integer()].
+-type max_list_size() :: non_neg_integer().
+-type tmp_file_usage() :: allowed | not_allowed | info_msg
+                        | warning_msg  | error_msg.
+-type tmp_directory() :: [] | file:name().
+
+-spec qlc:e(qlc:query_handle_or_list()) ->
```
Zalastax: `qlc:query_handle_or_list()` is not exported so this won't work, right?

https://github.com/erlang/otp/blob/8d6174377c97f16235cd1bb42c3a0fc783720f0e/lib/stdlib/src/qlc.erl#L131

zuiderkwast: ...... we could fix this by adding a flag `-gradualizer([{allow_unexported_remote_types, true}]).`. It would be a hack but it could save us the pain of duplicating large types.

josefs: @Zalastax Good catch!

josefs: @zuiderkwast yeah, that could be a potentially useful thing. For now, I just duplicated the unexported type because it was so small.

PR #166, opened 2019-02-21
Type annotations using '::'/2

zuiderkwast:
Calling `'::'(Expr, "integer()")` has the same effect as calling a typed function such as

```Erlang
-spec int_id(integer()) -> integer().
int_id(N) -> N.
```

### DIFF
--- a/include/gradualizer.hrl
+++ b/include/gradualizer.hrl
@@ -1,0 +1,35 @@
+%% Syntax elements for type annotations
+%%
+%% The macro ?annotate_type/2 can be used to annotate an expression with a
+%% type.  This is useful to add type annotation after fetching a value from an
+%% ets table, received a message on a known form or read from a dict, proplist,
+%% process dictionary, etc.
+%%
+%% The specified type must be compatible with the type detected by Gradualizer.
+%% Otherwise a type error is reported.
+%%
+%%     N = ?annotate_type( Message, non_neg_integer() )
+%%
+%% The macro ?assert_type/2 can be used to override any type detected by
+%% Gradualizer.  For example, the programmer may know that the length of
+%% a list is within a certain range, rather than any non_neg_integer():
+%%
+%%     Arity = ?assert_type( length(Args), arity() )
+%%
+%% The function '::' can also be used directly if the type is quoted:
+%%
+%%     N = '::'(Message, "non_neg_integer()")
+%%
+%% Gradualizer occurences of the '::'/2 function and adjusts type checking
+%% accordingly.  The macros are supplied only for convenience.
+
+-compile({inline, ['::'/2]}).
+
+'::'(Expr, _Type) -> Expr.
+
+%% Type annotation
+-define(annotate_type(Expr, Type), '::'(Expr, ??Type)).
+
+%% Forced type cast, via any()
+-define(assert_type(Expr, Type), '::'('::'(Expr, "any()"), ??Type)).
+
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1418,6 +1418,15 @@ do_type_check_expr(Env, {bin, _, BinElements} = BinExpr) ->
     {RetTy,
      union_var_binds(VarBinds, Env#env.tenv),
      constraints:combine(Css)};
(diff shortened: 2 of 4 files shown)


zuiderkwast:
Thanks for the positive feedback! :-) This solution has actually just been lying around in the back of my head since we came up with it in this thread: https://github.com/josefs/Gradualizer/pull/83#discussion_r233227770

As for the `?refine_type` @gomoripeti suggests, it's a downcast, while `?annotate_type` is an upcast, right?

In the test cases where I used `?assert_type`, I actually wanted `?refine_type`. Specifically, in the example with `length(Args)`, we want to refine `non_neg_integer()` to `arity()` (i.e. 0..255) because in the imagined scenario, we know that Args comes from an AST. As long as we don't have a type of for lists of a certain maximum length, this example shows a valid use for `?refine_type`.

@Zalastax mentioned some time ago that type assertions in TypeScript are upcast/downcast, i.e. the type has to be either a compatible subtype or a supertype of the propagated type. I think this semantics makes  lot of sense. Better avoid the wild anything-to-anything cast when possible.

How about adding another function, e.g. `':::'/2` for upcast/downcast? This means that we require T1 ⪏ T2.

josefs:
I think it'd be useful to have a way of doing a safe-ish downcast, as @Zalastax suggested in some other ticket. @zuiderkwast your suggestion about using `:::/2` could work, but I would prefer if it was a little more visually distinct from `::/2` so that it is easier to tell them apart. Though, I can't really think of a better suggestion myself :-/

This patch also makes the `ty.erl` module obsolete. That module provided a way to add type annotations using regular erlang functions. But it was rather clunky.

Zalastax:
Why don't we take `::/2` as safe-ish downcast? To get unsafe cast you can just cast to `any()` first.

zuiderkwast:
I suggest these names and behaviours:

Macro | Function | Behaviour | Safety level
----|----|----|----
`?annotate_type` | `::` | upcast only (including `any()` to anything) | "safe"
`?assert_type` | `:::` | downcast or upcast | "safe-ish"
Only manually via `any()` | - | unrestricted cast | "unsafe"

gomoripeti:
I wonder why not have just a downcast function instead of "downcast or upcast"? wouldn't it be more explicit from the user? (in addition to the "only upcast" function)

In what scenario does a user need an upcast? isn't it that we only need `any()` => static casting, which can be considered to be both directions. (here I also mean more dynamic to less dynamic eg `{atom(), any(), any()} => {atom(), float(), any()}`)

please enlighten me

if there would be a downcast only function it could be called `:>`.

Zalastax:
@gomoripeti the upcast is primarily used for casting to the type you already have (i.e. just annotating), but it's also useful when you start involving type variables or to get better error messages, e.g. if infer is on `A = ?annotate_type([true], [boolean()]), lists:last(A)`. My argument would be that it doesn't hurt that we do upcast instead of annotate and sometimes it's useful.

We could go for an operator that is downcast only. I believe that @zuiderkwast went for both directions because that's how `as` works in TypeScript and my experience is that it works well. Even though I'm more familiar with what he suggests for `:::`, keeping downcast separate feels like a really good idea: it's more clear when to use which operator and downcasts would stand out more.

zuiderkwast:
@gomoripeti 
> I wonder why not have just a downcast function instead of "downcast or upcast"? wouldn't it be more explicit from the user? (in addition to the "only upcast" function)

We could of course have a `-define(refine_type(Expr, Type), ':>'(Expr, ??Type))` as well, but then we have more macros to mix up. I don't see how forbidding upcasting would help anyone. I think the terminology *type refinement* and *upcast/downcast* is slightly abstract to an avarage user. Better have just one cast.

Casting in many languages allow upcasting and downcasting in one operation but forbids casts between incompatible types (e.g. TypeScript, Java, C). The purpose of annotating an expression is not to do an explicit upcast as @Zalastax says. Upcasting is just a side effect. Upcasting is implicit in almost every language construct anyway (just look at all the places we use `subtype/3` in typechecker.erl).

gomoripeti:
thank you @Zalastax @zuiderkwast , all clear, no need for a downcast-only function

PR #167, opened 2019-02-23
Updates to type annotation/assertion after discussion/review

zuiderkwast:
#166 got merged so here are the discussed updates in a separate PR


include/gradualizer.hrl
```diff
@@ -10,26 +10,28 @@
...
 %%
-%% Gradualizer occurences of the '::'/2 function and adjusts type checking
-%% accordingly.  The macros are supplied only for convenience.
-
--compile({inline, ['::'/2]}).
+%% Gradualizer detects occurences of the functions '::'/2 and ':::'/2 and
+%% adjusts type checking accordingly.  The macros are supplied only for
+%% convenience.
+%%
+-compile({inline, ['::'/2, ':::'/2]}).
```
josefs: It's really helpful that you've written this documentation. I think it would be even more helpful if you further clarified the distinction between `::` and `:::`. Right now it's not clear which one does annotation and which does assertion.

test/should_fail/annotated_types_fail.erl
```diff
@@ -16,7 +16,15 @@ g(X) ->
     N = ?annotate_type(X, integer()),
     atom_or_die(N).
 
-syntax() ->
+-spec horizontal_cast(integer()) -> atom().
+horizontal_cast(N) ->
+    ?assert_type(N, atom()).
+
+syntax_error_1() ->
+    '::'(banana, "not a type").
```
gomoripeti: I guess one of the syntax errors should call the new `':::'`

src/typechecker.erl
```diff
@@ -4155,9 +4181,9 @@ handle_type_error({type_error, map, P, ResTy, MapTy}) ->
 handle_type_error({type_error, mismatch, Ty, Expr}) ->
     io:format("The expression ~s at line ~p does not have type ~s~n",
               [erl_pp:expr(Expr), erl_anno:line(element(2, Expr)), typelib:pp_type(Ty)]);
-handle_type_error({bad_type_annotation, P, TypeStr}) ->
+handle_type_error({bad_type_annotation, TypeLit}) ->
     io:format("The type annotation ~p on line ~p is not a valid type~n",
-              [TypeStr, erl_anno:line(P)]);
+              [erl_pp:expr(TypeLit), line_no(TypeLit)]);
```
gomoripeti: you could add `maybe_format_column` too

include/gradualizer.hrl
zuiderkwast: Oh, that's true.

test/should_fail/annotated_types_fail.erl
zuiderkwast: Yes, that's the line without test coverage.

src/typechecker.erl
zuiderkwast: Maybe. I thought it can be done when column is added to all error messages in a uniform way. One function call `format_location` would be preferable.

zuiderkwast:
> I apologize if you think I was a little trigger happy with merging your previous PR.

No problem. The merged part can be used already. I'm happy you like it. :-)

test/should_fail/annotated_types_fail.erl
zuiderkwast: `::` and `:::` are handled by the same code, but there is one test case for `type_check_expr` and another for `type_check_expr_in`.

PR #168, opened 2019-02-27
Define our own abstract_type()

### LINKED ISSUE #150: Define our own abstract_type()
Currently, we are using `erl_types:abstract_type()` as the type of types. This causes some type errors to be reported when typechecking the gradualizer itself. The problem is that `erl_types:abstract_type()` has had some bugs. We may want to backport the fixes:

* https://github.com/erlang/otp/pull/1853
* https://github.com/erlang/otp/pull/2069
* https://github.com/erlang/otp/pull/2074

We may also want to add our extensions:

* `{type, erl_anno:anno(), intersection, [abstract_type()]}` (#97, #130)
* `bounded_fun` types
* unbounded ranges: `neg_inf...-42`, `42..pos_inf` (comments in #11)
* possibly also `cons/2` (#127)

To fix this, I suggest this method: Copy the type definitions in `erl_types` to a new module, e.g. `gradualizer_type`, export `abstract_type/0` (and possibly some more types), compile and get a list of unused types. Delete those.

Zalastax:
Closes #150

### DIFF
--- a/src/gradualizer_type.erl
+++ b/src/gradualizer_type.erl
@@ -1,0 +1,355 @@
+%% %CopyrightBegin%
+%%
+%% Copyright Ericsson AB 1996-2018. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% %CopyrightEnd%
+
+-module(gradualizer_type).
+
+-export_type([abstract_type/0]).
+
+%% Export the additional types that gradualizer uses
+-export_type([af_character/0]).
+
+%% Start of Abstract Format
+
+-type anno() :: erl_anno:anno().
+
+-type abstract_expr() :: af_literal()
+                       | af_match(abstract_expr())
+                       | af_variable()
+                       | af_tuple(abstract_expr())
+                       | af_nil()
+                       | af_cons(abstract_expr())
+                       | af_bin(abstract_expr())
+                       | af_binary_op(abstract_expr())
+                       | af_unary_op(abstract_expr())
+                       | af_record_creation(abstract_expr())
+                       | af_record_update(abstract_expr())
+                       | af_record_index()
+                       | af_record_field_access(abstract_expr())
+                       | af_map_creation(abstract_expr())
+                       | af_map_update(abstract_expr())
+                       | af_catch()
+                       | af_local_call()
+                       | af_remote_call()
+                       | af_list_comprehension()
+                       | af_binary_comprehension()
+                       | af_block()
+                       | af_if()
+                       | af_case()
+                       | af_try()
+                       | af_receive()
+                       | af_local_fun()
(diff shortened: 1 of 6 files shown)


Zalastax:
@josefs yes, this PR is an enabler for PRs to come.
Not sure if adding an intersection type will be needed to solve #169 but I heavily believe we should add support for intersection types, not just intersections in specs.

josefs:
No, I don't think adding intersection types is needed to solve #169. I think, but haven't checked, the problem lies with how we approximate unions of tuples when pattern matching on them.

src/gradualizer_type.erl
```diff
@@ -0,0 +1,355 @@
...
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% %CopyrightEnd%
+
+-module(gradualizer_type).
+
+-export_type([abstract_type/0]).
```
gomoripeti: you could also export `abstract_expr`, I think it's not necessary for the definition of `abstract_type` but was also buggy in older Erlang versions, so we can use our own definitions too.
we could also export `abstract_pattern() :: af_pattern()` as it would be useful in `-spec add_types_pats(Pats :: [erl_parse:abstract_expr()],...` (and it is not a subtype of `abstract_expr()`)

src/gradualizer_type.erl
```diff
@@ -0,0 +1,355 @@
...
+
+-type af_record_creation(T) ::
+        {'record', anno(), record_name(), [af_record_field(T)]}.
+
+-type af_record_field(T) :: {'record_field', anno(), af_field_name(), T}.
+
+-type af_map_pattern() ::
+        {'map', anno(), [af_assoc_exact(abstract_expr())]}.
+
+-type abstract_type() :: af_annotated_type()
```
gomoripeti: I think we also want to include ` af_constrained_function_type()` in our `abstract_type()` (this can go in a separate commit as it is a modification of the upstream version)

we can possibly have two `type()`s (not sure about the naming)
- one which can come from an AST
- one extended that Gradualizer can handle internally (which have intersection, bounded_fun types, maybe cons type and any other additions)

josefs: @gomoripeti can you expand a bit about what different situations you imagine we should use the two different `type()` types?

gomoripeti: I imagined that certain functions could avoid handling the extended parts like intersection. Those for example which directly receive a type from a `-type` definition from source. Because you cannot express intersection in general there (yet)

But I realise we cannot define it as `extended_type() :: type() | <extra stuff>` as they are recursive, so would need to duplicate all the subdefinitions of `abstract_type()` which is not desirable. Probably better to handle the extra additions (like intersection) everywhere.

So please ignore this idea.

ps: if we would change the internal representation (eg to records) we would know which functions should still handle the AST type and which the internal type. Sometimes it is useful to encode such additional information in the type definitions even if the actual types don't differ. (just like `temperature_C() :: float()` and `temperature_F() :: float()`)

src/gradualizer_type.erl
```diff
@@ -0,0 +1,355 @@
+%% %CopyrightBegin%
```
zuiderkwast: Delete this?

src/gradualizer_type.erl
zuiderkwast: Why isn't a abstract_pattern() a subtype of abstract_expr()? Is it because they are recursive?

src/gradualizer_type.erl
Zalastax: No, don't we need to maintain license notices?

zuiderkwast: Ah, right! This was copied from a file in OTP.

Issue #130, opened 2018-12-31
Subtyping rules for intersection types


gomoripeti:
@zuiderkwast wrote:
What are the subtyping rules for intersection types? Do we need a rule for merging clauses, such as this?
```
A|C :: E
B|D :: F
---
fun((A) -> B; (C) -> D) :: fun((E) -> F)
```
As an example how do we check that:
```
fun( (integer()) -> integer();
     (float())   -> float() ) <: fun( (number()) -> number() )
```

zuiderkwast:
Peter, we should sleep soon

gomoripeti:
:D haha, ... maybe I won't sleep till new year, it's less then 24 hours

zuiderkwast:
OK, let's finish the beta release before the end of the year! :laughing:

zuiderkwast:
Thanks for a good lecture and beautiful ASCII art! :100:

> `A` and `C`  cannot overlap (although it also says you should get a warning if they do, which I'm pretty sure is a lie)

Perhaps we can give a warning for this. It provides a reason to incorporate gradualizer in OTP. :grin:

As long as there is no syntax for writing intersection types as a return type or an argument type, we don't need to support intersection types in the 2nd argument of subtype, I suppose.

Zalastax:
> A and C cannot overlap (although it also says you should get a warning if they do, which I'm pretty sure is a lie)

I've gotten dialyzer warnings about overlapping domains, so I believe the warning is there.

PR #169, opened 2019-02-27
Add known problems for function intersections

Zalastax:
In order to fix this I believe we need to extend the constraint language. [Flow](https://arxiv.org/pdf/1708.08021.pdf) has both disjunction and conjunction. This is crucial in order to not get false positives such as the one in `test/known_problems/should_fail/intersection.erl`.

Also see #130.

### DIFF
--- a/test/known_problems/should_fail/intersection.erl
+++ b/test/known_problems/should_fail/intersection.erl
@@ -1,0 +1,9 @@
+-module(intersection).
+-export([j/1]).
+
+-spec i(a, b) -> {a, b};
+       (d, e) -> {d, e}.
+i(V, U) -> {V, U}.
+
+-spec j({d, b} | {a, e}) -> {a, b} | {d, e}.
+j({V, U}) -> i(V, U).
--- a/test/known_problems/should_pass/intersection.erl
+++ b/test/known_problems/should_pass/intersection.erl
@@ -1,0 +1,21 @@
+-module(intersection).
+-export([f/1, g/1, j/1]).
+
+-spec f(a|b) -> a|b.
+f(V) -> h(V).
+
+-spec g(a|b) -> a|b.
+g(V) ->
+    A = h(V),
+    A.
+
+-spec h(a) -> a;
+       (b) -> b.
+h(V) -> V.
+
+-spec i(a, b) -> {a, b};
+       (d, e) -> {d, e}.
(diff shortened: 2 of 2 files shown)


Zalastax:
I'll merge this soon but want some discussion first.
Am I barking up the right tree regarding constraints?
@gomoripeti @UlfNorell @zuiderkwast

Zalastax:
What we are interested of in flow all resides in [flow_js.ml](https://github.com/facebook/flow/blob/master/src/typing/flow_js.ml) which is well commented! I recommend looking at every comment block in the file since they reveal trade offs/optimizations they've made but in particular look at how they [handle union and intersection types](https://github.com/facebook/flow/blob/fb8a5928debd32cea5169660f5082b5def49bd81/src/typing/flow_js.ml#L7713). I was expecting to find some trace of the constraints from their original paper and a section clearly marked as "solve constraints". Perhaps it's still there or clearly replaced with something else but the vast code size makes it too hard for me to find it.

Zalastax:
I found this comment that I think explain why I don't find a "solver part": https://github.com/facebook/flow/pull/7298#issuecomment-453608319
> the type checker works by adding constraints into a big graph, then solving those constraints. A constraint is either trivially true or false (string <: number is false, empty <: mixed is true) or decomposes into simpler constraints (A => B <: C => D is true if C <: A and B <: D).
>
> These constraints are "evaluated" eagerly as they are seen, but the order they are seen is roughly the source order -- which is not the same as program order. As a design point, any logic that deals with constraints needs to be "monotonic." In other words, the result of type inference / checking should not be sensitive to the order that constraints are evaluated.

P.S.
I've never implemented constraint solving before so I don't feel like I can tackle this without guidance and hand holding. I have no clue how difficult it would be nor if it's worth holding my hand versus just doing it.

josefs:
@Zalastax I'll try to find some time to read the Flow paper. I appreciate that you keep pushing on this issue so that we can get a correct and efficient implementation.

When starting to implement the Gradualizer I had imagined that we could get away with a fairly simple constraint solver with only conjuctions and no negations. But that's clearly not possible.

Zalastax:
@josefs great! Seems like we would need to sit down a few of us and work this out on a whiteboard. Not sure if a video conference can substitute.

PR #170, opened 2019-02-28
Alternative specs for mnesia

josefs:


Zalastax:
Wow! Lots of new specs and even increased coverage!
Your changes seem to be mainly `term() -> any()`, and as far as I can see it looks correct.

Our prelude is growing quite a lot and I'm getting worried about maintaining these types. Should we try to backport these changes?

josefs:
We should definitely change the types upstream. But we want to maintain these types for a while so that people can use Gradualizer with older versions of OTP.
Also, I have been a little overzealous about simplifying the types when putting them into `gradualizer_prelude`. My thinking early on was to eliminate as many type variables as possible because of problems with polymorphism and subtyping. But those were partly resolved by Ulf's way of interpreting bounded type variables. Going forward, I think the right approach is to change these types as little as possible so that we have a greater chance of getting them accepted when submitting them upstream.

zuiderkwast:
According to the official OTP docs, the is no difference between `term()` and `any()` so I can't see that they will accept such a change upstream. First we should make them redefine `any()` to mean "unknown" in the docs.

I was thinking we can perhaps add an option for treating `term()` as `any()` under certain circumstances, perhaps per module or so. I'm just thinking that with such option, we don't need to override as many specs. What do you think?

KennethL:
Since there is no difference between `term()` and `any()` for Dialyzer I don't think there is a problem to change all usage of `term()` to `any()`. And to change the meaning of `any()` to unknown in the docs. 
But when do you from a Gradualizer point of view think it is motivated to use `term()` in a type or spec?
It would be good if you write down a suggestion of improvements we could do in OTP to make the typing more consistent and useful for Gradualizer, still being compatible for Dialyzer and modules referring to OTP types.
The we can decide what we can do in OTP and when. For example after the OTP 22 release in mid May the master branch would be open for changes (aiming for OTP 23 May 2020).

Zalastax:
`term()` should be a rare type, only to be used when you have a function that accepts any Erlang term with no constraints on other inputs / output. Prime example is `io_lib:format/2`. When gradualizer is perfect, using  `term()` gives us the opportunity to warn if the function body doesn't handle every Erlang term. This is in stark contrast to when using `any()` where you can handle only the items you want. I don't know how useful this would be in practice though.

PR #171, opened 2019-03-01
Additions to own type

gomoripeti:
With using our own abstract_expr() type I was hoping to get rid of some self-gradualizing errors, but there must be other reasons (eg see known problem added)

### DIFF
--- a/src/gradualizer_bin.erl
+++ b/src/gradualizer_bin.erl
@@ -7,7 +7,7 @@
 %% of the elements. The returned type is a normalized bitstring type.
 -spec compute_type(ExprOrPat) -> gradualizer_type:abstract_type()
         when ExprOrPat :: {bin, _, _},
-             ExprOrPat :: erl_parse:abstract_expr().
+             ExprOrPat :: gradualizer_type:abstract_expr().
 compute_type(Bin) ->
     View = bin_view(Bin),
     bitstr_view_to_type(View).
--- a/src/gradualizer_type.erl
+++ b/src/gradualizer_type.erl
@@ -21,7 +21,7 @@
 -export_type([abstract_type/0]).
 
 %% Export the additional types that gradualizer uses
--export_type([af_character/0]).
+-export_type([abstract_pattern/0]).
 
 %% Start of Abstract Format
 
@@ -56,6 +56,8 @@
                        | af_remote_fun()
                        | af_fun()
                        | af_named_fun().
+
+-type abstract_pattern() :: af_pattern().
 
 -type af_record_update(T) :: {'record',
                               anno(),
--- a/src/gradualizer_type.erl
+++ b/src/gradualizer_type.erl
@@ -18,7 +18,9 @@
 
 -module(gradualizer_type).
 
--export_type([abstract_type/0]).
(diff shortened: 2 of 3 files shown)


src/gradualizer_bin.erl
```diff
@@ -7,7 +7,7 @@
 %% of the elements. The returned type is a normalized bitstring type.
 -spec compute_type(ExprOrPat) -> gradualizer_type:abstract_type()
         when ExprOrPat :: {bin, _, _},
-             ExprOrPat :: erl_parse:abstract_expr().
+             ExprOrPat :: gradualizer_type:abstract_expr() | gradualizer_type:af_pattern().
```
gomoripeti: I'm not sure explicitly added `af_pattern()` is necessary.

src/typechecker.erl
```diff
@@ -85,7 +85,7 @@ compatible(Ty1, Ty2, TEnv) ->
 
 %% The first argument is a "compatible subtype" of the second.
 
--spec subtype(type() | gradualizer_type:af_character(), type(), #tenv{}) -> {true, any()} | false.
```
gomoripeti: `af_character()` is now fully part of `type()`

src/gradualizer_type.erl
```diff
@@ -21,7 +21,7 @@
 -export_type([abstract_type/0]).
 
 %% Export the additional types that gradualizer uses
--export_type([af_character/0]).
+-export_type([af_pattern/0]).
```
zuiderkwast: How about naming the exported type `abstract_pattern`, similarly to the other exported types?

zuiderkwast:
Looks good so far. Is there more to come? bounded_fun, intersection?

src/gradualizer_bin.erl
zuiderkwast: I though pattern is a subtype of expression, syntactically.

test/known_problems/should_pass/pattern_ann_type.erl
```diff
@@ -0,0 +1,13 @@
+-module(pattern_ann_type).
```
josefs: I suggest you put this file addition in a separate patch.

gomoripeti: ok

(It is a simplification of the below error that comes out during self-gradualization, I thought that that error is related to the changes in this PR, but it turns out it is not
```
The pattern {var,_,'_'} on line {3345,14} doesn't have the type Pat :: gradualizer_type:af_pattern()
```
)

src/gradualizer_bin.erl
gomoripeti: semantically they are different, that's why I felt the difference, but indeed syntactically they are not.
I even asked gradualizer... so I will remove this modification
```
1> typechecker:subtype(typelib:parse_type("gradualizer_type:af_pattern()"),typelib:parse_type("gradualizer_type:abstract_expr()"), {tenv, #{}, #{}}).
{true, ...}
```

gomoripeti:
@zuiderkwast exactly bounded_fun, intersection will come too but I'm not sure if/when. so not in this PR.

Issue #172, opened 2019-03-26
dialyzer source code module attribute support


tim2CF:
Hello! I would like to discuss possibility of `dialyzer` source code module attribute support in `gradualizer`. The reason for this are some places, for example in elixir core, like this:

https://github.com/elixir-lang/elixir/blob/a6da9e59b73877b60b97957742552f904b2ced1b/lib/elixir/lib/protocol.ex#L704-L706

which explicitly says us - hey, we know this place is not safe, and we know what we are doing.  Of course this is compromise, but probably this make sense? At the moment `gradualizer` says 

```
The pattern #{'__struct__' := __@1} on line 1 doesn't have the type term()
The clause on line 1 cannot be reached
```

probably because it can't handle dialyzer module attributes

gomoripeti:
Hi @tim2CF 

- we were thinking about adding gradualizer specific attributes.
- in the general case I imagine dialyzer options are not really applicable for gradualizer. however in the special case of `:nowarn_function` it makes sense, and could result in Gradualizer just skipping checking that function all together.
- if Gradualizer would parse and understand a few dialyzer attributes that would reduce a bit of noise and duplication in the code. Although in the general case it would be good not to mix the two and gradualizer to have its separate module-inlined options.

- in this particular example any pattern should be a subtype of `term()` so there must be some other error, I suspect a bug in Gradualizer. Did you run Gradualizer on the `Protocol` module itself or could you post a minimal Elixir code example that triggers the same error?

tim2CF:
Hi again @gomoripeti 
Thanks for response!
This bug appears if I'm trying to define any custom protocol, so minimal Elixir code is 

```elixir
defprotocol Foo do
  def bar(_)
end
```

If we translate generated BEAM file to Erlang, probably problematic place is

```erlang
-spec impl_for(term()) -> atom() | nil.

impl_for(#{'__struct__' := __@2 = __@1})
    when erlang:is_atom(__@2) ->
    struct_impl_for(__@1);
```

Full snipped of decompiled Erlang code is

<details><summary>here</summary>
<p>

```erlang
-file("lib/foo.ex", 1).

-module('Elixir.Foo').

-compile([no_auto_import, debug_info,
	  {inline, [{struct_impl_for, 1}]}]).

-callback bar(t()) -> term().

-spec 'impl_for!'(term()) -> atom().

-spec impl_for(term()) -> atom() | nil.

-spec '__protocol__'(module) -> 'Elixir.Foo';
[...]

PR #173, opened 2019-03-26
Map pattern improvements

gomoripeti:
thanks for @tim2CF for pointing out the issue in #172

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3708,10 +3708,15 @@ add_any_types_pat({bin, _, BinElements}, VEnv) ->
                     end,
                     VEnv,
                     BinElements);
-add_any_types_pat({var, _,'_'}, VEnv) ->
+add_any_types_pat({var, _, '_'}, VEnv) ->
     VEnv;
-add_any_types_pat({var, _,A}, VEnv) ->
-    VEnv#{ A => type(any) };
+add_any_types_pat({var, _, A}, VEnv) ->
+    case VEnv of
+        #{A := _} ->
+            VEnv;
+        _ ->
+            VEnv#{ A => type(any) }
+    end;
 add_any_types_pat({op, _, '++', _Pat1, Pat2}, VEnv) ->
     %% Pat1 cannot contain any variables so there is no need to traverse it.
     add_any_types_pat(Pat2, VEnv);
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3526,6 +3526,19 @@ add_type_pat({record, P, Record, Fields}, Ty, TEnv, VEnv) ->
             {VEnv2, Cs2} = add_type_pat_fields(Fields, Record, TEnv, VEnv),
             {type(none), Ty, VEnv2, constraints:combine(Cs1, Cs2)}
     end;
+add_type_pat({map, _, _} = MapPat, {var, _, Var} = TyVar, _TEnv, VEnv) ->
+    %% FIXME this is a quite rudimentary implementation
+    %% - variables from the map pattern become any()
+    %% - the contraint could contain the map keys
+    Cs = constraints:add_var(
+           Var, constraints:upper(Var, type(map, any))),
+    {type(none), TyVar, add_any_types_pat(MapPat, VEnv), Cs};
+add_type_pat({map, _, _} = MapPat, ?type(term), TEnv, VEnv) ->
+    %% TODO instead of implemented an expect_map_type properly
+    %% just handle the one missing case here
+    %% expect_map_type(term()) would return #{term() => term()}
+    AllMapsTy = type(map, [type(map_field_assoc, [type(term), type(term)])]),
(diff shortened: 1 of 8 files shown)


src/typechecker.erl
```diff
@@ -3526,6 +3526,19 @@ add_type_pat({record, P, Record, Fields}, Ty, TEnv, VEnv) ->
             {VEnv2, Cs2} = add_type_pat_fields(Fields, Record, TEnv, VEnv),
             {type(none), Ty, VEnv2, constraints:combine(Cs1, Cs2)}
     end;
+add_type_pat({map, _, _} = MapPat, {var, _, Var} = TyVar, _TEnv, VEnv) ->
+    %% FIXME this is a quite rudimentary implementation
+    %% - variables from the map pattern become any()
+    %% - the contraint could contain the map keys
+    Cs = constraints:add_var(
```
Zalastax: I think it's fine that we leave this in as a FIXME. We need to redo the constraint system anyways.

test/should_fail/map_pattern.erl
```diff
@@ -9,3 +9,8 @@ f(#{bepa := Bepa}) ->
 
 -spec badkey(#{apa => atom()}) -> ok.
 badkey(#{bepa := _Bepa}) -> ok.
+
+-spec map_term(term() | #{atom() => atom()} | term()) -> any().
```
Zalastax: Shouldn't the spec be:
`-spec map_term(term() | #{atom() => atom() | term()}) -> any().`

I'm confused about what the spec even means. Either the parameter is a `term()` or it's a `#{atom() => atom() | term()}`, but `#{atom() => atom() | term()}` already is a `term()` so that's superfluous, right?

test/should_fail/var_pattern_bound.erl
```diff
@@ -0,0 +1,16 @@
+-module(var_pattern_bound).
+
+-export([f/1]).
+
+-spec f(atom()) -> integer().
+f(Atom) ->
+    case g() of
+        {Atom} ->
```
Zalastax: Testing `{Atom} ->` is great, but given the name of the module I'm expecting a bunch of other tests as well, e.g. `Atom ->`.

test/should_fail/map_pattern.erl
gomoripeti: ops, I was just trying to provoke some other error case in `add_type_pat` and I forgot the type here. the type is normalized anyway so I guess it should be just `term()`.

test/should_fail/var_pattern_bound.erl
gomoripeti: I just wanted to add a negative test that fails without the fix. `Atom ->` already passed because `add_type_pat` for a straight var handles it without calling `add_any_types_pat`. We need a var nested in a pattern to provoke the error. `Atom ->` is already covered according to coverage but I can add one here too, should I?

Zalastax: I think it would be a good idea to add a second, similar, function testing `Atom ->` unless we have a test case that explicitly tests this already.

josefs:
Thanks for responding to the bug report so quickly!

test/should_pass/any_pattern.erl
```diff
@@ -0,0 +1,14 @@
...
+%% This test case just increases coverage of add_any_types_pat
+%% with various patterns.
+%% We would expect that type of I, S and L is integer(), string() and
+%% list() respectively but currently they all become any()
+-spec pat_any(any()) -> ok.
+pat_any({$a}) -> ok;
+pat_any({[]}) -> ok;
+pat_any({<<I>>}) -> I;
+pat_any({S = "string"}) -> S;
+pat_any({"prefix" ++ L}) -> L.
```
Zalastax: Add the cases with I, S, and L to a separate module in the `known_problems/should_fail` instead.

gomoripeti: I am happy to add them, but...
what do we consider as a known problem? definitely false positives (ie Gradualizer emits a type error although there is no type error). Also definitely when Gradualizer propagates the wrong type (that is not compatible with the real type)
But in this case Gradualizer is just not precise enough. There will always be error cases which Gradualizer cannot find, or when Gradualizer propagates a too general type. This I'd consider more like a `known_improvement_opportunity`, and has lower priority than fixing the known bugs.
(for the binary matching there is already a [TODO](https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L3686) in the code)

I don't try to bully here :) just your comment started me thinking. What do you guys think?

(maybe I will just add a `known_todo` dir that behaves exactly like `known_problems` and then problem solved :) )

zuiderkwast: I think we can add it as a known problem. It is not as bad as a false positive, but we don't need to distinguish between more and less serious problems. Just give it a descriptive name.

The TODO for binaries can be known_problem test too if anyone cares to create one.

PR #174, opened 2019-03-29
Refactor operator patterns

gomoripeti:
- use `erl_eval:partial_eval/1` exactly what the compiler also uses
- remove duplicated code of handling integer/float pattern within op patterns

just a bit of ad hoc refactoring

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3554,7 +3554,7 @@ add_type_pat({op, _, '++', Pat1, Pat2}, Ty, TEnv, VEnv) ->
     {_, _, VEnv1, Cs1} = add_type_pat(Pat1, Ty, TEnv, VEnv),
     {_, _, VEnv2, Cs2} = add_type_pat(Pat2, Ty, TEnv, VEnv1),
     {type(none), Ty, VEnv2, constraints:combine(Cs1,Cs2)};
-add_type_pat(Expr={op, P, _Op, _Pat1, _Pat2}, Ty, TEnv, VEnv) ->
+add_type_pat(OpPat = {op, _Anno, _Op, _Pat1, _Pat2}, Ty, TEnv, VEnv) ->
     %% Operator patterns are evaluated at compile-time by the compiler.
     %% So we simply evaluate them and check the type of the resulting value.
     %% This simplified situations like this: suppose Ty = non_neg_integer() and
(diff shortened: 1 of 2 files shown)


src/typechecker.erl
```diff
@@ -3563,39 +3563,30 @@ add_type_pat(Expr={op, P, _Op, _Pat1, _Pat2}, Ty, TEnv, VEnv) ->
...
     add_type_pat(Pat, Ty, TEnv, VEnv);
 
 add_type_pat(Pat, Ty, _TEnv, _VEnv) ->
     throw({type_error, pattern, element(2, Pat), Pat, Ty}).
 
+add_type_pat_literal(Pat, Ty, TEnv, VEnv) ->
+    case erl_eval:partial_eval(Pat) of
+        Pat ->
+            %% illegal, non-constant pattern - this should not happen
+            throw({illegal_pattern, Pat});
```
zuiderkwast: Should not happen for what reason? I guess only hand-coded ASTs can include this. Perhaps we don't need this clause but instead just `Literal when Literal =/= Pat` in the other clause? Then `Pat` gives a case_clause error and code coverage can increase. (I assume this is an uncovered clause.)

gomoripeti: I was thinking about exactly the same when I saw codecov complaining. However at the undefined-records (somewhere in this thread https://github.com/josefs/Gradualizer/pull/147#issuecomment-457855199) we agreed that we assume the least from the AST and (although proactively don't look for such errors) when we hit it, we show a meaningful message.

So I will just add a little test to cover this case similar to this https://github.com/josefs/Gradualizer/blob/master/test/typechecker_tests.erl#L546

zuiderkwast: Sure, if you add a test for this, at least it's not completely dead code. I do think there's a difference to undefined records though.

For the undefined records, it's still valid syntax, so it's still a realistic scenario to hit when we're type checking source files just using the parser, which is something we do support. An AST that the parser cannot produce is not the same; it's not realistic IMO.

We were talking about calling the compiler for the validation step when checking source files, which would catch undefined records, but we didn't implement it yet, did we? If we do, I think we don't need the code for handling undefined records either.

gomoripeti: Actually the parser successfully parses this illegal patter (see the added testcase) only the linter catches it.
If gradualizer will ever support running as a parse_transform (to plug it straight into the compilation pipeline) that is the only scenario when we might receive such unchecked AST, and I think in that case we cannot call out to the compiler/linter. This use case (which we might decide to never implement for some reason or the other) stopped me to implement the erl_lint check for source code, and convinced me that we need to handle every case the parser can possibly spit out.

PR #177, opened 2019-04-04
Small fixes

gomoripeti:


src/typechecker.erl
```diff
@@ -2777,8 +2777,8 @@ type_check_comprehension_in(Env, ResTy, Compr, Expr, P,
             {_VB2, Cs2} = type_check_comprehension_in(NewEnv, ResTy, Compr, Expr, P, Quals),
             {#{}, constraints:combine(Cs1, Cs2)};
         {elem_ty, ElemTy, Cs} ->
-            {_PatTy, _UBound, NewVEnv, Cs2} =
-                add_type_pat(Pat, ElemTy, Env#env.tenv, Env#env.venv),
+            {_PatTys, _UBounds, NewVEnv, Cs2} =
+                add_types_pats([Pat], [ElemTy], Env#env.tenv, Env#env.venv),
```
gomoripeti: Forgot to add it to the commit:
the fix here is that `add_types_pats` normalizes `ElemTy` while `add_type_pat` doesn't.

src/typechecker.erl
```diff
@@ -2823,6 +2823,10 @@ type_check_assocs(Env, [{Assoc, P, Key, Val}| Assocs])
 type_check_assocs(_Env, []) ->
     {[], #{}, constraints:empty()}.
 
+update_map_type({ann_type, _, [_, Ty]}, AssocTys) ->
+    %% FIXME we really should get rid of `ann_type's right in the
+    %% begining or during normalization.
```
zuiderkwast: Better not add this, considering your other PR. :stuck_out_tongue_winking_eye:

src/typechecker.erl
zuiderkwast: Makes sense.

src/typechecker.erl
gomoripeti: I have low hopes of addressing the other PR soon, so I thought until then let's add a quick-fix.

josefs: Right now we ignore `ann_type`s and I think that's the wrong thing to do. We really should check them when they show up, otherwise people are going to be very confused.
But I'm not saying that you have to change that for this PR, but I don't think removing the `ann_type`s is the right way forward.

gomoripeti: let's discuss the role of annotated types in Gradualizer in #180

PR #178, opened 2019-04-04
Normalize away annotated types

gomoripeti:
This small change triggers a new "infinite loop" while self-gradualizing
possibly again in GLB. A somewhat minimal extract of
`typechecker:type_check_comprehension_on` which triggers the behaviour
is put into `should_pass/glb_ann_types.erl`.

Should not be merged! Needs to be investigated.

### DIFF
--- a/test/should_pass/glb_ann_types.erl
+++ b/test/should_pass/glb_ann_types.erl
@@ -1,0 +1,20 @@
+-module(glb_ann_types).
+
+-compile(export_all).
+
+-type type() :: erl_parse:abstract_type().
+
+-define(TYPE_MOD, erl_parse).
+%%-define(TYPE_MOD, gradualizer_type).
+
+-spec type_check_comprehension_in(
+        Expr       :: erl_parse:abstract_expr(),
+        Qualifiers :: [ListGen |Filter|BinGen]) ->
+     {map(), constraints:constraints()}
+       when
+        ListGen :: {generate, erl_anno:anno(), ?TYPE_MOD:abstract_expr(), ?TYPE_MOD:abstract_expr()},
+        BinGen  :: {b_generate, erl_anno:anno(), ?TYPE_MOD:abstract_expr(), ?TYPE_MOD:abstract_expr()},
+        Filter  :: ?TYPE_MOD:abstract_expr().
+type_check_comprehension_in(Expr, [{generate, P_Gen, Pat, Gen} | Quals]) ->
+    {#{}, constraints:empty()}.
+
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -627,6 +627,8 @@ normalize({op, _, _, _Arg1, _Arg2} = Op, _TEnv) ->
     erl_eval:partial_eval(Op);
 normalize({type, Ann, range, [T1, T2]}, TEnv) ->
     {type, Ann, range, [normalize(T1, TEnv), normalize(T2, TEnv)]};
+normalize({ann_type, _, [_Name, T]}, TEnv) ->
+    normalize(T, TEnv);
 normalize(Type, _TEnv) ->
     expand_builtin_aliases(Type).
 


test/should_pass/glb_ann_types.erl
```diff
@@ -0,0 +1,20 @@
...
+
+-compile(export_all).
+
+-type type() :: erl_parse:abstract_type().
+
+-define(TYPE_MOD, erl_parse).
+%%-define(TYPE_MOD, gradualizer_type).
+
+-spec type_check_comprehension_in(
+        Expr       :: erl_parse:abstract_expr(),
```
zuiderkwast: `?TYPE_MOD` here?

test/should_pass/glb_ann_types.erl
```diff
@@ -0,0 +1,20 @@
+-module(glb_ann_types).
+
+-compile(export_all).
+
+-type type() :: erl_parse:abstract_type().
```
zuiderkwast: `?TYPE_MOD` here?

zuiderkwast:
> `gradualizer_tests: type_check_dir_test...*timed out*`

Perhaps there is another infinite loop?

zuiderkwast:
Btw, I think the title is slightly confusing. I'm note 100% sure about the terminology, but my intuitions says that type annotations is stuff annotated with types, while annotated types is types annotated with stuff.

gomoripeti:
I tried hard to get the title right but in that late hour I picked the wrong choice :)

test/should_pass/glb_ann_types.erl
gomoripeti: indeed. (but this whole testcase is just temporary to help track down the issue)

gomoripeti:
`type_check_dir_test` runs all the should_pass testcases, also the one that is a negative example. So (I really hope) there is just one infinite loop.

When I hit this issue I didnt want to go any further before it is addressed (removing ann_type's from everywhere else). This issue is most probably not related to ann_type but until now Gradualizer would error out before reaching the issue.

Zalastax:
Some notes:

If `T` is a type definition `Name :: Type`, where `Name` is a variable and `Type` is a type, then `Rep(T) = {ann_type,LINE,[Rep(Name),Rep(Type)]}` (from [absform](http://www2.erlang.se/documentation/doc-7.1/erts-7.1/doc/html/absform.html)).

```
--------------------------------------------------------------------------------
```

Type variables can be used in specifications to specify relations for the input and output arguments of a function. For example, the following specification defines the type of a polymorphic identity function:
```Erlang
  -spec id(X) -> X.
```
Notice that the above specification does not restrict the input and output type in any way. These types can be constrained by guard-like subtype constraints and provide bounded quantification:
```
  -spec id(X) -> X when X :: tuple().
```
Currently, the :: constraint (read as «is a subtype of») is the only guard constraint that can be used in the when part of a -spec attribute.

## Note
The above function specification uses multiple occurrences of the same type variable. That provides more type information than the following function specification, where the type variables are missing:
```
  -spec id(tuple()) -> tuple().
```
The latter specification says that the function takes some tuple and returns some tuple. The specification with the `X` type variable specifies that the function takes a tuple and returns the same tuple.

However, it is up to the tools that process the specifications to choose whether to take this extra information into account or not.
(from [typespec](http://erlang.org/doc/reference_manual/typespec.html)).

```
----------------------------------------
```

Do we want to take that extra information into account? I suspect that no tool takes advantage of the extra information and that most specs that reuse the same variable does so accidentally. I would opt for ignoring it and lobby for OTP to replace that note to instead say that it doesn't have to return the same thing.

@josefs was this what you meant with checking so they make sense? What do we loose by normalizing?

gomoripeti:
@Zalastax I think there are two things here: type variables in general and annotated types specifically.

Re: annotated types let's discuss in #180

zuiderkwast:
Closing this in favor of #228, which is basically the same change.

PR #179, opened 2019-04-05
Exact type for pseudo-function record_info/2

zuiderkwast:


zuiderkwast:
Sure, I can add error handling for that. Note though that `record_info` isn't a real function, so calling it with anything other than two correct atoms is a compilation error. We can't test invalid record_info the usual way, because rebar compiles the test modules before running the tests.

We have been talking about running the compiler's `basic_validation` step when checking and ASTs from parsed source code, before type checking. If we do, there is no need to handle this in the typechecker and we can also drop any handling of undefined records.

For the record: Even something like `record_info(size, _R = fruitz)` (where fruitz is a valid record name) gives a compilation error:

```
_build/test/lib/gradualizer/test/should_pass/record_info.erl:8: illegal record info
```

In the shell:

```
1> record_info(foo, banana).
* 1: illegal record info
```

Using record_info in a fun reference as in `foo() -> fun record_info/2.` even crashes the compiler:

```
_build/test/lib/gradualizer/test/should_pass/record_info.erl:none: internal error in beam_dead;
crash reason: {case_clause,{'EXIT',{undefined_label,25}}}
```

josefs:
> We have been talking about running the compiler's basic_validation step when checking ASTs from parsed source code, before type checking. If we do, there is no need to handle this in the typechecker and we can also drop any handling of undefined records.

My preference would be for doing our own error checking, when needed. The risk is that adding the basic_validation step would slow down gradualizer needlessly.

The compiler crash you've found is pretty bad. Is there a ticket for that in otp?

zuiderkwast:
I've added error checking for calls to record_info/2 so only valid atoms are accepted, matching what the compiler accepts. I filed [ERL-907](https://bugs.erlang.org/browse/ERL-907) for the compiler crash.

PR #181, opened 2019-04-14
 Option to specify include path for checking erl files

zuiderkwast:
I tried adding `-include("gradualizer.hrl")` to some gradualizer module to be able to use the type assertion macros, but when typechecking source code, `epp` doesn't find the include path. Therefore, I'm adding `-I` to the CLI (same as for `erlc`) and `{i, Dir}` (same as for `compile`) to be able to specify include paths. Now `make gradualize` can call gradualizer with `-I include`.

In the second commit, some false positives reported as type errors are hidden using `?assert_type/2`.

### DIFF
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -1,13 +1,15 @@
 %%% @doc Main external API of the Gradualizer
 %%%
 %%% The functions `type_check(file|module|dir)' accept the following options:
+%%% - `{i, Dir}': Include path for `-include' and `-include_lib' when checking
+%%%   Erlang source files. Specify multiple times for multiple include paths.
 %%% - `stop_on_first_error': if `true' stop type checking at the first error,
 %%%   if `false' continue checking all functions in the given file and all files
 %%%   in the given directory.
 %%% - `print_file': if `true' prefix error printouts with the file name the
 %%%   error is from.
-%%% - `crash_on_error': if `true` crash on the first produced error
-%%% - `return_errors': if `true`, turns off error printing and errors
+%%% - `crash_on_error': if `true' crash on the first produced error
+%%% - `return_errors': if `true', turns off error printing and errors
 %%%   (in their internal format) are returned in a list instead of being
 %%%   condensed into a single ok | nok.
 %%% - `fmt_location': how to format location when pretty printing errors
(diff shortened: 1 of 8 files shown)


src/gradualizer.erl
```diff
@@ -1,13 +1,15 @@
...
+%%% - `{i, Dir}': Include path for `-include' and `-include_lib' when checking
+%%%   Erlang source files. Specify multiple times for multiple include paths.
 %%% - `stop_on_first_error': if `true' stop type checking at the first error,
 %%%   if `false' continue checking all functions in the given file and all files
 %%%   in the given directory.
 %%% - `print_file': if `true' prefix error printouts with the file name the
 %%%   error is from.
-%%% - `crash_on_error': if `true` crash on the first produced error
-%%% - `return_errors': if `true`, turns off error printing and errors
+%%% - `crash_on_error': if `true' crash on the first produced error
```
Zalastax: use matching ticks, either ' or `

gomoripeti: this is good old edoc notation for verbatim http://erlang.org/doc/apps/edoc/chapter.html#verbatim-quoting

src/typelib.erl
```diff
@@ -44,6 +46,7 @@ pp_type(Type = {type, _, bounded_fun, _}) ->
                           [{capture, all_but_first, list}, dotall]),
     "fun(" ++ S ++ ")";
 pp_type({var, _, TyVar}) ->
+    %% TODO: In type(), TyVar should be an atom but we use a string.
```
gomoripeti: hm, I guess this only works because `io:format("~s, [atom])` works. This should really be an `atom_to_list/1`.
good catch.

(or we can support both atom and string to avoid creating a lot of new atoms, but then it needs to be reflected both in type and code)

src/gradualizer.erl
Zalastax: That's super weird, but okay

Zalastax: That's super weird, but okay

zuiderkwast: Yeah, it's weird but probably older than markdown. If we'd want to generate HTML docs from this, we could publish it on github pages or on Hex.pm if we make a hex package. If we don't, there is no need to use edoc syntax.

src/typelib.erl
zuiderkwast: Yes, you're right. What's a high realistic number of generated type variables for a function? If it's never more than say 5 in practice, perhaps we can generate var names in a way so that they can be reused, rather than creating new ones for each function. Then we can use atoms and avoid extending the type.... Otherwise, we better extend our type.

I found it using `make gradualize`, after fixing the previous type error in this function using type assertion. :-)

PR #182, opened 2019-04-15
Allow wildcards in path args for cli

gomoripeti:
Note the wildcards mustn't be expanded by the shell because the `-pa` option only takes one argument
```
gradualizer -pa myapp/deps/\*/ebin ...
or
gradualizer -pa 'myapp/deps/*/ebin' ...
```

Should wikdcards be supported by the new `-I` option as well?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,11 +47,11 @@ escript:
 
 .PHONY: gradualize
 gradualize: escript
-	./gradualizer -pa src/ -I include src/*.erl
+	./gradualizer -pa src/ -I include -- src/*.erl
 
 .PHONY: nocrashongradualize
 nocrashongradualize: escript
-	./gradualizer -pa src/ src/*.erl; \
+	./gradualizer -pa src/ -I include -- src/*.erl; \
     EXIT=$$?; \
     if [ $$EXIT -eq 0 ] || [ $$EXIT -eq 1 ]; then \
         exit 0; \
--- a/src/gradualizer_cli.erl
+++ b/src/gradualizer_cli.erl
@@ -12,10 +12,11 @@ handle_args(Args) ->
     HasHelp = proplists:get_bool(help, Opts),
     HasVersion = proplists:get_bool(version, Opts),
     Status = if
-        HasHelp -> print_usage(), ok;
-        HasVersion -> print_version(), ok;
-        true -> gradualizer:type_check_files(Rest, Opts)
-    end,
+                 HasHelp -> print_usage(), ok;
+                 HasVersion -> print_version(), ok;
+                 Rest =:= [] -> erlang:error("No files specified to check (try --)");
+                 true -> gradualizer:type_check_files(Rest, Opts)
+             end,
     case Status of
         ok -> halt(0);
         nok -> halt(1)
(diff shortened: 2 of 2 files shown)


Zalastax:
I think it should be added to `-I` as well.

gomoripeti:
it makes sense and would be more convenient to use but isn't it ambiguous to have `... -pa path1 path2 path3` (is path2 to be added to the code path or should it be checked?). Maybe can decide based on if path2 is a directory.

gomoripeti:
(wrong button)

zuiderkwast:
Isn't it kind of inconvenient if you can't write `gradualizer -pa ebin src/*.erl` anymore? I think take while `is_dir` seems like nice heuristics.

Is it even allowed to add paths that don't exist? Is it allowed with erl and erlc?

If we want to allow -pa with dirs that don't exist, what about type checking files that don't exist? If that's not allowed, perhaps we use `splitwith` with a predicate like "is not regular file and doesn't start with a dash"?

Erl doesn't have this problem because it doesn't accept plain filenames arguments. It does need to stop when there is another parameter starting with a dash though, so it must be some logic for that. How about looking that up for inspiration?

Zalastax:
> Is it even allowed to add paths that don't exist? Is it allowed with erl and erlc?

Yes! I tried it out just now. This makes sense to me.

> what about type checking files that don't exist?

`erlc` fails with `no such file or directory`. This also makes sense to me.

I think the predicate should just be "doesn't start with a dash". Most CLIs seems to work this way and there are workarounds, e.g. [expanding the filename a bit](https://unix.stackexchange.com/a/1529).

gomoripeti:
next try :)

I think it's acceptable to have to use -- in the simplest case of one dir and one file.

(- note1: erlc does not allow multiple dirs after -pa (same way as -I)
 - note2: current master of gradualizer does not allow non-existing dirs, cos `code:add_patha` (singleton) will error out - while `code:add_pathsa` (plural) ignores it silently)

Zalastax:
@gomoripeti good job!
I'm surprised by your note1 since I tested it and thought it worked for erlc. I'll have to check again...

src/gradualizer_cli.erl
```diff
@@ -34,6 +35,9 @@ print_usage() ->
     io:format("Usage: gradualizer [options] [PATH...]~n"),
     io:format("A type checker for Erlang/Elixir~n~n"),
     io:format("       PATH                      Files or directories to type check~n"),
+    io:format("       --                        Signals that no more options will follow. The rest~n"),
+    io:format("                                 of the arguments is treated as filenames, even if~n"),
```
zuiderkwast: is -> are

Makefile
```diff
@@ -47,11 +47,11 @@ escript:
 
 .PHONY: gradualize
 gradualize: escript
-	./gradualizer -pa src/ -I include src/*.erl
+	./gradualizer -pa src/ -I include -- src/*.erl
```
zuiderkwast: `--` is not really needed here until we allow multiple dirs for `-I`. Perhaps it's still good to use `--`, for anyone looking at the Makefile for inspiration.

src/gradualizer_cli.erl
gomoripeti: taken verbatim from `erl` doc, quickfixing

Makefile
gomoripeti: yes, it's not necessary but exactly as you say. and I wanted to make it future proof  if we happen to allow multiple dirs in include too.

src/gradualizer_cli.erl
gomoripeti: ...actually the "rest of sg" might be singular :)

Issue #186, opened 2019-05-01
Develop a new, more advanced, constraint system


Zalastax:
As I raised in #169, the constraint system we have is unfortunately not advanced enough to express the patterns we see (as far as I can tell).

The discussion there boils down to the conclusion that my judgement seems correct, that we need to read up on the research that exists, and that this is a substantial task.
This seems like a really fun thing to dig into but I need help to continue.

zuiderkwast:
Where we're collecting the constraints, we'd replace constraints:combine/1 with constraints:union/1 and constraints:intersect/1 and represent them as trees of these internally.

When it comes to solving them, it can still be done after checking a function, I assume. Regarding how to solve them, I don't really know, but a naive brute-force approach could work as a prototype, I think.

PR #188, opened 2019-09-15
Exhaustiveness

josefs:
I decided to try my had at implementing exhaustiveness checking. It is not quite as good as I had hoped due to limitations in type refinement but it works reasonably well and the code is quite short. I think it'll do for now.


zuiderkwast:
Hi Josef! I'll look at the changes tomorrow. Just want to say: Really nice to see that Gradualizer is alive again! We (me and Pierre) were just talking about picking up constraints solving and/or type refinement. In the last Hackathon, I tried running it on our product and found some minor issues and was talking to David and a guy called Yevgeniy who seemed interested in getting involved, or at least in running it in SGSN-MME-AMF. So, I'm looking forward in a Gradualizer sprint until the next code beam! or so.

src/typechecker.erl
```diff
@@ -854,6 +854,9 @@ int_range_to_types({neg_inf, I}) when I > 0 ->
     [type(neg_integer),
      {type, erl_anno:new(0), range, [{integer, erl_anno:new(0), 0}
                                     ,{integer, erl_anno:new(0), I}]}];
+int_range_to_types({neg_inf, I}) -> % I < -1
+    [{type, erl_anno:new(0), range, [{integer, erl_anno:new(0), neg_inf}
```
zuiderkwast: The integer `neg_inf`, interesting! Perhaps add this to `gradualizer_type:abstract_type()`?

src/typechecker.erl
```diff
@@ -3167,14 +3175,27 @@ check_clauses(Env, [], {var, _, TyVar}, Clauses) ->
...
         lists:foldl(fun (Clause, {VBs, Css, RefinedArgsTy}) ->
                             {NewRefinedArgsTy, VB, Cs} =
                                 check_clause(Env, RefinedArgsTy, ResTy, Clause),
                             {[VB | VBs], [Cs | Css], NewRefinedArgsTy}
                     end,
                     {[], [], ArgsTy},
                     Clauses),
+    % Checking for exhaustive patternmatching
+    case {Env#env.exhaust
+	 ,ArgsTy =/= any andalso
```
zuiderkwast: Indentation looks odd. Is it tabs?

src/typechecker.erl
```diff
@@ -73,6 +73,7 @@
              ,tenv             :: #tenv{}
              ,infer    = false :: boolean()
              ,verbose  = false :: boolean()
+	     ,exhaust  = true  :: boolean()
```
zuiderkwast: Tabs?

Btw, are you sure it should be enabled by default? I think in many cases, we can't handle guards nearly as good as we should, so the following would be considered non-exhaustive:

```Erlang
-spec foo(integer()) -> integer().
foo(N) when is_integer(N) -> N.
```

src/typechecker.erl
```diff
@@ -3335,6 +3364,71 @@ pick_one_refinement_each([Ty|Tys], [RefTy|RefTys]) ->
...
+    lists:all(fun refinable/1, Tys);
+refinable({ann_type, _, [_, Ty]}) ->
+    refinable(Ty);
+refinable(_) ->
+    false.
+
+% Given a type, pick a value of that type.
+% Used in exhaustiveness checking to show an example value
+% which is not covered by the cases.
+pick_value(List) when is_list(List) ->
```
zuiderkwast: Perhaps this function can be defined in another module? This module is huge already.

src/typechecker.erl
```diff
@@ -4042,6 +4136,10 @@ print_error(Error, Opts) ->
 handle_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
   when is_tuple(Expression) ->
     print_type_error(Expression, ActualType, ExpectedType, Opts);
+handle_type_error({nonexhaustive, Anno, Example}, Opts) ->
+    io:format("Nonexhaustive patterns starting on line ~p.~n"
+	      "Example values which is not covered: ~p~n"
+	     ,[format_location(Anno, brief, Opts), Example]);
```
zuiderkwast: Note that `format_location(Anno, brief, Opts)` returns the empty string if some option is set to show non-brief location.

test/known_problems/should_fail/intersection.erl
```diff
@@ -1,9 +1,6 @@
 -module(intersection).
--export([j/1]).
+-export([i/2]).
```
zuiderkwast: `i/2` shouldn't fail, should it? It has no type errors IMO. Only `j/1` was supposed to fail here.

josefs:
Victor, thanks for the review! It's nice to hear that you played with gradualizer in the last hackathon! It'd be nice to get the momentum going again and it's great to hear that you guys have plans to do some more hacking!

test/known_problems/should_fail/intersection.erl
josefs: image

Yeah, this file should be removed now.

src/typechecker.erl
josefs: image

src/typechecker.erl
josefs: Yeah.

src/typechecker.erl
josefs: I tried to fix it but I only made things worse. There's some Woodoo going on here. I'll look at it some other day when I'm less tired.

Btw, I avoid doing exhaustiveness checking when there are guards so that shouldn't be a problem. Just think about the fact that all our tests pass with this option on, which made me confident that it can be on by default. But it also probably means that it doesn't actually catch that many cases to begin with.

src/typechecker.erl
josefs: See below. Not sure what's going on...

src/typechecker.erl
josefs: Yeah, maybe that's a good idea actually. We need to be able to talk about ranges which contain negative and positive infinity in order to do accurate type refinement. I initially thought that we only needed these types during intermediate stages of type refinement but you're right that they will leak out into actual types. I'll ponder this some more but I think your suggestion makes sense.

src/typechecker.erl
zuiderkwast: OK, all test cases pass is a good thing (apart from one test case where you added a catch-all clause). I didn't see you skip exhaustiveness checking when there are guards, but that's very good! I think it's a good feature, definitely. Don't get me wrong on that. :-)

The type refinement with guards like the one above is something I have struggled with before so I can probably pick that up again. Both refinement and exhaustiveness would benefit from solving it so the reason for solving it is double.

src/typechecker.erl
josefs: This looks fine if I go to the "Files changed" tab so it's probably just a bug in github.

src/typechecker.erl
josefs: Yeah, improving type refinement with guards is a really nice project!

Again, the whitespace issue seems to be a bug with github.

josefs:
Regarding the example from Code Beam 2018: yeah, you're right. I was totally confused when talking about that example.

src/typechecker.erl
zuiderkwast: Yeah. We should have some way to "rebase" our version of the abstract type when it's updated in OTP as well... It's not good if we divert too much...

zuiderkwast:
> Regarding the example from Code Beam 2018: yeah, you're right. I was totally confused when talking about that example.

OK, all good then. :-) Now that we're at it, the similar example from the same talk for which there were many questions:

```Erlang
-spec append(list(A), list(A)) -> list(A).
append([], T) -> T;
append([X | Xs], T) -> [X | append(Xs, T)].
```

You said something along the lines that the elements in the lists have to be of the same type. But with singleton types, a call like `append([1, 2, 3, 4], [5])` breaks this since the first list has element type `1..4`, the second list `5` and the result list has elements `1..5`. So, is this a type error? and should the correct spec be `-spec append(list(A), list(B)) -> list(A | B)`?

src/typechecker.erl
josefs: I've now added `neg_inf` and `pos_inf` to `gradualizer_type:abstract_type()`.

Regarding keeping up with the OTP version of abstract type; sure, we should try to keep up. But how often does it actually change? Perhaps there is a relatively easy to make some kind of diff between our type and OTP and also different version of OTP and track the changes that way. Though I don't think this will be a huge problem in practice.

josefs:
> You said something along the lines that the elements in the lists have to be of the same type.

This was an oversimplification on my part. I chose to be slightly inaccurate for pedagogical purposes.

A more accurate statement would have been that they all have to be subtypes of the same type. When we check elements of a list we only require it to be a subtype of the element type. So `append([1, 2, 3, 4], [5])` can typecheck just fine by picking an appropriate supertype for the element type like `integer()` or even `[1..5]` as you suggest.

zuiderkwast:
> A more accurate statement would have been that they all have to be subtypes of the same type. 

OK, thanks. I was just stuck around the universal quantification. To involve "for all", I guess it be should be read like this: "For all A, if the arguments are of type A, the result has to be of type A"? (This would mean that for all common supertypes of [1..4] and [5] for the example with append.)

(Or generally, "F(A<sub>1</sub>, ..., A<sub>n</sub>) -> T" should be read "for all A<sub>i</sub>, if the V<sub>i</sub> :: A<sub>i</sub>, then F(V<sub>1</sub>, ..., V<sub>n</sub>) -> T".) Just trying to understand the constraints solving fully.

src/typechecker.erl
```diff
@@ -73,20 +73,21 @@
...
     case {subtype(Ty1, Ty2, TEnv), subtype(Ty2, Ty1, TEnv)} of
-        {{true, C1}, {true, C2}} ->
-            {true, constraints:combine(C1,C2)};
-        {false, T={true, _C2}} ->
-            T;
-        {T={true, _C1}, false} ->
-            T;
-        {false, false} ->
-            false
+	{{true, C1}, {true, C2}} ->
```
zuiderkwast: Here tabs are introduced in a new place. :-)

src/typechecker.erl
```diff
@@ -854,6 +854,9 @@ int_range_to_types({neg_inf, I}) when I > 0 ->
     [type(neg_integer),
      {type, erl_anno:new(0), range, [{integer, erl_anno:new(0), 0}
                                     ,{integer, erl_anno:new(0), I}]}];
+int_range_to_types({neg_inf, I}) -> % I < -1
```
zuiderkwast: Why not add `when I < -1` as an actual guard, instead of a comment?

src/typechecker.erl
```diff
@@ -864,9 +867,14 @@ int_range_to_types({0, pos_inf}) ->
     [type(non_neg_integer)];
 int_range_to_types({1, pos_inf}) ->
     [type(pos_integer)];
+int_range_to_types({I, pos_inf}) -> % I > 1
```
zuiderkwast: Actual guard instead of comment?

src/typechecker.erl
```diff
@@ -3167,14 +3175,27 @@ check_clauses(Env, [], {var, _, TyVar}, Clauses) ->
...
-    {VarBindsList, Css, _RefinedArgsTy} =
-        lists:foldl(fun (Clause, {VBs, Css, RefinedArgsTy}) ->
-                            {NewRefinedArgsTy, VB, Cs} =
-                                check_clause(Env, RefinedArgsTy, ResTy, Clause),
-                            {[VB | VBs], [Cs | Css], NewRefinedArgsTy}
-                    end,
-                    {[], [], ArgsTy},
-                    Clauses),
+    {VarBindsList, Css, RefinedArgsTy} =
+	lists:foldl(fun (Clause, {VBs, Css, RefinedArgsTy}) ->
```
zuiderkwast: Tabs

src/typechecker.erl
```diff
@@ -4042,6 +4098,10 @@ print_error(Error, Opts) ->
 handle_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
   when is_tuple(Expression) ->
     print_type_error(Expression, ActualType, ExpectedType, Opts);
+handle_type_error({nonexhaustive, Anno, Example}, Opts) ->
+    io:format("Nonexhaustive patterns starting on line ~p.~n"
+	      "Example values which is not covered: ~p~n"
+	     ,[format_location(Anno, verbose, Opts), Example]);
```
zuiderkwast: Now, `--fmt-location verbose` works but running with `--location-fmt brief` doesn't show any line number at all.

The way to do this is slightly confusing, but if option `--fmt-location brief` is specified, `format_location(Anno, verbose, Opts)` returns nothing and `format_location(Anno, brief, Opts)` returns something. Vice versa if the option is `--fmt-location verbose`. The brief output is always the first in line since it's on the form ":line" or ":line:col" (to be shown directly after the filename, like the output of `grep -n`). Try with the CLI to be sure.

src/typechecker.erl
josefs: Oh, so github transforms tags to three spaces. That's why it looks different. 
Also, I thought that I had tabs turned off in my editor but apparently not. I'll make sure to fix that.

src/typechecker.erl
josefs: Code is better than comments :-)
I somehow was thinking that it'd be easier for the exhaustiveness checker to see that the cases were exhaustive this way. But I was clearly thinking backwards, we shouldn't necessarily make things easy for the exhaustiveness checker, we should make it understand the code we write.

src/typechecker.erl
gomoripeti: Indeed it's currently quite inconvenient, but that was the only way I could implement to get the intended output :(

Zalastax:
For the append example, right now we don't really check the type of A right? We need constraint solving for that. This type checks:
```Erlang
-spec append(list(A), list(A)) -> list(A).
append([], T) -> T;
append([X | Xs], T) -> [X | append(Xs, T)].

-spec foo([integer()], [integer()]) -> [string()].
foo(As, Bs) ->
    append(As, Bs).
```

zuiderkwast:
> For the append example, right now we don't really check the type of A right? We need constraint solving for that.

I though we already have some `known_problem` for this, but I can't find any. Anyway, we should probably continue this discussion to #186 or so. (It was all my fault bringing the append example here. Sorry!)

src/typechecker.erl
```diff
@@ -4099,9 +4099,11 @@ handle_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
   when is_tuple(Expression) ->
     print_type_error(Expression, ActualType, ExpectedType, Opts);
 handle_type_error({nonexhaustive, Anno, Example}, Opts) ->
-    io:format("Nonexhaustive patterns starting on line ~p.~n"
+    io:format("~sNonexhaustive patterns~s~n"
 	      "Example values which is not covered: ~p~n"
```
zuiderkwast: Singular or plural? (values ... is)

josefs: Thanks!

erlandsona:
So I'm not getting any errors for this from gradualixer... which seems to be pointing to this master branch so I'm curious what might be missing?

gomoripeti:
hi @erlandsona, the `mix.lock` file might lock gradualizer version to an earlier one, did you try `mix deps.update` ?

josefs:
@erlandsona, note that there are plenty of cases where the  exhaustiveness checker doesn't yet complain where you think it might. Can you give an example of what you tried that you think should result in an error?

erlandsona:
I'm expecting this to throw a non-exhaustive pattern match error but it does not.
https://github.com/erlandsona/opaque/blob/master/lib/contact.ex#L130

I'm doing a talk on types in Elixir (specifically opaque types for building smarter primitives) for my local meetup and I've been dying for this functionality to work so I can show how you can force implementation details into the constructor of an opaque type to help mitigate cyclomatic complexity by centralizing the rules about primitives into "smarter" primitives and composing those instead of modeling everything with booleans.

For the most part the opaqueness thing works but dialyzer will "success" type and incorrect usage of a type if it can just find a single path through the code that will pass. <- this is why I want exhaustiveness checking against types to enforce the checking that an opaque type is not being misused.

josefs:
Thanks for the example, @erlandsona. I don't know how modules in Elixir are compiled but a not so wild guess is that they're compiled to maps. Unfortunately, gradualizers exhaustiveness check doesn't handle maps at the moment. It's rather non-trivial to implement good support for maps and I'm afraid that it's not high on my personal list of projects. @zuiderkwast is this something you will be looking into?

zuiderkwast:
Not very soon, I'm afraid. I want to cover basic guards such as `is_integer/1` first.

erlandsona:
Ah fair point... 
So, in Elixir there's this concept of Structs which are built on top of Maps but specify a `__struct__` key with the Module name. 
For reference: https://elixir-lang.org/getting-started/structs.html.
This would enable them to be differentiated by Gradualizer as different from a traditional Map... I don't assume that makes much of a difference tho :/

josefs:
Using the `__struct__` key might be useful during type checking. Though, again, I'm afraid that we're a little understaffed to take on this problem right now. It would be great to have an Exilir point man that could look into Elixir related issues for gradualizer.

PR #189, opened 2019-09-17
Fix broken CLI features and add CLI tests

zuiderkwast:
E.g. checking a directory and --no-print-file were broken.

Since eunit captures stdout, it's quite hard to test the CLI from
within eunit. Instead, simple CLI tests are added in the Makefile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ matrix:
 # The latest OTP version is special and therefore explicitly
 # included above instead.
 otp_release:
-  - 22.0
+
   - 21.0
   # Last minor version of older OTP releases
   - 20.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ matrix:
 # The latest OTP version is special and therefore explicitly
 # included above instead.
 otp_release:
-
+  - 22.0
   - 21.0
   # Last minor version of older OTP releases
   - 20.3
--- a/Makefile
+++ b/Makefile
@@ -65,9 +65,38 @@ clean:
 	rm -rf _build
 	rm -f gradualizer
 
-.PHONY: tests
-tests:
+.PHONY: tests eunit cli-tests
+tests: eunit cli-tests
+
+eunit:
 	rebar3 eunit
+
+cli-tests: escript
+	# CLI test cases
+	# 1. When checking a dir; printing filename is the default
+	./gradualizer test/dir \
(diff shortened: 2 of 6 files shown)


zuiderkwast:
Code coverage is decreased according to codecov, ironically, although this PR is actually about testing something untested.

gomoripeti:
great that you are fixing this @zuiderkwast 
- I guess the assumption was that if you gradualize multiple files you always want to see which one has the error...but maybe not
- Unfortunately eunit does not have an `?assertCmdNotMatchOutput` macro but you could build something maybe from `?cmd` http://erlang.org/doc/apps/eunit/chapter.html#macros-for-running-external-commands
Alternatively I would move the cli tests into a standalone shell script file to have a main makefile that is easy to overview

josefs:
> * Unfortunately eunit does not have an `?assertCmdNotMatchOutput` macro but you could build something maybe from `?cmd` http://erlang.org/doc/apps/eunit/chapter.html#macros-for-running-external-commands

Why not submit a PR to eunit?

zuiderkwast:
Thx for the reviews!

> Why not submit a PR to eunit?

I thought about that. Someone submitted an issue for eunit not supporting specs with multiple clauses, but it was closed with reason ain't-gonna-fix, so I was discouraged to improve eunit since it's obviously not prioritized. Maybe if I fix it myself, it would be accepted though... Maybe if I have time some day.

> but you could build something maybe from `?cmd`

Do you think coverage works for code runing in a separate OS process, separate BEAM node? Probably not, right?

> Alternatively I would move the cli tests into a standalone shell script file to have a main makefile that is easy to overview

That's a possibility. I like Makefile for tests though, because once a line fails, the rest of the test cases don't run, by the design of make. A separate Makefile would be good though, but let's do that when/if the test code grows a bit larger, ok?

zuiderkwast:
> I guess the assumption was that if you gradualize multiple files you always want to see which one has the error...but maybe not

Fair enough, but the CLI was (and is) using `type_check_files`, even when checking a single file.

gomoripeti:
> ... the CLI was (and is) using `type_check_files`, even when checking a single file.
Oops, great that we have tests for cli now, and such oversight will be caught in the future.

And indeed coverage is not measured even if cmd is started from eunit, so no gain there, your makefile/perl solution is the right tool for the job.

👍

PR #190, opened 2019-09-24
Misc fixes

Zalastax:
I ran Gradualizer on some internal modules and found an errors and some crashes. This fixes the crashes and adds tracking of the errrors.

### DIFF
--- a/src/gradualizer_db.erl
+++ b/src/gradualizer_db.erl
@@ -344,19 +344,18 @@ import_module_from_erl(Mod, State) ->
 import_erl_files([File | Files], State) ->
     EppOpts = [{includes, guess_include_dirs(File)}],
     {ok, Forms} = epp:parse_file(File, EppOpts),
-    [{attribute, _, file, _} | Forms1] = Forms,
-    {attribute, _, module, Module} = lists:keyfind(module, 3, Forms1),
-    check_epp_errors(File, Forms1),
-    import_erl_files(Files, import_absform(Module, Forms1, State));
+    {attribute, _, module, Module} = lists:keyfind(module, 3, Forms),
+    check_epp_errors(File, Forms),
+    import_erl_files(Files, import_absform(Module, Forms, State));
 import_erl_files([], St) ->
     St.
 
 -spec import_beam_files([file:filename() | binary()], state()) -> {ok, state()} | gradualizer_file_utils:parsed_file_error().
 import_beam_files([File | Files], State) ->
     case gradualizer_file_utils:get_forms_from_beam(File) of
-        {ok, [{attribute, _, file, _} | Forms1]} ->
-            {attribute, _, module, Module} = lists:keyfind(module, 3, Forms1),
-            import_beam_files(Files, import_absform(Module, Forms1, State));
+        {ok, Forms} ->
+            {attribute, _, module, Module} = lists:keyfind(module, 3, Forms),
+            import_beam_files(Files, import_absform(Module, Forms, State));
         Error = {Status, _} when (Status /= ok) ->
             Error
     end;
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -596,7 +596,7 @@ normalize({user_type, P, Name, Args} = Type, TEnv) ->
                 opaque ->
                     Type;
                 not_found ->
-                    throw({undef, user_type, {Module, Name, length(Args)}})
+                    throw({undef, user_type, P, {Module, Name, length(Args)}})
             end;
         none ->
             %% Local user-defined type
@@ -4077,6 +4077,11 @@ handle_type_error({undef, user_type, Anno, {Name, Arity}}, Opts) ->
     io:format("~sUndefined user type ~p/~p~s~n",
               [format_location(Anno, brief, Opts),
                Name, Arity,
+               format_location(Anno, verbose, Opts)]);
+handle_type_error({undef, user_type, Anno, {Module, Name, Arity}}, Opts) ->
+    io:format("~sUndefined user type ~p:~p/~p~s~n",
+              [format_location(Anno, brief, Opts),
+               Module, Name, Arity,
                format_location(Anno, verbose, Opts)]);
 handle_type_error({not_exported, remote_type, {{atom, Anno, _} = Module, Name, Arity}}, Opts) ->
     io:format("~sThe type ~s:~s/~p~s is not exported~n",
--- a/priv/test/undefined_user_type_helper.erl
+++ b/priv/test/undefined_user_type_helper.erl
@@ -1,0 +1,4 @@
+-module(undefined_user_type_helper).
+-export_type([j/0]).
+
+-type j() :: undefined_type().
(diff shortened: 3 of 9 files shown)


src/typechecker.erl
```diff
@@ -4078,6 +4078,11 @@ handle_type_error({undef, user_type, Anno, {Name, Arity}}, Opts) ->
               [format_location(Anno, brief, Opts),
                Name, Arity,
                format_location(Anno, verbose, Opts)]);
+handle_type_error({undef, user_type, Anno, {Module, Name, Arity}}, Opts) ->
```
zuiderkwast: The code which was supposed to handle this type error is on line 4070, the clause with the pattern `{undef, Type, {{atom, _, Module}, {atom, _, Name}, Arity}}`. That clause used to match in some earlier OTP version, where the module and function name were represented as `{atom, _, _}`s.

So, please remove the duplicate one and preferable add a test case so we can see if it works in all our supported OTP versions.

src/gradualizer_db.erl
```diff
@@ -344,19 +344,18 @@ import_module_from_erl(Mod, State) ->
 import_erl_files([File | Files], State) ->
     EppOpts = [{includes, guess_include_dirs(File)}],
     {ok, Forms} = epp:parse_file(File, EppOpts),
-    [{attribute, _, file, _} | Forms1] = Forms,
```
zuiderkwast: Good.

What I think we should use these for is when code is `-include`'d, in the error messages. We should keep track of in which file each piece of code is.

src/typechecker.erl
Zalastax: I tried to create a test but any test I could come up with is stopped by the compiler. I think these cases can#t ever happen so I#m removing them.

test/known_problems/should_pass/listspecs.erl
```diff
@@ -0,0 +1,7 @@
+-module(listspecs).
+-export([flatten/0]).
+
+% Defining types in gradualizer_prelude doesn't work yet
+-spec flatten() -> [integer()].
+flatten() ->
+    lists:flatten([[1,2],[3,4]]).
```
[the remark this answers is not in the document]
zuiderkwast:
```Erlang
-spec flatten(DeepList) -> List when
      DeepList :: [term() | DeepList],
      List :: [term()].
```

Is it this recursion we can't handle?

src/typechecker.erl
zuiderkwast: But it happened, right?

I think it can't happen for beam files, but for erl files it can, because of the compiler checks that come after parsing.

I'm pretty sure gradualizer_db can return not_found in some case. If it can't, we should remove that code too.

test/known_problems/should_pass/listspecs.erl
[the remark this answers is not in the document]
Zalastax: No, it's due to us not supporting `-type` here: https://github.com/josefs/Gradualizer/blob/eb41baca61ee966a0ff6afa59fa7184a08d08743/src/gradualizer_prelude.erl#L98

test/known_problems/should_pass/nested_pattern_match.erl
```diff
@@ -0,0 +1,12 @@
+-module(nested_pattern_match).
+-compile([debug_info]).
+-export([f/1]).
+
+-record(s, {a :: c | d}).
+-record(r, {a :: #s{} | boolean()}).
+
+-spec f(#s{}) -> boolean().
+f(R) -> if
+    (R#r.a)#s.a == c -> true;
```
josefs: I must be missing something because I don't understand why this should type check. In the spec, the argument is declared to have type `#s{}` where the field `a` has type `c | d`. But here it is used at type `#r` where the `a` field has type `#s{} | boolean`. How is that correct?

Zalastax: Whoops, `#s{}` started out as `#r{}` and then I played around and committed the wrong thing.

test/known_problems/should_pass/nested_pattern_match.erl
```diff
@@ -0,0 +1,12 @@
+-module(nested_pattern_match).
+-compile([debug_info]).
+-export([f/1]).
+
+-record(s, {a :: c | d}).
+-record(r, {a :: #s{} | boolean()}).
+
+-spec f(#r{}) -> boolean().
+f(R) -> if
+    (R#r.a)#s.a == c -> true;
```
zuiderkwast: I think this is a guards specific problem. Exceptions (e.g. runtime type errors) in guards are treated as just no match and this is used in practice in code such as this. So I think we should complain only if it can never match, as you say.

Nested patterns for pattern matching works, I hope, doesn't it?

src/typechecker.erl
```diff
@@ -594,9 +594,7 @@ normalize({user_type, P, Name, Args} = Type, TEnv) ->
                 {ok, T} ->
                     normalize(T, TEnv);
                 opaque ->
-                    Type;
-                not_found ->
```
zuiderkwast: I still think this can happen, but we cannot test this, because gradualizer_db needs a module which doesn't compile, because it references undefined types internally. Rebar doesn't allow that, because it thinks all erl files under test/ are test suites.

As long as we want to handle errors which the compiler also catches, which we have talked about before, I think we should come up with a way to test this rather than removing the handling. The alternative is to run the compiler's checks whenever we parse erl files.

Zalastax: How can we get a module that does not compile into gradualizer_db? If we can't do this via eunit/rebar3, a similar trick as for when testing the CLI could be used.

zuiderkwast: `gradualizer -pa bad_module.erl -- module_calling_bad_module.erl`

test/known_problems/should_pass/listspecs.erl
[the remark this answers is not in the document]
zuiderkwast: This test case is quite similar to this existing known problem:

```Erlang
-module(cyclic_otp_specs).

-compile(export_all).

%% We shouldn't fail on cyclic dependencies in OTP specs.
%% Either make them work, or override the specs.

-spec flatten([term()]) -> [term()].
flatten(Xs) -> lists:flatten(Xs).
```

src/typechecker.erl
zuiderkwast: Sorry, `-pa` needs a directory, but you get the idea. (We could put bad modules under priv/test/ or so, to confuse rebar.)

But this doesn't need to be done now, does it? We can open an issue instead.

Zalastax: -pa did not work - it needs a directory of beams. We do have `gradualizer_db:import_erl_files/1` though. By using that directly I was able to trigger the code I removed.

Zalastax:
@zuiderkwast, any more comments?

zuiderkwast:
> 
> 
> @zuiderkwast, any more comments?

Only that the which was intended to catch this originally can never match for `user_type`, so it could be refactored, possibly unified with the one you added:

```Erlang
handle_type_error({undef, Type, {{atom, Anno, Module}, {atom, _, Name}, Arity}}, Opts)
  when Type =:= user_type; Type =:= remote_type ->
```

No big deal, so feel free to merge!

PR #191, opened 2019-09-24
Rework how overriden otp types are compiled and loaded

gomoripeti:
This is just a bare minimum to allow compilation on OTP 22.


josefs:
I'm merging since I think this is a fairly urgent PR to get in. I've been holding off on upgrading my erlang installation for this sole reason.

Issue #187, opened 2019-05-25
Gradualizer does not compile with OTP 22


gomoripeti:
The compiler now complains about the preludes trick :(
```
===> Compiling src/gradualizer_prelude.erl failed
src/gradualizer_prelude.erl:5: spec for function erlang:apply/2 from other module
...
```
this was expected to come, from OTP 22 release notes
> --- POTENTIAL INCOMPATIBILITIES -------------------------------------
>
> OTP-15563    Application(s): compiler, stdlib
> Related Id(s): [ERL-845](https://bugs.erlang.org/browse/ERL-845), OTP-15562
>
> Do not allow function specifications for functions
> residing in other modules.

zuiderkwast:
This is pretty serious.

Our list of overriding specs is pretty long and it makes me wonder if it's a practical approach or if we should go for something simpler. How about treating `term()` as `any()` for external functions? It's a bit radical, but `term()` almost never makes sense, so I think it could be a possible solution.

Otherwise, I guess it's possible to come around the problem by importing the specs in some other way. In worst case, we can put them in strings inside the code (e.g. `"-spec erlang:hd([T]) -> T."`) or so.

hauleth:
Another possibility is to move all that definitions upstream.

josefs:
@hauleth, we do plan to improve the typespecs in OTP. But we also want to be able to support a few older versions as well so we need a solution which works with the type specs in OTP right now.

KennethL:
I suggest that you override the otp type specs by using a name convention.
Put them in the same place as now but use typenames and function names like
module_type and module_function and let gradualizer check if there is a
type or function named list_to_atom(lists:concat([Module,TypeOrFunc)) among
your overrides.

By using this approach you can still put the overrides in a normal module
and have it parsed and syntax checked.

/Kenneth

zuiderkwast:
> I suggest that you override the otp type specs by using a name convention. Put them in the same place as now but use typenames and function names like module_type and module_function

That's a good idea @KennethL! I guess there are no underscores in the module names we're overriding, so it should be unambiguous. Otherwise, we can use two underscores instead.

KennethL:
There are underscores in some modules so it is better to select a more
unique separator.

But I guess you are only creating the name and then checking if that type
exists so there might not be any problems anyway. You are not splitting
typenames into Module and Type? In that case it is better with double
underscore or some other unique separator.

/Kenneth

Zalastax:
[...]

PR #193, opened 2019-09-25
Allow overriding types for other modules in prelude

zuiderkwast:
With this change, we're also not using the remote spec syntax
which is not supported in OTP anymore.


zuiderkwast:
This is the follow-up to #191, as suggested by @gomoripeti in https://github.com/josefs/Gradualizer/issues/187#issuecomment-534583656.

I used the syntax `-override_module(Module)`, since it's not only specs anymore but also types and records. One known problem was fixed.

src/gradualizer_db.erl
```diff
@@ -253,9 +253,18 @@ code_change(_OldVsn, State, _Extra) ->
...
+    FormsByModule = partition_forms_by_override_module(gradualizer_prelude, Forms),
+    %% Import forms each of the modules to override
+    State1 = lists:foldl(fun ({Module, FormsThisModule}, StateAcc) ->
+                                 import_absform(Module, FormsThisModule, StateAcc)
+                         end,
+                         State,
+                         FormsByModule),
+    %% Mark the just overridden modules as not yet loaded, to make sure they
+    %% are loaded on demand
+    State1#state{loaded = Loaded#{gradualizer_prelude => true}}.
```
gomoripeti: hah, nice

zuiderkwast:
That's a good point @josefs.

The `'module:name'` syntax would be a bit strange for overriding internal types though, e.g. `deep_list` below, a type which is not exported from lists.

```Erlang
-type 'lists:deep_list'(A) :: [A | deep_list(A)].
-spec 'lists:flatten'(deep_list(A))      -> [A].
```

Here, because of the `'lists:deep_list'` representation of the name, these definitions would be treated as part of the lists module, so in the RHS, simply `deep_lists` is used. Also in the spec below, `deep_lists` is used without module prefix. This looks a bit strange. The alternative would be to write `'lists:deep_list'` everywhere...

```Erlang
-type 'lists:deep_list'(A) :: [A | 'lists:deep_list'(A)].
-spec 'lists:flatten'('lists:deep_list'(A))      -> [A].
```

...but then we'd obviously have to handle this syntax in a lot more places.

The beauty of the `-override_type(lists)` syntax IMO is that it's clear that the code below this attribute is simply taken as part of the lists module, unchanged. (It was also easy to implement.)

----

An alternative would be to have one file per module we want to override, e.g. `priv/overrides/lists.erl`. I would put them under `priv` so that rebar doesn't think the're erl files to be compiled. Instead we can read these files in the prelude parse transform and let them be returned by the prelude module.

josefs:
I like the idea of having one file per overridden module. We've already talked about splitting up this module anyway so that seems like a really nice solution actually!

gomoripeti:
One more thing we could consider when defining the format of the override file is that in the future users could also define their own override files (eg for 3rd party libs). In that case I think its more convenient to allow a single file to override let's say 3 specs each from 3 different libs. Similar to a single dialyzer.ignore file. In that example it would be a bit less convenient to create 3 files for 3 modules each with only 1 function.

But I also agree the prelude file grew too large so it makes sense to split it up and support multiple override files.

josefs:
@gomoripeti yes, if we allow users to override libraries in the future it'd be somewhat less convenient to have one file per overridden module. Are you arguing that this inconvenience is a killer and that we should allow multiple overridden modules per file? 
For me it doesn't weigh particularly heavy and I think the problems that are solved by having multiple files clearly wins out in this case.

Zalastax:
Making a feature easy to explain is a high priority for me so I want us to go for split files. Users won't complain over having to split their preludes and split files is more intuitive.

zuiderkwast:
We can allow users so specify their own *prelude directories*,

```sh
$ gradualizer --prelude-dir ~/my/prelude -pa xyz -- files_to_check`
```

Such dir can have the same structure, i.e. one file per module to override stuff for (e.g. lists.erl). Many small files? Who cares? It's a simple interface. Maybe we should consider a different suffix than `.erl` though, as these files won't compile as Erlang modules. How about `.erlspec` or similar?

josefs:
Having a separate file suffix sounds like a good idea. I don't have a strong opinion on the actual name

src/gradualizer_prelude_parse_trans.erl
```diff
@@ -3,13 +3,27 @@
...
+    [Form | replace_get_modules_and_forms(RestForms)].
+
+%% The value to be returned by the function in the parse transformed module
+-spec get_module_forms_tuples() -> [{module(), gradualizer_file_utils:abstract_forms()}].
+get_module_forms_tuples() ->
+    Files = filelib:wildcard(filename:join([code:priv_dir(gradualizer), "prelude", "*.erl"])),
+    lists:map(fun get_module_and_forms/1, Files).
+
+%% Parses and returns the forms of a file along with the module given in the -module attribute
+-spec get_module_and_forms(file:filename()) -> gradualizer_file_utils:abstract_forms().
```
josefs: dialyzer seems to have a problem with the return type.

zuiderkwast: Yes, it was a bad spec (AKA type error).

priv/prelude/mnesial.specs.erl
```diff
@@ -0,0 +1,164 @@
+-module(mnesia).
```
KennethL: Why is the filename mnesial.specs.erl? Should it not be mnesia.specs.erl?

zuiderkwast: Oops, you're right of course! (The file name is not checked/used so it makes no functional difference but still.)

gomoripeti:
great work @zuiderkwast 

Im not sure if this is the case but  if `gradualizer_prelude.beam` is already compiled and one changes a specs file, rebar might not recompile the module as `gradualizer_prelude.erl` hasn't changed and rebar does not know there is a dependency between the specs file and the source file. If this is indeed a problem, maybe better to use a compile hook escript instead of the parse trans to conditionally regenerate the `gradualizer_prelude.erl` file. What do you all think?

Zalastax:
I don't trust the recompilation process anyway so I always delete the `_build` folder. I'm happy with the current solution.

zuiderkwast:
@gomoripeti 
> Im not sure if this is the case but if `gradualizer_prelude.beam` is already compiled and one changes a specs file, rebar might not recompile the module as `gradualizer_prelude.erl` hasn't changed and rebar does not know there is a dependency between the specs file and the source file.

Right.... In a Makefile (let's say erlang.mk) it would be so easy to just add a dependency, like `ebin/gradualizer_prelude.beam: priv/prelude` or so. (Why did people invent rebar anyway?)

> If this is indeed a problem, maybe better to use a compile hook escript instead of the parse trans to conditionally regenerate the `gradualizer_prelude.erl` file. What do you all think?

I guess it needs to be addressed whenever someone adds or changes a prelude spec the next time. Otherwise users won't see any effect. Personally, I don't want to have hard dependencies on rebar, so I prefer a parse transform, but as long as rebar is used to build, a hook for at least invalidating the prelude would be good. (IMO gradualizer should be compatible with rebar as a plugin, etc. but preferably it should also be possible to build and use it with e.g. Mix or GNU Make or even Erlang `make:all()` as an alternative. It shouldn't be that hard since we don't have any dependencies. For dev stuff like running the tests, requiring rebar is fine.)

@Zalastax 
> I don't trust the recompilation process anyway so I always delete the _build folder. I'm happy with the current solution.

Wow, it really is a ridiculous build system then. IMO you can't expect people to `make clean` at every update.

josefs:
I'm just going to add that I've lost a lot of time because the recompilation checker in rebar doesn't do the right thing. And I still have an issue when using dialyzer which I cannot seem to fix no matter how many files I delete. At this point I'm considering setting my laptop on fire to see if that helps.
I'm very tempted to replace rebar for the day-to-day development of gradualizer because it just slows me down right now. Though as @zuiderkwast says, we should still make sure we're compatible with rebar e.g. as a plugin. But we should have another build system which doesn't act as a roadblock.

Issue #196, opened 2019-10-04
Cannot compile Gradualizer


aleklisi:
I am using Erlang OTP 22.0.7
```
erl -version
Erlang (SMP,ASYNC_THREADS,HIPE) (BEAM) emulator version 10.4.4
```
```
erl
Erlang/OTP 22 [erts-10.4.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe] [dtrace]

Eshell V10.4.4  (abort with ^G)
1> erlang:system_info(otp_release).
"22"
```

I cloned your repo and tried `make escript` on the master branch. So I run:

```
make escript
rebar3 escriptize
===> Verifying dependencies...
===> Compiling gradualizer
===> Compiling src/gradualizer_prelude.erl failed
src/gradualizer_prelude.erl:none: error in parse transform 'gradualizer_prelude_parse_trans': {undef,
                                                             [{gradualizer_file_utils,
                                                               get_forms_from_erl,
                                                               ["/Users/aleksanderlisiecki/Documents/Github/Gradualizer/_build/default/lib/gradualizer/priv/prelude/application.specs.erl",
                                                                []],
                                                               []},
                                                              {gradualizer_prelude_parse_trans,
                                                               get_module_and_forms,
                                                               1,
                                                               [{file,
                                                                 "/Users/aleksanderlisiecki/Documents/Github/Gradualizer/src/gradualizer_prelude_parse_trans.erl"},
                                                                {line,26}]},
                                                              {lists,map,2,
                                                               [{file,
                                                                 "lists.erl"},
                                                                {line,1239}]},
                                                              {gradualizer_prelude_parse_trans,
                                                               replace_get_modules_and_forms,
                                                               1,
                                                               [{file,
                                                                 "/Users/aleksanderlisiecki/Documents/Github/Gradualizer/src/gradualizer_prelude_parse_trans.erl"},
                                                                {line,10}]},
                                                              {gradualizer_prelude_parse_trans,
                                                               replace_get_modules_and_forms,
                                                               1,
                                                               [{file,
                                                                 "/Users/aleksanderlisiecki/Documents/Github/Gradualizer/src/gradualizer_prelude_parse_trans.erl"},
                                                                {line,15}]},
                                                              {compile,
                                                               '-foldl_transform/3-anonymous-2-',
                                                               3,
                                                               [{file,
                                                                 "compile.erl"},
                                                                {line,1147}]},
                                                              {compile,
[...]

PR #197, opened 2019-10-04
Force compilation order

aleklisi:
This PR solves compilation order issue, reported by me in 
 https://github.com/josefs/Gradualizer/issues/196


zuiderkwast:
Not needed. The commit 97e8a9363fb21538c4a7223278ab59499ca2be19 removes the dependency to gradualizer_file_utils in the parse transform. Now, it's enough that the parse transform is compiled first, which rebar does automatically.

Issue #199, opened 2019-10-11
Binary comprehensions don't match subtyping


NelsonVides:
```erlang
-spec iodata_binary() -> iodata().
iodata_binary() ->
    <<<<"A">> || _ <- lists:seq(1, 10)>>.
```

This is a testcase extracted from some actual code of mine where I would expect Gradualizer not to give me the error:
> `The binary comprehension on line 7 at column 5 is expected to have type iodata() which has no binary subtypes`

Which is pretty much senseless, `binary()` _is_ a subtype of `iodata()`, for example here: https://github.com/josefs/Gradualizer/blob/beac8edf9bfb1d5fe77d4068126c55249a6a6ab7/src/typechecker.erl#L663-L664

Trying to have a look at where Gradualizer makes this mistake, I arrived at this line: https://github.com/josefs/Gradualizer/blob/beac8edf9bfb1d5fe77d4068126c55249a6a6ab7/src/typechecker.erl#L2771-L2772

This function is called with parameters `ResTy`: `{type,{5,26},union,[{type,{5,26},iolist,[]},{type,{5,26},binary,[]}]}`, which doesn't match any of the case statements.

I'm trying to have a look if I can fix it and make a PR myself, (from this branch https://github.com/NelsonVides/Gradualizer/tree/binary_comprehension), so if you have any suggestions for me I'd very much appreciate it 🙂

PR #201, opened 2019-10-11
Binary comprehension

NelsonVides:
Trying to solve https://github.com/josefs/Gradualizer/issues/199 by checking that `binary` is a subtype of the expected result type of the binary comprehension.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1418,7 +1418,10 @@ expect_binary_type({var, _, Var}, _) ->
     {elem_ty
     ,{var, erl_anno:new(0), TyVar}
     ,constraints:add_var(TyVar,
-      constraints:upper(Var, {type, erl_anno:new(0), binary, [TyVar]}))
+      constraints:upper(Var, {type, erl_anno:new(0), binary,
+                              [{integer, erl_anno:new(0), 0},
+                               {integer, erl_anno:new(0), 1}]}
+                       ))
     };
 expect_binary_type(Ty, _) ->
     {type_error, Ty}.
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1393,6 +1393,82 @@ expect_tuple_union([], AccTy, AccCs, any, N, _Env) ->
 expect_tuple_union([], AccTy, AccCs, _NoAny, _N, _Env) ->
     {AccTy, AccCs}.
 
+-spec purge_binary(type()) -> type().
+purge_binary(ResTy) ->
+    case ResTy of
+        {type, _, binary, [{integer, _, 0}, {integer, _, _N}]} ->
+            %% The result is a multiple of N bits.
+            %% Expr must be a multiple of N bits too.
+            ResTy;
+        {type, _, binary, [{integer, _, M}, {integer, _, _N}]}
+          when M > 0 ->
+            %% The result is a binary with a minimum size of M. This
+            %% requires that the generators are non-empty. We don't
+            %% check that. At least, we can check that Gen is a
+            %% bitstring.
+            %% TODO: properly check these values
+            {type, erl_anno:new(0), binary,
+             [{integer, erl_anno:new(0), 0},
+              {integer, erl_anno:new(0), 1}]}
+    end.
+
+-spec expect_binary_type(type(), env()) -> R when
+      R :: any
+         | {elem_ty, type(), constraints()}
+         | {elem_tys, [type()], constraints()}
+         | {type_error, type()}.
+expect_binary_type(?type(any), _) ->
+    any;
+expect_binary_type(ElemTy = {type, _, binary, _}, _) ->
+    {elem_ty, ElemTy, constraints:empty()};
+expect_binary_type(Union = {type, _, union, UnionTys}, Env) ->
+    {Tys, Cs} = expect_binary_union(UnionTys, [], constraints:empty(), Env),
+    case Tys of
+        [] ->
+            {type_error, Union};
+        [Ty] ->
+            {elem_ty, Ty, Cs};
+        _ ->
+            {elem_tys, Tys, Cs}
+    end;
+expect_binary_type({var, _, Var}, _) ->
+    TyVar = gradualizer_tyvar:new(Var, ?MODULE, ?LINE),
+    {elem_ty
+    ,{var, erl_anno:new(0), TyVar}
+    ,constraints:add_var(TyVar,
+      constraints:upper(Var, {type, erl_anno:new(0), binary, [TyVar]}))
+    };
+expect_binary_type(Ty, _) ->
+    {type_error, Ty}.
+
+-spec expect_binary_union([type()], _, _, env()) -> R when
+      R :: any
+         | {elem_ty, type(), constraints()}
+         | {elem_tys, [type()], constraints()}
+         | {type_error, type()}.
+expect_binary_union([Ty|Tys], AccTy, AccCs, Env) ->
+    case expect_binary_type(normalize(Ty, Env), Env) of
+        {type_error, _} ->
+            expect_binary_union(Tys, AccTy, AccCs, Env);
+        any ->
+            expect_binary_union(Tys
+                                ,[type(any) | AccTy]
+                                ,AccCs
+                                ,Env);
+        {elem_ty, NTy, Cs} ->
+            expect_binary_union(Tys
+                                ,[NTy | AccTy]
+                                ,constraints:combine(Cs, AccCs)
+                                ,Env);
+        {elem_tys, NTys, Cs} ->
+            expect_binary_union(Tys
+                                ,NTys ++ AccTy
+                                ,constraints:combine(Cs, AccCs)
+                                ,Env)
+    end;
+expect_binary_union([], AccTy, AccCs, _Env) ->
+    {AccTy, AccCs}.
+
 
 -spec allow_empty_list(type()) -> type().
 allow_empty_list({type, P, nonempty_list, []}) ->
@@ -3334,28 +3410,20 @@ type_check_comprehension_in(Env, ResTy, OrigExpr, lc, Expr, _P, []) ->
             throw(type_error(OrigExpr, type(list), ResTy))
     end;
 type_check_comprehension_in(Env, ResTy, OrigExpr, bc, Expr, _P, []) ->
-    ExprTy = case ResTy of
-                 {type, _, binary, [{integer, _, 0}, {integer, _, _N}]} ->
-                     %% The result is a multiple of N bits.
-                     %% Expr must be a multiple of N bits too.
-                     ResTy;
-                 {type, _, binary, [{integer, _, M}, {integer, _, _N}]}
-                   when M > 0 ->
-                     %% The result is a binary with a minimum size of M. This
-                     %% requires that the generators are non-empty. We don't
-                     %% check that. At least, we can check that Gen is a
-                     %% bitstring.
-                     {type, erl_anno:new(0), binary,
-                            [{integer, erl_anno:new(0), 0},
-                             {integer, erl_anno:new(0), 1}]};
-                 _ ->
-                     Ty = {type, erl_anno:new(0), binary,
-                            [{integer, erl_anno:new(0), 0},
-                             {integer, erl_anno:new(0), 1}]},
-                     throw(type_error(OrigExpr, Ty, ResTy))
-             end,
-    {_VB, Cs} = type_check_expr_in(Env, ExprTy, Expr),
-    {Env, Cs};
+    case expect_binary_type(ResTy, Env) of
+        any ->
+            {_Ty, _VB, Cs} = type_check_expr(Env, Expr),
+            {#{}, Cs};
+        {elem_ty, ElemTy, Cs1} ->
+            %% TODO: Don't purge that binary like that
+            {_VB, Cs2} = type_check_expr_in(Env, purge_binary(ElemTy), Expr),
+            {#{}, constraints:combine(Cs1, Cs2)};
+        {elem_tys, ElemTys, Cs1} ->
+            {VB, Cs2} = type_check_union_in(Env, ElemTys, Expr),
+            {VB, constraints:combine(Cs1, Cs2)};
+        {type_error, Ty} ->
+            throw({type_error, OrigExpr, Ty, ResTy})
+    end;
 type_check_comprehension_in(Env, ResTy, OrigExpr, Compr, Expr, P,
                             [{generate, _, Pat, Gen} | Quals]) ->
     {Ty, _VB1, Cs1} = type_check_expr(Env, Gen),
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1393,24 +1393,6 @@ expect_tuple_union([], AccTy, AccCs, any, N, _Env) ->
 expect_tuple_union([], AccTy, AccCs, _NoAny, _N, _Env) ->
     {AccTy, AccCs}.
 
--spec purge_binary(type()) -> type().
-purge_binary(ResTy) ->
-    case ResTy of
-        {type, _, binary, [{integer, _, 0}, {integer, _, _N}]} ->
-            %% The result is a multiple of N bits.
-            %% Expr must be a multiple of N bits too.
-            ResTy;
-        {type, _, binary, [{integer, _, M}, {integer, _, _N}]}
-          when M > 0 ->
-            %% The result is a binary with a minimum size of M. This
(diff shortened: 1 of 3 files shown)


src/typechecker.erl
```diff
@@ -2783,7 +2783,11 @@ type_check_comprehension_in(Env, ResTy, bc, Expr, P, []) ->
                             [{integer, erl_anno:new(0), 0},
                              {integer, erl_anno:new(0), 1}]};
                  _ ->
-                     throw({type_error, bc, P, ResTy})
+                     case subtype({type, erl_anno:new(0), binary, []}, ResTy, Env) of
```
zuiderkwast: This covers ResTy `iodata()`, but what if ResTy is a union containing `bitstring()` or another bit type such as `<<_:7, _:_*3>>`? I think the general solution would be do like in the list comprehension case and create a function `expect_bitstring_type`.

Another solution: I think we can get rid of all the `expect_` functions though if we split any union (e.g. `iodata()` after normalizing it) before this function is called. I'm not sure if there is a performance concern for doing this, @josefs? Personally I'm more concerned about the code complexity than the performance for these things.

test/should_pass/binary_in_union.erl
```diff
@@ -0,0 +1,6 @@
+-module(binary_in_union).
+-export([iodata_binary/0]).
+
+-spec iodata_binary() -> iodata().
+iodata_binary() ->
+    <<<<"A">> || _ <- lists:seq(1, 10)>>.
```
zuiderkwast: Good test case.

src/typechecker.erl
NelsonVides: Ok, thinking about that, I came up with this other testcase, which fails for my current solution:
```erlang
-spec bitstring_fun() -> <<_:7, _:_*3>> | string().
bitstring_fun() -> << <<1:N>> || N <- lists:seq(3, 12)>>.
```

I think I'm coming for a solution for that, with the idea of splitting the unions, but the code is getting a bit uglier at least from my perspective. I'll push it here when I have it, but maybe I could put it on a different PR so that you can choose the best solution, maybe?

NelsonVides: An even crazier example
```erlang
-type nested() :: string() | bitstring() | binary().
-spec nested_bitstrings() -> nested() | boolean() | pid().
nested_bitstrings() ->
    << <<1:N>> || N <- lists:seq(3, 12)>>.
```
:smiley:

test/should_pass/binary_in_union.erl
```diff
@@ -0,0 +1,19 @@
...
+         nested_bitstrings/0
+        ]).
+
+-spec iodata_binary() -> iodata().
+iodata_binary() ->
+    <<<<"A">> || _ <- lists:seq(1, 10)>>.
+
+-spec bitstring_fun() -> <<_:7, _:_*3>> | string().
+bitstring_fun() ->
+    << <<1:N>> || N <- lists:seq(3, 12)>>.
```
zuiderkwast: This shouldn't type check. The value is not of the expected type.

NelsonVides: Hmmm... I think I just saw why it shouldn't: https://github.com/josefs/Gradualizer/blob/c31c2326ad6b9674cefdb5c202f036bd86a150ce/src/typechecker.erl#L2776 There, it's pretty much accepting any sizes of M and N and making them simply 0 and 1, which will be compatible with everything else :slightly_frowning_face:

src/typechecker.erl
```diff
@@ -2302,7 +2302,17 @@ do_type_check_expr_in(Env, ResTy, {call, P, Name, Args}) ->
...
-    type_check_comprehension_in(Env, ResTy, bc, Expr, P, Qualifiers);
+    Error = {type_error, bc, P, ResTy},
+    {Res, Acc} = lists:mapfoldl(
+        fun(FlatTy, Acc) ->
+            case catch {type_check_comprehension_in(Env, FlatTy, bc, Expr, P, Qualifiers), Acc} of
+                {{#{}, C1}, {#{}, C2}}                 -> {ok, {#{}, constraints:combine(C1, C2)}};
+                {{#{}, _} = C, {type_error, _, _, _}} -> {ok, C};
+                {type_error, _, _, FlatTy}            -> {type_error,Acc}
+            end
+        end, Error, flatten_type(ResTy, Env)),
```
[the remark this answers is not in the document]
NelsonVides: I'm catching exactly on `{type_error...}` because that is the only thing that is thrown on the callees while exploring the code. Also, if there's any other error, this is not where I want to catch it. But I've realised I might be masking different `{type_error,...` messages being thrown and making them all be only the main one... :thinking: 

About `flatten_type/2`, it is used in normalised, but then normalised also does a `case merge_union_types/2 of` which, when matching on a list longer than one, returns again a type-union of them (having removed possible duplicates and so on, but in this case it means nothing changed. That's why I directly used `flatten_type/2`, I want the list to iterate through.

Also, where do you mean we can move the logic? To `do_type_check_expr_in` caller, that is, the ubiquitous `type_check_expr_in`, right? Hmm... this is getting a lot more complex than I expected as an entry point for me, but well, sounds cool :smile:

test/should_pass/binary_in_union.erl
zuiderkwast: Right... Only the word TODO is missing in the comment.

Your test case can be put under `test/known_problems/should_fail/`.

If the fixed bit size is zero though (e.g. a type like `<<_:_*3>>`) we are able to handle it.

NelsonVides:
@zuiderkwast I moved that test to `test/known_problems/should_fail/`, I tried having a look at fixing it, but I honestly didn't know how for now.

I also changed the matching on the throws, let me know what do you think about it by now and what else can I do to improve this PR.

src/typechecker.erl
```diff
@@ -2302,7 +2302,19 @@ do_type_check_expr_in(Env, ResTy, {call, P, Name, Args}) ->
...
+                {{#{}, C1}, {#{}, C2}}                -> {ok, {#{}, constraints:combine(C1, C2)}};
+                {{#{}, _} = C, {type_error, _, _, _}} -> {ok, C};
+                Err when element(1, Err)=:=type_error -> {Err,Acc}
+            end
+        end, {type_error, bc, P, ResTy}, flatten_type(ResTy, Env)),
+    case {lists:any(fun(El) -> El =:= ok end, Res), proplists:lookup(type_error,Res)} of
+        {true, _} -> Acc;
+        {_, none} -> Acc;
+        {_,Err} -> throw(Err)
+    end;
```
zuiderkwast: The tuple `{type_error, _, _, _}` is not enough for a mismatch. For example, if a type such as `binary() | nonexisting_module:foo()` is expected (which easily happens if the code path to deps isn't available), you will get a crash here. In general, all exceptions of class `throw` can happen, so use try-catch and some more exact logic. (Take a look at `handle_type_error` and all the throws in this module to see what it can be. More can also be added in the future.)

Btw, this is way different to how the same thing is done for list comprehensions. Take a look at `expect_list_type/1`. It isn't even using exceptions, but filters the type union on a different level. I'm not saying it has to be done the same way for bc, but consider the differences. (Personally, I'd like to replace these with loop over the union on a higher level, in a clause for `do_type_check_expr_in(Env, {type, _, union, Tys}, Expr)`, but that's a different PR and discussion.)

Don't use flatten_type/2. It's an unintuitive function because it turns a type into a list of types, which isn't clear from its name. It's intended as a helper for flatten_unions/2, which is a helper for normalize/2, only. (Yes, they should be private in another module, I agree.) Pattern match on `{type, union, _, _}` instead.

I think this is somewhat hard to read as well. Consider breaking it up into multiple functions.

test/known_problems/should_fail/binary_comprehension.erl
```diff
@@ -0,0 +1,10 @@
+-module(binary_comprehension).
+-compile([debug_info]).
+-export([
+         bitstring_fun/0
+        ]).
+
+-spec bitstring_fun() -> <<_:7, _:_*3>> | string().
+bitstring_fun() ->
+    << <<1:N>> || N <- lists:seq(3, 12)>>.
+
```
zuiderkwast: "fun" means lambda expression in Erlang, so that part isn't very good in the function name.

test/should_pass/binary_in_union.erl
```diff
@@ -0,0 +1,15 @@
+-module(binary_in_union).
+-export([
+         iodata_binary/0,
+         bitstring_fun/0,
```
zuiderkwast: This one is moved to a different module.

src/typechecker.erl
NelsonVides: Ah, there's one way to do all the stuff I was doing on only three lines of codes, that `type_check_union_in` that was used for the list comprehension case. What do you think now?

josefs:
It's nice to see binary comprehensions getting some love because they're almost completely forgotten right now. In order to type check them properly we need a new function `expect_binary_type` similar to `expect_list_type`, `expect_tuple_type` etc. It's a great little project if you want to wade in a little deeper into the code base.

NelsonVides:
@josefs 

> It's nice to see binary comprehensions getting some love because they're almost completely forgotten right now. In order to type check them properly we need a new function `expect_binary_type` similar to `expect_list_type`, `expect_tuple_type` etc. It's a great little project if you want to wade in a little deeper into the code base.

As we've been talking with @zuiderkwast, the fact that there are all these `expect_` functions seems a bit reduntant. I'm trying to implement it now that way and it feels like duplicated code everywhere, doing the union handling for each language construct. Unions should be folded earlier in the call stack, so these points don't need to handle all sorts of unions separately :disappointed:

josefs:
Yes, I agree that the is room for improvement in the code. Patches are most welcome. I'm afraid it's not very high on my priority list at the moment.

zuiderkwast:
> As we've been talking with @zuiderkwast, the fact that there are all these `expect_` functions seems a bit reduntant. I'm trying to implement it now that way and it feels like duplicated code everywhere, doing the union handling for each language construct. Unions should be folded earlier in the call stack, so these points don't need to handle all sorts of unions separately disappointed

I agree with you @NelsonVides! I didn't mean to tell you to rewrite the code into the style of the old stuff. I just wanted you to be aware of it, so that we can refactor it as whole. Folding higher up on the call stack is exactly what I've been tying to suggest, but then we need to take the error messages into account, perhaps not simply re-throwing the one for one element in the union.

src/typechecker.erl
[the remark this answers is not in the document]
zuiderkwast: > I'm catching exactly on {type_error...} because that is the only thing that is thrown on the callees while exploring the code. Also, if there's any other error, this is not where I want to catch it.

That's not true. You're catching all kinds of exceptions and then crashing on a case clause.

This is what's happening in a catch without try:

```Erlang
1> Thrower = fun() -> case catch throw(boomerang) of [] -> [] end end.
#Fun<erl_eval.20.128620087>
2> try Thrower() catch C:T:E -> io:format("~p:~p~n at ~p~n", [C, T, E]) end.
error:{case_clause,boomerang}
 at [{erl_eval,case_clauses,6,[{file,"erl_eval.erl"},{line,968}]},
     {erl_eval,try_clauses,8,[{file,"erl_eval.erl"},{line,914}]},
     {shell,exprs,7,[{file,"shell.erl"},{line,686}]},
     {shell,eval_exprs,7,[{file,"shell.erl"},{line,642}]},
     {shell,eval_loop,3,[{file,"shell.erl"},{line,627}]}]
ok
```

To catch selectively, you must use try-catch, not only catch.

> That's why I directly used flatten_type/2, I want the list to iterate through.

I understand it does exactly what you need here, but the name is not very comprehensive and there is no documentation for it. Please rename it! If it has a comprehensive name, it can be used and I can remove it from my list of functions to move and make unexported in another modules. Please rename. (I'm just trying to line out some sort of modularization of the giant module.)

> Also, where do you mean we can move the logic?

`type_check_expr_in` is normalizing the type and then calls its helper `do_type_check_expr_in`. Put the function clause `do_type_check_expr_in(Env, {type, _, union, Tys}, Expr)` before any function clauses dissecting Expr. Then it can safely, recursively call itself, because there is no more union on the same level.

One more thing: Please, don't do as I suggest if you don't agree! It's just comments. :-)

NelsonVides:
@zuiderkwast It's all rewritten now, following the style of `expect_` functions. I was aware of those from the very beginning, but I had the feeling they were a bit of an overengineering for something so simple. In any case I also accept the idea of keeping things consistent, even if the style chosen is not the best, it's better to have just _one_ style rather than a bunch of incoherent mixes. We can fix that when we move the union folding higher up the call stack :)

src/typechecker.erl
```diff
@@ -1119,6 +1119,71 @@ expect_tuple_union([], AccTy, AccCs, any, N) ->
...
+expect_binary_union([Ty|Tys], AccTy, AccCs, TEnv) ->
+    case expect_binary_type(normalize(Ty, TEnv), TEnv) of
+        {type_error, _} ->
+            expect_binary_union(Tys, AccTy, AccCs, TEnv);
+        any ->
+            expect_binary_union(Tys
+                                ,[type(any) | AccTy]
+                                ,AccCs
+                                ,TEnv);
+        {elem_ty, NTy, Cs} ->
```
josefs: My memory is rather foggy about how this works but from what I recall, handling `any()` in this way doesn't actually work. Look at how it is implemented for lists and tuples. There we always return just `any()` as the element type if in one of the unions there is a case that the element type is `any()`.

zuiderkwast: @josefs I don't know about the case for lists, but we have working examples of handling unions with any: https://github.com/josefs/Gradualizer/blob/master/test/should_fail/union_with_any.erl

src/typechecker.erl
```diff
@@ -2774,28 +2839,20 @@ type_check_comprehension_in(Env, ResTy, OrigExpr, lc, Expr, _P, []) ->
...
-             end,
-    {_VB, Cs} = type_check_expr_in(Env, ExprTy, Expr),
-    {#{}, Cs};
+    case expect_binary_type(ResTy, Env#env.tenv) of
+        any ->
+            {_Ty, _VB, Cs} = type_check_expr(Env, Expr),
+            {#{}, Cs};
+        {elem_ty, ElemTy, Cs1} ->
+            %% TODO: Don't purge that binary like that
+            {_VB, Cs2} = type_check_expr_in(Env, purge_binary(ElemTy), Expr),
```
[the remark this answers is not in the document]
NelsonVides: This quickly named purge function is something that shouldn't be happening, it's precisely that place were we turn any binary with prefixes. I'm putting that back to where it was, inside of the main function, it's just that this gets a bit longer. Adding the TODO we talked about as well.

src/typechecker.erl
```diff
@@ -1119,6 +1119,71 @@ expect_tuple_union([], AccTy, AccCs, any, N) ->
...
+                                ,AccCs
+                                ,TEnv);
+        {elem_ty, NTy, Cs} ->
+            expect_binary_union(Tys
+                                ,[NTy | AccTy]
+                                ,constraints:combine(Cs, AccCs)
+                                ,TEnv);
+        {elem_tys, NTys, Cs} ->
+            expect_binary_union(Tys
+                                ,NTys ++ AccTy
```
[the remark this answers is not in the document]
NelsonVides: But the problem is that `expect_binary_type` is called with a type that passed a call to `normalize`, and can still have several binary types within the union, check the test `test/should_pass/binary_in_union:nested_bitstrings/0`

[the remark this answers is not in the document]
NelsonVides: I think that what you're missing is the `TODO` on `merge_union_types`, precisely.

src/typechecker.erl
NelsonVides: I think that in this time I did it pretty much as in the case for lists: https://github.com/josefs/Gradualizer/blob/0b4c6b0b78741bb77f1226db7f316bdf1b38787f/src/typechecker.erl#L1014 `any` doesn't make the rest of the list be ignored, right?

josefs: Yes, you're right. I read the code wrong.

Zalastax:
This pull request has unfortunately lost momentum. What are the current blocking items? How can we help get this merged?

erszcz:
I've rebased this on top of the current master and made sure all the tests pass - they do, together with the self-check. I'll merge it now, as otherwise it might linger forever.

test/should_pass/binary_in_union.erl
Mohammedfffllll: ghkfyfnvhoh@gmail.com

PR #204, opened 2019-10-16
Make spec overrides configurable

Zalastax:
Let's users use their own "preludes" without recompiling Gradualizer.
Could be improved to not re-import the preludes if multiple files are checked, but perhaps that can wait for another PR?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,6 +161,12 @@ cli-tests: bin/gradualizer
 	# 8. No location, no filename
 	bin/gradualizer --fmt-location none --no-print-file test/dir/test_in_dir.erl \
 	|perl -ne '/^The variable N is expected/ or die "CLI 8 ($$_)"'
+	# 9. Possible to exclude prelude (-0777 from https://stackoverflow.com/a/30594643/497116)
+	bin/gradualizer --no-prelude test/should_pass/cyclic_otp_specs.erl \
+	|perl -0777 -ne '/^The type spec/g or die "CLI 9 ($$_)"'
+	# 10. Excluding prelude and then including it is a no-op
+	bin/gradualizer --no-prelude --specs-override-dir priv/prelude \
+	  test/should_pass/cyclic_otp_specs.erl || (echo "CLI 10"; exit 1)
 
 .PHONY: cover
 cover: EUNIT_OPTS =
--- a/src/gradualizer_db.erl
+++ b/src/gradualizer_db.erl
@@ -16,8 +16,8 @@
          get_modules/0, get_types/1,
          save/1, load/1,
          import_module/1,
-         import_erl_files/1, import_beam_files/1,
-         import_app/1, import_otp/0]).
+         import_erl_files/1, import_beam_files/1, import_extra_specs/1,
+         import_app/1, import_otp/0, import_prelude/0]).
 
 %% Callbacks
 -behaviour(gen_server).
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -161,6 +161,14 @@ cli-tests: bin/gradualizer
 	# 8. No location, no filename
 	bin/gradualizer --fmt-location none --no-print-file test/dir/test_in_dir.erl \
 	|perl -ne '/^The variable N is expected/ or die "CLI 8 ($$_)"'
+	# 9. Possible to exclude prelude (-0777 fromhttps://stackoverflow.com/a/30594643/497116)
```
josefs: Space beetwen 'from' and the url.

Makefile
```diff
@@ -161,6 +161,14 @@ cli-tests: bin/gradualizer
...
+	bin/gradualizer --no-prelude test/should_pass/cyclic_otp_specs.erl \
+	|perl -0777 -ne '/^The type spec/g or die "CLI 9 ($$_)"'
+	# 10. Excluding prelude and then including it is a no-op
+	if ! bin/gradualizer --no-prelude --specs-override-dir priv/prelude \
+	  test/should_pass/cyclic_otp_specs.erl ; \
+	  then echo "CLI 10" ; exit 1;\
+	fi
```
josefs: A test for `--specs-override-dir` also?

Zalastax: > This test is sufficient to me since we know that --no-prelude works.

josefs: Duh! Yes, sorry for not reading the code properly.

src/gradualizer_db.erl
```diff
@@ -265,6 +278,32 @@ import_prelude(State = #state{loaded = Loaded}) ->
...
+    Files = filelib:wildcard(filename:join([Dir, "*.specs.erl"])),
+    lists:map(fun get_module_and_forms/1, Files).
+
+%% Parses and returns the forms of a file along with the module given in the -module attribute
+-spec get_module_and_forms(file:filename()) -> {module(), gradualizer_file_utils:abstract_forms()}.
+get_module_and_forms(File) ->
+    {ok, Forms} = gradualizer_file_utils:get_forms_from_erl(File, []),
+    [{attribute, _, file, _},
+     {attribute, _, module, Module} | _] = Forms,
+    {Module, Forms}.
```
zuiderkwast: Can these functions be exported in `gradualizer_prelude_parse_trans` and used here? To avoid duplicated code while not adding any dependency to the parse transform. They are copied from there, right?

Makefile
```diff
@@ -161,6 +161,14 @@ cli-tests: bin/gradualizer
...
 	bin/gradualizer --fmt-location none --no-print-file test/dir/test_in_dir.erl \
 	|perl -ne '/^The variable N is expected/ or die "CLI 8 ($$_)"'
+	# 9. Possible to exclude prelude (-0777 from https://stackoverflow.com/a/30594643/497116)
+	bin/gradualizer --no-prelude test/should_pass/cyclic_otp_specs.erl \
+	|perl -0777 -ne '/^The type spec/g or die "CLI 9 ($$_)"'
+	# 10. Excluding prelude and then including it is a no-op
+	if ! bin/gradualizer --no-prelude --specs-override-dir priv/prelude \
+	  test/should_pass/cyclic_otp_specs.erl ; \
+	  then echo "CLI 10" ; exit 1;\
+	fi
```
zuiderkwast: Isn't this the same?

```sh
bin/gradualizer --no-prelude --specs-override-dir priv/prelude \
	  test/should_pass/cyclic_otp_specs.erl || (echo "CLI 10"; exit 1) \
```

PR #205, opened 2019-10-16
File consult prelude

Zalastax:
I'm not sure if this prelude change is the right thing. I think that most times when a user calls `file:consult/1`, they know what type the result will have. However, the file *could* contain something else, so perhaps `term()` is actually correct. Forcing the user to type out their expectations could be beneficial as well. Is `file:consult/1` actually one of the few functions where it's right that they return `term()`?

### DIFF
--- a/priv/prelude/file.specs.erl
+++ b/priv/prelude/file.specs.erl
@@ -1,0 +1,31 @@
+-module(file).
+
+%% This module contains specs to replace incorrect or inexact specs in OTP.
+
+-type deep_list() :: [char() | atom() | deep_list()].
+-type name_all()  :: string() | atom() | deep_list() | (RawFilename :: binary()).
+-type posix() ::
+        'eacces' | 'eagain' |
+        'ebadf' | 'ebadmsg' | 'ebusy' |
+        'edeadlk' | 'edeadlock' | 'edquot' |
+        'eexist' |
+        'efault' | 'efbig' | 'eftype' |
+        'eintr' | 'einval' | 'eio' | 'eisdir' |
+        'eloop' |
+        'emfile' | 'emlink' | 'emultihop' |
+        'enametoolong' | 'enfile' |
+        'enobufs' | 'enodev' | 'enolck' | 'enolink' | 'enoent' |
+        'enomem' | 'enospc' | 'enosr' | 'enostr' | 'enosys' |
+        'enotblk' | 'enotdir' | 'enotsup' | 'enxio' |
+        'eopnotsupp' | 'eoverflow' |
+        'eperm' | 'epipe' |
+        'erange' | 'erofs' |
+        'espipe'  | 'esrch'  | 'estale' |
+        'etxtbsy' |
+        'exdev'.
+
+-spec consult(Filename) -> {ok, Terms} | {error, Reason} when
+      Filename :: name_all(),
+      Terms :: [any()],
+      Reason :: posix() | badarg | terminated | system_limit
+              | {Line :: integer(), Mod :: module(), Term :: term()}.
--- a/test/should_pass/file_consult.erl
+++ b/test/should_pass/file_consult.erl
@@ -1,0 +1,8 @@
+-module(file_consult).
+
+-export([f/0]).
+
+-spec f() -> [integer()].
+f() ->
+    {ok, Terms} = file:consult("primes.txt"),
+    Terms.


priv/prelude/file.specs.erl
```diff
@@ -0,0 +1,31 @@
+-module(file).
+
+%% This module contains specs to replace incorrect or inexact specs in OTP.
+
+-type deep_list() :: [char() | atom() | deep_list()].
+-type name_all()  :: string() | atom() | deep_list() | (RawFilename :: binary()).
+-type posix() ::
+        'eacces' | 'eagain' |
```
zuiderkwast: Is it necessary to override posix()? I think it is defined in this module in all supported OTP versions. Then it doesn't need to be defined. It can be used anyway.

Zalastax:
When does `term()` make sense? My feeling is that we would be better of treating `term()` and `any()` as the same thing and inventing a new type to replace `term()` in those special cases it's actually useful.

josefs:
It makes sense theoretically as the top of the subtyping hierarchy. But I can't think of a good use for it in practical applications. Perhaps you're right, it'd be better to let `any()` and `term()` be the same thing and just have an internal representation for the top of the subtyping hierarchy. Or, as you suggest, have a completely new type which programmers can use to denote this.

The upshot of unifying `term()` and `any()` would be that we can use the OTP specs as is, and that would be a huge simplification!

I'm trying to think about downsides to doing this change but I can't really find any compelling reasons not to do it. The change would make gradualizer a bit more lenient because some programs involving `term()` would pass the typechecker instead of getting a type error. But those type errors would probably have been an annoyance anyway.

Am I missing some major downside?

josefs:
I don't think having a flag for treating `term()` as `any()` is a good idea. It would be extremely confusing when mixing modules which does have the flag with those that doesn't. If we're making this change we should do it unconditionally.

gomoripeti:
this is only about when term() is the return type or also as input?
- I can image `tem()` making sense as input (like `term_to_binary` or `appliscation:set_env` should really handle all types)
- but when calling a function that returns term is problematic, because the caller usually has an extra knowledge about the stricter type of what was stored in that file or application env var

otoh treating `term()` as `any()` only as return type might be just as confusing as having a flag per module

zuiderkwast:
1. Regarding the flag, the way I meant it was simply as a way to replace the overrides and that it should only apply to remote specs and types, as a way of keeping the behaviour the same as with the overrides but with less code. The conversion would be done in gradualizer_db or just after getting the remote types from it. Way less radical than term() = any() IMO.

2. Regarding abandoning the distinction between term and any altogether, maybe it's a good thing to do, although it seems radical. I remember some people really liked the distinction (Kostis?).

    If we do it, anyone who wants the top of the typesystem has to write `number() | atom() | tuple() | list() | map() | function() | binary() | pid() | reference() | port()` instead. (Did I forget something?) Nobody ever wants this (apart from anyone writing a fully typed implementation of term_to_binary/1 or the distribution protocol or fully typed usage of e.g. file:consult/1...) I find it hard to see how we can introduce a new type to use instead of term and get it accepted by OTP and the general public. Technically though, it would be easy to define a type alias for this union of everything and export it as `gradualizer:everything()` or something like that.

3. If we have a global flag for --term-as-any, it can be used a little bit like --infer, for exploring the different approaches.

A different perspective: Imagine we can give a measure in percentage of how much statically typed a program is (a "graduality measure"). Then projects will aim at having 100% statically typed code, to maximize the benefit of graduality. Then, there is suddenly a point in getting rid of all the `any()`s in a program. Replacing them with term() whenever possible is a way to do that, but only if term() is considered a static type...

Zalastax:
> Replacing them with term() whenever possible is a way to do that, but only if term() is considered a static type...

They could reach 100% staticness using `gradualizer:everything()`?

@gomoripeti, I suggest everywhere, not just for return types.

zuiderkwast:
> They could reach 100% staticness using gradualizer:everything()?

Yes, but that this slightly strange convention gets wide acceptance (unlikely?) and it requires having gradualizer as a dependency to the program. I would rather write out the full union if I were to write such a program. (An gradualizer option `--term-as-everything` could be nice for that case, even if we make term() = any() by default. Just kidding.)

I don't think this is an issue. I vote for term() = any(), globally and unconditionally. 🏆

josefs:
Cool! It seems that we've reached an agreement on treating `term()` as the dynamic type as well.

We still need an internal representation for the top of the subtyping hierarchy, for example in `glb`. But that's an implementation detail for whoever implements this change.

PR #206, opened 2019-10-16
Refining types on fun guards

NelsonVides:
I think that my intentions are better explained in the tests:

Now these are passing:
```erlang
-spec guard_is_atom(atom() | list()) -> list().
guard_is_atom(Name) when is_atom(Name) ->
    erlang:atom_to_list(Name);
guard_is_atom(_Name) ->
    [].

-spec remote_guard(atom() | list()) -> list().
remote_guard(Name) when erlang:is_atom(Name) ->
    erlang:atom_to_list(Name);
remote_guard(_Name) ->
    [].

-spec fun_correct_arity(any()) -> boolean().
fun_correct_arity(Fun) when is_function(Fun, 2) ->
    Fun(0, 1);
fun_correct_arity(_Fun) -> true.

-spec fun_unspecified_arity(any()) -> boolean().
fun_unspecified_arity(Fun) when is_function(Fun) ->
    Fun(0,1);
fun_unspecified_arity(_Fun) -> true.
```

And these are not:
```erlang
-spec fun_wrong_arity(any()) -> boolean().
fun_wrong_arity(Fun) when is_function(Fun, 2) ->
    Fun(0);
fun_wrong_arity(_Fun) -> true.

-spec guard_unreachable_branch(atom() | list()) -> list().
guard_unreachable_branch(Name)
  when is_atom(Name), is_list(Name) ->
    erlang:atom_to_list(Name);
guard_unreachable_branch(_Name) ->
    [].
```

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3227,14 +3227,14 @@ check_clause(_Env, [?type(none)|_], _ResTy, {clause, P, _Args, _Guards, _Block})
 check_clause(_Env, ArgsTy, _ResTy, {clause, P, Args, _Guards, _Block})
   when length(ArgsTy) =/= length(Args) ->
     throw({argument_length_mismatch, P, length(ArgsTy), length(Args)});
-check_clause(Env, ArgsTy, ResTy, C = {clause, _, Args, Guards, Block}) ->
+check_clause(Env, ArgsTy, ResTy, C = {clause, P, Args, Guards, Block}) ->
     ?verbose(Env, "~sChecking clause :: ~s~n", [format_location(C, brief), typelib:pp_type(ResTy)]),
     {PatTys, _UBounds, VEnv2, Cs1} = add_types_pats(Args, ArgsTy, Env#env.tenv, Env#env.venv),
     EnvNew      = Env#env{ venv =  VEnv2 },
     %%%
     %% Here I want to simply check if the block inside the clause typechecks,
     %% taking the refined guards into consideration
-    VarBinds1   = check_guards(EnvNew, Guards),
+    VarBinds1   = check_guards(EnvNew, Guards, P),
     EnvNewest   = EnvNew#env{ venv = refine_var_binds(EnvNew#env.venv, VarBinds1, Env#env.tenv) },
     {VarBinds2, Cs2} = type_check_block_in(EnvNewest, ResTy, Block),
     %%%
@@ -3410,15 +3410,26 @@ no_guards({clause, _, _, Guards, _}) ->
     Guards == [].
 
 
+fail_if_unreachable_clause(VarBinds, P) ->
+    maps:fold(fun
+                  (_K,?type(none),_Acc) ->
+                      throw({type_error, unreachable_clause, P});
+                  (_K,_V,Acc) ->
+                      Acc
+              end,
+             VarBinds,
+             VarBinds).
+
 %% Here at least one GuardSeq should be true, so we calculate the least upper bound
-check_guards(Env, Guards) ->
+check_guards(Env, Guards, P) ->
     X = lists:map(fun(GuardSeq) ->
                           check_guards_sequence(Env, GuardSeq)
                   end, Guards),
     Glb = fun(_K, Ty1, Ty2) ->
                   normalize({type, erl_anno:new(0), union, [Ty1, Ty2]}, Env#env.tenv)
           end,
-    union_var_binds_help(X, Glb).
+    VarBinds = union_var_binds_help(X, Glb),
+    fail_if_unreachable_clause(VarBinds, P).
 
 %% Here, all Guards must be true, hence we calculate the greatest lower bound
 check_guards_sequence(Env, GuardSeq) ->
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -2915,6 +2915,12 @@ type_check_fun(_Env, {remote, _, _Expr, _}, Arity)->
            lists:duplicate(Arity, type(any))},
           {type,0,any,[]}]},
         []]}], #{}, constraints:empty()};
+type_check_fun(Env, Var = {var, _, _Fun}, Arity) ->
+    case do_type_check_expr(Env, Var) of
+        FunType = {{type,0,'fun', [{type,0,product,Vars}, {type,0,any,[]}]}
+                 ,_,_} when length(Vars) =:= Arity -> FunType;
+        _ -> type_check_expr(Env, Var)
+    end;
 type_check_fun(Env, Expr, _Arity) ->
     type_check_expr(Env, Expr).
 
@@ -3218,24 +3224,25 @@ check_clauses(Env, ArgsTy, ResTy, Clauses) ->
         {RefinedTys :: [type()] , VarBinds :: map(), constraints:constraints()}.
 check_clause(_Env, [?type(none)|_], _ResTy, {clause, P, _Args, _Guards, _Block}) ->
     throw({type_error, unreachable_clause, P});
-check_clause(Env, ArgsTy, ResTy, C = {clause, P, Args, Guards, Block}) ->
+check_clause(_Env, ArgsTy, _ResTy, {clause, P, Args, _Guards, _Block})
+  when length(ArgsTy) =/= length(Args) ->
+    throw({argument_length_mismatch, P, length(ArgsTy), length(Args)});
+check_clause(Env, ArgsTy, ResTy, C = {clause, _, Args, Guards, Block}) ->
     ?verbose(Env, "~sChecking clause :: ~s~n", [format_location(C, brief), typelib:pp_type(ResTy)]),
-    case {length(ArgsTy), length(Args)} of
-        {L, L} ->
-            {PatTys, _UBounds, VEnv2, Cs1} =
-                add_types_pats(Args, ArgsTy, Env#env.tenv, Env#env.venv),
-            EnvNew      = Env#env{ venv =  VEnv2 },
-            VarBinds1   = check_guards(EnvNew, Guards),
-            EnvNewest   = EnvNew#env{ venv = add_var_binds(EnvNew#env.venv, VarBinds1, Env#env.tenv) },
-            {VarBinds2, Cs2} = type_check_block_in(EnvNewest, ResTy, Block),
-            RefinedTys  = refine_clause_arg_tys(ArgsTy, PatTys,
-                                                Guards, Env#env.tenv),
-            {RefinedTys
-            ,union_var_binds([VarBinds1, VarBinds2, VEnv2], Env#env.tenv)
-            ,constraints:combine(Cs1, Cs2)};
-        {LenTy, LenArgs} ->
-            throw({argument_length_mismatch, P, LenTy, LenArgs})
-    end;
+    {PatTys, _UBounds, VEnv2, Cs1} = add_types_pats(Args, ArgsTy, Env#env.tenv, Env#env.venv),
+    EnvNew      = Env#env{ venv =  VEnv2 },
+    %%%
+    %% Here I want to simply check if the block inside the clause typechecks,
+    %% taking the refined guards into consideration
+    VarBinds1   = check_guards(EnvNew, Guards),
+    EnvNewest   = EnvNew#env{ venv = refine_var_binds(EnvNew#env.venv, VarBinds1, Env#env.tenv) },
+    {VarBinds2, Cs2} = type_check_block_in(EnvNewest, ResTy, Block),
+    %%%
+    RefinedTys  = refine_clause_arg_tys(ArgsTy, PatTys,
+                                        Guards, Env#env.tenv),
+    {RefinedTys
+     ,union_var_binds([VarBinds1, VarBinds2, VEnv2], Env#env.tenv)
+     ,constraints:combine(Cs1, Cs2)};
 %% DEBUG
 check_clause(_Env, _ArgsTy, _ResTy, Term) ->
     io:format("DEBUG: check_clause term: ~p~n", [Term]),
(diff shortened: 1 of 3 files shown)


src/typechecker.erl
```diff
@@ -2906,6 +2906,12 @@ type_check_fun(_Env, {remote, _, _Expr, _}, Arity)->
            lists:duplicate(Arity, type(any))},
           {type,0,any,[]}]},
         []]}], #{}, constraints:empty()};
+type_check_fun(Env, Var = {var, _, _Fun}, Arity) ->
+    case do_type_check_expr(Env, Var) of
+        FunType = {{type,0,'fun', [{type,0,product,Vars}, {type,0,any,[]}]}
+                 ,_,_} when length(Vars) =:= Arity -> FunType;
+        _ -> type_check_expr(Env, Var)
+    end;
```
zuiderkwast: I think we shouldn't report any type error for `fun_wrong_arity/1`.

```Erlang
-spec fun_wrong_arity(any()) -> boolean().
fun_wrong_arity(Fun) when is_function(Fun, 2) ->
    Fun(0);
fun_wrong_arity(_Fun) -> true.
```

Any() is the dynamic type, so we don't want static type errors for it. The same if this function would have no spec at all.

(Unless you've found out: `type_check_fun` (and `type_check_expr`, etc., the ones without `_in`) are used for checking code against no particular type, propagating types outward, types originating from calls to functions with specs. The ones with `_in` are checking code against a particular type.)

src/typechecker.erl
```diff
@@ -3391,16 +3398,108 @@ refinable(_) ->
...
+             VarBinds,
+             VarBinds).
+
+%% Here at least one GuardSeq should be true, so we calculate the least upper bound
+check_guards(Env, Guards, P) ->
+    X = lists:map(fun(GuardSeq) ->
+                          check_guards_sequence(Env, GuardSeq)
+                  end, Guards),
+    Glb = fun(_K, Ty1, Ty2) ->
+                  normalize({type, erl_anno:new(0), union, [Ty1, Ty2]}, Env#env.tenv)
```
zuiderkwast: Lub =

zuiderkwast: Is this code with `Guard1A, Guard1B; Guard2A, Guard2B` correct? I think it needs test cases.

Note that a guard with alternatives on different variables `when is_integer(A); is_integer(B)` can't really be refined, (except possibly if they are considered together, i.e. `{A, B} :: {integer(), _} | {_, integer()}`).

test/should_fail/fun_guard_info.erl
```diff
@@ -0,0 +1,17 @@
+-module(fun_guard_info).
+-export([
+         fun_wrong_arity/1,
+         guard_unreachable_branch/1
+        ]).
+
+-spec fun_wrong_arity(any()) -> boolean().
+fun_wrong_arity(Fun) when is_function(Fun, 2) ->
+    Fun(0);
+fun_wrong_arity(_Fun) -> true.
```
zuiderkwast: Not sure if we want this to fail. See comment above.

src/typechecker.erl
NelsonVides: Ok, that's first of all a distinction I wasn't aware of, and for which the function names don't help a bit :no_mouth:

But in this function, the guard is a refinement. I have no idea what kind of evil things I'm giving to this function, but if it enters the first branch, then I do have information about it: it is a function of arity exactly two. That's all I know, so I cannot check the type of the arguments or the returns, but I can check its top-level type and its arity.

As a matter of fact, I believe the BEAM is a strongly typed machine, just that Erlang is dynamically checked, and not statically. And that is what this guard is doing, a dynamic check, right?

src/typechecker.erl
NelsonVides: Yeah, bad variable naming sweat_smile

About the guards, yes, very good case, when is_integer(A); is_integer(B), my code is indeed incorrectly refining both as integers. Gonna get that fix!

NelsonVides: Ok, I have a question about the desired behaviour:

If we have something like
```erlang
-spec fun_different_variables(any(), any()) -> any().
fun_different_variables(A,B)
  when is_integer(A); is_integer(B) ->
    A ++ [B].
```

I assume that we want that to pass, as in that case we're just not having enough information for any refinement. But what about the case:

```erlang
-spec fun_different_variables(number() | list(), number() | list()) -> any().
fun_different_variables(A,B)
  when is_integer(A); is_integer(B) ->
    A ++ [B].
```

In that case, the guards give us no information, hence `A` inside of the clause will have type `number() | list()`, which is not promising enough and so we should fail, right?

If my assumption is correct here, then I have a ready solution for that :smiley:

src/typechecker.erl
zuiderkwast: Josef had this idea that only the specs are used as static type information, to have a clear rule for when there are type errors and when not. But now we're also using type specifiers in bitstring constructions (e.g. float in `<<(X + Y)/float>>`). Maybe it makes sense to also use this kind of guards as static type information. Or maybe not.

@josefs what's your thoughts on this?

src/typechecker.erl
zuiderkwast: AFAIK that's what we want, yes. However, `++` doesn't crash for non-list tail. It just returns an improper list. We're not supporting improper lists (yet) though.

test/should_fail/fun_guard_info.erl
```diff
@@ -15,3 +17,20 @@ guard_unreachable_branch(Name)
...
 guard_unreachable_branch(_Name) ->
     [].
+
+% We know here that for sure at least one of them is an atom,
+% so this operation has no chance to pass.
+-spec at_least_one_of_them_is_an_atom(any(), any()) -> any().
+at_least_one_of_them_is_an_atom(A,B)
+  when is_atom(A), is_integer(B);
+       is_integer(A), is_atom(B) ->
+    A + B.
```
zuiderkwast: What you're saying here is that we should treat `is_atom(A)` as `A :: atom()`. I'm not against this, but it's a new source of type information and we should consider what the implications are.

We have the following options, at least:

1. We accept these as type annotations
2. We don't use them, with the reasoning that code without specs shouldn't give type errors
3. We introduce an option.

src/typechecker.erl
zuiderkwast: Yes, I think the guard adds a dynamic check. The compiler adds a lot of dynamic checks, but then in the optimization passes, some of the checks are removed if it can be proven that it's covered by an earlier check such as these guards.

Anyway, if the spec of the function would be `-spec fun_wrong_arity(fun((_,_) -> _) | fun((_) -> _)) -> boolean()`, this refinement is definitely wanted. I'm just not so sure about any() and that is all about allowing a dynamic style and not giving any errors in such cases. With the `--infer` flag, we actually do give type errors even if there are no specs. That's sort of why that flag is not enabled by default.

josefs: Refining `any()` is clearly wrong to me, especially if the programmer has written the type as in your example @zuiderkwast.  That would defeat the purpose of the type.

But the idea of using guards as a source of type information is quite nice, in particular if inference is enabled.

NelsonVides: Right, so we can throw it all into type inference then. So a priority would be to have a way to test it, for example the `-gradualizer({infer, true}).` that we mentioned elsewhere. Then it's going to be tricky to distinguish in code if I want to refine `-spec fun_wrong_arity(fun((_,_) -> _) | fun((_) -> _)) -> boolean()` but not `-spec fun_wrong_arity(any()) -> boolean()`.

Then I'll just move all this code under that infer flag.

Uh, and is there anybody planning to refactor `typechecker.erl` a bit? :sob:

Zalastax: I'm not convinced that refining `any()` is the wrong thing to do in general (for Gradualizer it might be). This TypeScript code errors since g is refined from any to string. I find it to be very similar to our guard situation:
```TypeScript
function f(g: any): boolean {
    if (typeof g === "string") {
        return g
    }
    return true
}
```

zuiderkwast: any() means "something unknown between none() and everything". Consider this:

```Erlang
%% no spec
my_sleep(N) when is_integer(N) -> timer:sleep(N);
my_sleep(true) -> timer:sleep(1000);
my_sleep(false) -> ok.
```

If any() is refined to integer() in the first clause, a type error would be reported for calling timer:sleep/1, which requires non_negative_integer() | infinity. In this case, we can imagine the programmer knows that "of course it's a positive integer or a boolean", so this error is clearly not wanted.

In order to not fail for the above example, we can only refine it to *something unknown between none() and integer()*, i.e. we can infer an upper bound only. Perhaps we can be represent this using an intersection type `any() & integer()` (assuming & means intersection).

As a comparison, normalizing the *union* integer() | any() doesn't remove integer() from the type. This is because, when I implemented normalization of unions, I had some sort of intuition that this should represent *something unknown between integer() and everything*, i.e. any() with integer() as a lower bound.

This is what I think we should have with `infer` (so we can report an error for `fun_wrong_arity/1` but no error for `my_sleep/1`). Without infer, just any().

zuiderkwast: @NelsonVides, regarding not refining any() while refining other types shouldn't be that tricky. Any() is handled differently from other things almost everywhere in the code.

> Then I'll just move all this code under that infer flag.

Perfect.

> Uh, and is there anybody planning to refactor typechecker.erl a bit?

Yes, it's in the back of my head. Renaming the functions `type_check_xxx(_in)`, `add_type(s)_pat(s)` and `add_any_type(s)_pat(s)` is a first step. Not sure what to name them though. :-) Ideas are welcome; perhaps open an issue to keep discussions separate?

josefs: @Zalastax we should avoid using typescript as reference. That typesystem is unsound.

josefs: @zuiderkwast please open a ticket to discuss refactoring of `typechecker.erl` and post your ideas there.

josefs: @zuiderkwast I wonder whether there is any way that you can make your intuition about intersection and union types a little more formal? I know you may not have that much experience with formalizing type systems but it'd be nice to try. I'm happy to help, if you can give a little bit more detail about your thinking.

The reason I ask is because your ideas sound compelling. They are also different from the existing literature: https://www.irif.fr/~gc/papers/icfp17.pdf which makes them extra intriguing. And I don't think your ideas are necessarily wrong just because they're different from a published paper.

Let me just give a tldr of the paper: `any() & integer()` is the same as `integer()` and `any() | integer()` is `any()`.

zuiderkwast: I think this paper is saying exactly what I am trying to say. This is spot on! Thanks @josefs for finding it! And no, `any() & integer()` is not the same as `integer()`.

Here, from the paper, the bottom of page 2, an example with type casts (AKA type assertions):

> ```ML
> let f (condition : Bool) (x : (Int | Bool)) : (Int | Bool) =
>   if condition then
>     succ ((Int) x)
>   else
>     not ((Bool) x)
> ```
>
> The code above shows that the use of union types fixes the shortcoming we pointed out. However in order to ensure that the applications of `succ` and `not` are both well typed we also had to add two type casts that check at run-time whether the argument has the required type and raise an exception otherwise.

Then, on the next page:

> By using set-theoretic types in conjunction with gradual types, it is possible both to ensure that `f` will only be applied to booleans or integers *and* to delegate the insertion of type casts to the system. This is shown by the following piece of code that our system will compile into the previous one.
>
> ```ML
> let f (condition : Bool) (x : (Int | Bool) & ?) : (Int | Bool) =
>   if condition then
>     succ x
>   else
>     not x
> ```
>
> In this example, the variable is of type `((Int | Bool) & ?)`, where “&” denotes an intersection. This indicates that x has both type `(Int | Bool)` and type `?`. Intuitively, this means that the function `f` accepts as a second argument a value of any type (which is indicated by `?`), as long as this value is also an integer *or* a Boolean. The effect of having added “& ?” in the type of the parameter is that the programmer is no longer required to add the explicit casts in the body of the function: the system will take care of it at compile time.

In other words, when adding "& any()" to a type, the type assertions can be removed. The type checker doesn't report a type error.

The explanation continues in the paper:

> The combination of a union type with “& ?” to type a parameter corresponds to a programming style in which the programmer asks the system to *statically* enforce that the function will be applied only to arguments in the union type and delegates to the system the *dynamic* checking, where/if necessary, [...]

Isn't this spot on?

NelsonVides:
@zuiderkwast, or maybe @josefs, I need some of your help. I'm working now on the case that Viktor mentioned:
```erlang
my_sleep(N) when is_integer(N), N > 0 -> timer:sleep(N);
my_sleep(true) -> timer:sleep(1000);
my_sleep(false) -> ok.
```
When the infer flag is given. In that case, this function should be perfectly safe, but I don't know how to tell to Gradualizer that by the time we arrive to `timer:sleep(N)`, then `N` is for sure a `non_neg_integer()`.

The problem is that in erlang, `N > 0` doesn't mean that `N` is a number, it means that it's basically anything in the universe except a non-negative number. And I'm not sure how to merge the deduced types from `is_integer(N)` and `N > 0`. And also in a way that doesn't depend on the order of checking (`when is_integer(N), N > 0` vs `when N > 0, is_integer(N)`). I'm having a hunch that that third field I've never know what is it good for, the `constraints`, could help me here, but I don't know how to use them. I'm also having a feeling for that `type_check_rel_op` function, it might be there where I want to apply my change.

Anybody has any suggestion? :slightly_smiling_face:

zuiderkwast:
From N > 0, can you produce `N :: pos_integer() | float() | atom() | ... | pid() | everything else ...`. From `is_integer(N)` we get `N :: integer()`. When this type is merged with integer() using LUB, we get `N :: pos_integer()`. Can work?

zuiderkwast:
Btw, my example didn't have the guard N > 0. In order to make sure we don't give any type error for my original my_sleep/1, even with inference enabled, I think we need to take any() into account as well, which makes it complicated (i.e. intersection type with any()). Then, the scope of this PR grows too large, so I don't want this to be done within this PR. 

Better focus on refining known types using guards. Also, to keep it minimal, perhaps only consider the is_TYPE guards and not N > 0 and similar. Whenever there is something that we don't handle yet, just skip refining. That's better than false errors IMO.

src/typechecker.erl
josefs: > Isn't this spot on?

No :-)

The part of the paper you're quoting has to do with how to compile the program in such a way that those parts which are statically typed are compiled without runtime checks. This is called "sound gradual typing" and means that whenever the execution switches from dynamically typed to statically typed, the code needs to first check that the values are indeed of the type that the static code expects them to be. The paper uses the word "cast" for this check. This part is tricky and that's what they're talking about in your quotes. 

In gradualizer we are not concerned with casts because after typechecking we throw away the types and compile the code as if it was completely dynamically typed.

zuiderkwast: Correct me if I'm wrong below.

Let's say we annotate every expression with a type using erl_anno. If the type system assigns an unknown type to an expression, it delegates the type checking to the runtime, which means that the compiler needs to insert type checks whenever the type is unknown where a specific type is needed. If the type system instead assigns a static type to an expression, it can be trusted by the compiler. That's the gradual guarantee, right?

If we would tag the expressions with a static type and the compiler would actually make use of this type, our type assertions would actually have to make sure that the value is of the that type, e.g. by inserting code asserts the type in runtime and raises an exception if it isn't (e.g. `?assert(is_integer(X))`). Only then, the type assertion can soundly assert a static type, which the compiler can trust. The only reason we don't is that the compiler treats every type as unknown anyway and inserts runtime checks everywhere.

I can't see how the rules for intersection and union should be different just because the current compiler doesn't make use of our type information.

FrankBro:
Should we close this since the other guard stuff was incorporated? I know there's some extra stuff in here. Maybe worth exploring in other PRs?

zuiderkwast:
Implemented in #241. It seems this PR has greatly inspired Frank's PR, so thanks @NelsonVides!

FrankBro:
Mine def. wouldn't be up without @NelsonVides work. Cheers 🎉

NelsonVides:
Thanks guys! Gradualizer is only getting better and better 🙂

PR #207, opened 2019-10-17
Add known problem for varbind in case

Zalastax:
N is supposed to be bound after the case end. Once fixed, the function should be moved to the module with the same name in should_pass.

### DIFF
--- a/test/known_problems/should_pass/varbind_in_case.erl
+++ b/test/known_problems/should_pass/varbind_in_case.erl
@@ -1,0 +1,13 @@
+-module(varbind_in_case).
+
+-compile([export_all, nowarn_export_all]).
+
+-spec tuple_bind({foo, integer()} | {bar, float()}) -> number().
+tuple_bind(A) ->
+    case A of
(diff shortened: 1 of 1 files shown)


test/known_problems/should_pass/varbind_in_case.erl
```diff
@@ -0,0 +1,13 @@
+-module(varbind_in_case).
+
+-compile([export_all, nowarn_export_all]).
+
+-spec tuple_bind({foo, integer()} | {bar, float()}) -> number().
+tuple_bind(A) ->
+    case A of
+        {foo, N} ->
+            ok;
```
zuiderkwast: It should probably return `N` instead of `ok`.

Zalastax:
There's no refinement as far as I know, only variable binding.

josefs:
Fantastic bug report! I've always suspected that we have scoping bugs because we haven't spent much time chasing after them and we don't really have any tests for them.
I'll fix this one, I know where the problem is.

PR #208, opened 2019-10-17
Unify tuple errors with common format

Zalastax:
Infer the type of the tuple before throwing. This lets us use the common type error format. I'm going to go through the code and find more places like this!

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -2152,7 +2152,7 @@ do_type_check_expr_in(Env, Ty, {bin, _Anno, _BinElements} = Bin) ->
           end,
     {_Ty, VarBinds, Cs2} = type_check_expr(Env, Bin),
     {VarBinds, constraints:combine(Cs1, Cs2)};
-do_type_check_expr_in(Env, ResTy, {tuple, LINE, TS}) ->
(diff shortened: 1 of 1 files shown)


src/typechecker.erl
```diff
@@ -2161,7 +2161,8 @@ do_type_check_expr_in(Env, ResTy, {tuple, LINE, TS}) ->
         {elem_tys, Tyss, Cs} ->
             case type_check_tuple_union_in(Env, Tyss, TS) of
                 none ->
-                    throw({type_error, tuple, LINE, ResTy});
+                    {Ty, _VB, _Cs} = type_check_expr(Env#env{infer = true}, Tup),
+                    throw({type_error, Tup, Ty, ResTy});
```
zuiderkwast: Aha! Clever trick using infer, just to get a nice error message!

josefs:
Can you give an example of an error message before an after this change?

Zalastax:
Before in `test/should_fail/tuple_union.erl`:
```
The tuple on line 7 at column 5 does not have type {undefined, binary()} | {integer(), undefined}
```
Now:
```
The expression {undefined, undefined} on line 7 at column 5 is expected to have type {undefined, binary()} | {integer(), undefined} but it has type {undefined, undefined}
```

PR #209, opened 2019-10-17
Add beams as Make dependency for shell target

Zalastax:
Just makes it so Make recompiles beam files before starting the shell if needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ ebin/gradualizer.app: src/gradualizer.app.src | ebin
 	cp $< $@
 
 .PHONY: shell
-shell:
+shell: app
 	erl -pa ebin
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -57,7 +57,7 @@ ebin/gradualizer.app: src/gradualizer.app.src | ebin
 	cp $< $@
 
 .PHONY: shell
-shell:
+shell: $(beams)
```
zuiderkwast: Maybe also `ebin/gradualizer.app` should be a prerequisite? I think it may be needed by `application` which is started on demand within the code.

Alternatively, just add `app` as a prerequisite.

Zalastax: I thought it wasn't but you're most likely right

PR #212, opened 2019-10-18
Handle -gradualizer(Options) in modules

zuiderkwast:
This is tested using the infer option in a test module.

The code is inspired by how it's done in `compile` in OTP.


NelsonVides:
Great! Thank you!

Issue #213, opened 2019-10-19
Dash or underscore for options?


josefs:
We have three ways of specifying options for gradualizer: on the command line, as term in a list when calling through the `gradualizer` api, and using the newly implemented `-gradualizer()` directive (#212, thanks @zuiderkwast). 

It is good that we have all these variants. But. They don't agree on the spelling of the options! For example, we have "--fmt-location" on the command line but `fmt_location` in the `gradualizer` api and in directives. A similar problem for "no-infer" and `no_infer`, "no-prelude" and `no-prelude` and so on.

I don't like this discrepancy and I'd like to get rid of it. I propose that we keep the atoms as they are and change the command line arguments to use underscore.

The root of the problem is that dash ("-") is not part of the atom syntax so if we were to use that, it'd be ergonomically unfriendly. We'd have to quote the atoms and that feels unnecessary. On the other hand, using underscores on the command line is somewhat non-idiomatic so it will look a little weird. But I prefer that in this case. I'd like to optimize for the use of the api and the directives.

zuiderkwast:
Dash as word-separator is common practice for shell commands. Options starting with a dash for single-letter optinos and double-dash for longer ones is common practice. Within Erlang, atoms with underscore is naturally used. I think we are following all the conventions. I wouldn't like `gradualizer --options '[{infer, false}, {fmt_location, brief}]' *.erl` for the shell. I like it the way it is.

zuiderkwast:
Btw, `--no-prelude` should correspond to `{prelude, false}`. I'm not sure it does, but that would be the convention I would expect.

zuiderkwast:
Sorry for spamming. I was thinking we could have a look at other tools such as dialyzer and compiler, how they do it. For dialyzer, they use a hybrid style like `--check_plt` vs `{check_plt, boolean()}`.

josefs:
Thanks for bringing up dialyzer as an example, @zuiderkwast. What I meant to say is that I'd gradualizer to have command line arguments in the same style as dialyzer, i.e. using underscore instead of dash. I **don't** propose that we write erlang terms on the command line. 

As for how the options are treated in the gradualizer api and in directives, I don't have a strong opinion on whether we use `{infer, boolean()}` the presence of absence of the `infer` atom in a list.

zuiderkwast:
Sounds good @josefs. This way, the CLI<->proplist mapping is simpler.

How should we represent the `false` valued AKA `--no-` options then?

[...]

PR #214, opened 2019-10-20
Make parts of the CLI module testable with coverage

zuiderkwast:
I factored out the conversion from CLI parameters to proplist options. This is sort of to highlight what we have in terms of CLI parameters. There are some inconsistencies, but let's talk in #213.

### DIFF
--- a/src/gradualizer_cli.erl
+++ b/src/gradualizer_cli.erl
@@ -6,6 +6,9 @@ main(Args) ->
     case handle_args(Args) of
         help    -> print_usage();
         version -> print_version();
+        {error, Message} ->
+            io:format(standard_error, "~s~n", [Message]),
+            halt(1);
         {ok, Files, Opts} ->
             case gradualizer:type_check_files(Files, Opts) of
                 ok -> ok;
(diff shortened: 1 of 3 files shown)


src/gradualizer_cli.erl
```diff
@@ -101,7 +114,7 @@ parse_opts([A | Args], Opts) ->
         "--stop-on-first-error"    -> parse_opts(Args, [stop_on_first_error | Opts]);
         "--no-stop-on-first-error" -> parse_opts(Args, [{stop_on_first_error, false} | Opts]);
         "--crash-on-error"         -> parse_opts(Args, [crash_on_error | Opts]);
-        "--no-crash-on-error"      -> parse_opts(Args, [{no_crash_on_error, false} | Opts]);
+        "--no-crash-on-error"      -> parse_opts(Args, [{crash_on_error, false} | Opts]);
```
josefs: Nice catch 😀

zuiderkwast: Yes, the tests helped me find it, believe it or not. :-)

josefs: Tests that actually help? Amazing!

PR #215, opened 2019-10-20
Enable overriding cli options with attributes

josefs:
The PR #212 introduced the `-gradualizer()` attribute, which allows
setting options for gradualizer in a module. I would have expected
these directives to override options on the command line, which is the
behaviour in other compilers and in dialyzer. This patch makes sure
that the attributes override cli options properly.

This patch also changes slightly how some options are treated internally.

### DIFF
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -176,7 +176,7 @@ type_check_forms(Forms, Opts) ->
                             ok | nok | [{file:filename(), any()}].
 type_check_forms(File, Forms, Opts) ->
     ReturnErrors = proplists:get_bool(return_errors, Opts),
-    OptsForModule = Opts ++ options_from_forms(Forms),
+    OptsForModule = normalize_options(options_from_forms(Forms), Opts),
     Errors = typechecker:type_check_forms(Forms, OptsForModule),
     case {ReturnErrors, Errors} of
         {true, _ } ->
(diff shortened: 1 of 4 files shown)


src/gradualizer.erl
```diff
@@ -196,3 +196,13 @@ options_from_forms([{attribute, _L, gradualizer, Opt} | Fs]) ->
     [Opt | options_from_forms(Fs)];
 options_from_forms([_F | Fs]) -> options_from_forms(Fs);
 options_from_forms([]) -> [].
+
+%% Normalize options so that module local options override options
+%% from the top level
+normalize_options(ModOpts, Opts) ->
+    proplists:normalize(ModOpts ++ Opts ++ [prelude] % using prelude is the default
+		       ,[{negations, [{no_infer, infer}
+				     ,{no_prelude, prelude}
```
zuiderkwast: `--no-infer` is returned as `{infer, false}` from CLI already.

josefs: Good point.

josefs:
Yeah, I think you're right that using `proplists:normalize` here is a bit overkill. I got a little carried away when I saw the function but I don't think we need it right now.

josefs:
> Btw, perhaps we should use a map of the options? It's better for type checking

That'd be nice, yes.

PR #216, opened 2019-10-20
Change cli options to use underscores to match the erlang api

josefs:
This patch implements the suggestion in #213 to harmonize the way
options are written. It used to be that we have e.g. `--no-infer` in
the command line but `no_infer` when calling gradualizer from the
api. Now we will use `--no_infer` on the command line.

This spelling of the options matches the convention of dialyzer.

### DIFF
--- a/src/gradualizer_cli.erl
+++ b/src/gradualizer_cli.erl
@@ -65,27 +65,27 @@ print_usage() ->
     io:format("  -h,  --help                    display this help and exit~n"),
     io:format("       --infer                   Infer type information from literals and other~n"),
     io:format("                                 language constructs~n"),
-    io:format("       --no-infer                Only use type information from function specs~n"),
+    io:format("       --no_infer                Only use type information from function specs~n"),
     io:format("                                  - the default behaviour~n"),
     io:format("       --verbose                 Show what Gradualizer is doing~n"),
-    io:format("  -pa, --path-add                Add the specified directory to the beginning of~n"),
+    io:format("  -pa, --path_add                Add the specified directory to the beginning of~n"),
     io:format("                                 the code path; see erl -pa             [string]~n"),
     io:format("  -I                             Include path for Erlang source files; see -I in~n"),
     io:format("                                 the manual page erlc(1)~n"),
-    io:format("       --print-file              prefix error printouts with the file name~n"),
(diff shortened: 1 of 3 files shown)


src/gradualizer_cli.erl
```diff
@@ -65,27 +65,27 @@ print_usage() ->
     io:format("  -h,  --help                    display this help and exit~n"),
     io:format("       --infer                   Infer type information from literals and other~n"),
     io:format("                                 language constructs~n"),
-    io:format("       --no-infer                Only use type information from function specs~n"),
+    io:format("       --no_infer                Only use type information from function specs~n"),
     io:format("                                  - the default behaviour~n"),
     io:format("       --verbose                 Show what Gradualizer is doing~n"),
-    io:format("  -pa, --path-add                Add the specified directory to the beginning of~n"),
+    io:format("  -pa, --path_add                Add the specified directory to the beginning of~n"),
```
zuiderkwast: For consistency with `erl` and `erlc`, while keeping the interface small, I think we should remove this alias and instead add `-pz`. Only `-pa` and `-pz` exist for erl and erlc, with no longer alias. (This is a CLI only option. From within Erlang, use `code:add_pathsa/1` and friends.)

src/gradualizer_cli.erl
```diff
@@ -65,27 +65,27 @@ print_usage() ->
...
-    io:format("       --print-file              prefix error printouts with the file name~n"),
+    io:format("       --print_file              prefix error printouts with the file name~n"),
     io:format("                                  - the default when checking a directory or more~n"),
     io:format("                                    than one file~n"),
-    io:format("       --print-basename          prefix error printouts with the file basename~n"),
-    io:format("       --print-module            prefix error printouts with the module~n"),
-    io:format("       --no-print-file           inverse of --print-file~n"),
+    io:format("       --print_basename          prefix error printouts with the file basename~n"),
+    io:format("       --print_module            prefix error printouts with the module~n"),
+    io:format("       --no_print_file           inverse of --print-file~n"),
```
zuiderkwast: These correspond to some special values, e.g. `{print_file, module}`. I suggest we change these a future to one of the styles `--print_file=module` or `--print_file module` (perhaps also with changed name of the option), so we get a 1-1 mapping between erlang and CLI.

PR #221, opened 2019-10-21
Defining types in gradualizer_prelude now works

gomoripeti:
hurray

### DIFF
--- a/test/should_pass/listsspecs.erl
+++ b/test/should_pass/listsspecs.erl
@@ -34,10 +34,9 @@ filtermap() ->
 flatmap() ->
     lists:flatmap(fun (X) -> integer_to_list(X) end, [1,2,3]).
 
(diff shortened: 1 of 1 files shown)


gomoripeti:
btw why is that that a recursive spec constraint gives the bellow warning but a recursive type does not
```
The type spec fun((deep_list) -> list() when deep_list :: [any() | deep_list]) has a cyclic dependency in variable deep_list
```

josefs:
> btw why is that that a recursive spec constraint gives the below warning but a recursive type does not

It would probably be possible to support recursive constraints. But it's not a super-useful feature and it would require quite a lot of effort to get it to work.

gomoripeti:
yes, I guess all cases of recursive constraints can be expressed by recursive local types and then issue is gone.

PR #223, opened 2019-10-21
Unify errors

Zalastax:
A few more tests, a few more errors that are handled the common way!

Before on `test/should_fail/call.erl`
```
The function f on line 11 at column 5 is expected to return boolean() but it returns {}
Expected function B on line 15 at column 5 to have a function type,
but it has the following type:
true | false
The function F on line 19 at column 5 is expected to return integer() but it returns boolean()
```
After:
```
The expression f() on line 11 at column 5 is expected to have type boolean() but it has type {}
The variable B on line 15 at column 5 is expected to have type fun() but it has type true | false
The expression F() on line 19 at column 5 is expected to have type integer() but it has type boolean()
```

Before on `bin/gradualizer test/should_fail/arith_op.erl`
```
...
The 'bnot' expression on line 36 at column 9 has a non-integer argument  of type [char()]
```

After:
```
...
The variable S on line 36 at column 14 is expected to have type integer() but it has type [char()]
```

Before on `test/should_fail/logic_op.erl`:
```
...
The 'not' expression on line 42 at column 9 has a non-boolean argument of type integer()
```
After:
```
...
The expression not N on line 42 at column 9 is expected to have type boolean() but it has type integer()
```


Zalastax:
Done!

Highlight from ` test/should_fail/logic_op.erl`:
```
The 'andalso' expression X
andalso
{} on line 33 at column 21 is expected to have type tuple() but it has type {} | false
The variable O on line 38 at column 5 is expected to have type tuple() but it has type false | false | true
The negation not N on line 42 at column 9 is expected to have type boolean() but it has type integer()
```

Issue #224, opened 2019-10-22
Unexpected error when using annotated types


erszcz:
I'm experiencing a strange type error with the following code:

```erlang
    66	-spec handle_purge_message_form(From :: jid:jid(), ArcJID :: jid:jid(),
    67	                                IQ :: jlib:iq()) ->
    68	    jlib:iq() | ignore | {error, term(), jlib:iq()}.
    69	handle_purge_message_form(#jid{} = From, #jid{} = ArcJID,
    70	                          #iq{xmlns=_MamNs, sub_el = QueryEl} = IQ) ->
    71	    ...
```

The error is:

```
The record patterns for record #jid on line 69 at column 27 is expected to have type From :: jid:jid().
```

When I modify the spec not to use the type variable `From`:

```erlang
-spec handle_purge_message_form(From :: jid:jid(), ArcJID :: jid:jid(),
                                IQ :: jlib:iq()) ->
    ...
```

the error is gone, but immediately an identical one for `ArcJID` pops up for the same line. Same happens for `IQ` when I proceed. Finally, no errors for line 69 are reported once the spec is free from type vars:

```erlang
-spec handle_purge_message_form(jid:jid(), jid:jid(), jlib:iq()) ->
    ...
```

This seems completely unexpected to me. Might there be a bug in some type equivalence checks between types named with a variable and same types, but with no explicit names attached?

zuiderkwast:
Seems to fail on the handling of annotated types, yes. Thanks for reporting the bug!

Feel free to add a minimal `known_problems` test case!

(For the solution, I think these should be removed in the normalization, which perhaps isn't happening where it should in some case.)

erszcz:
I've minimised the problem to:

```erlang
-record(r, {}).
-type r() :: #r{}.

-spec named_record_arg(R :: r()) -> {ok, R}.
named_record_arg(#r{} = R) -> {ok, R}.
```

Interestingly, the problem disappears when we rewrite it as:

```erlang
-spec named_record_arg(R) -> {ok, R} when
      R :: r().
named_record_arg(#r{} = R) -> {ok, R}.
```

[Here's the test case](https://github.com/josefs/Gradualizer/pull/225).

NelsonVides:
> Interestingly, the problem disappears when we rewrite it as:
> 
> ```erlang
> -spec named_record_arg(R) -> {ok, R} when
>       R :: r().
> named_record_arg(#r{} = R) -> {ok, R}.
> ```
[...]

Issue #180, opened 2019-04-05
Meaning of annotated types


gomoripeti:
Existing documentation:

- The erlang documentation for types and specs does not mention annotated types at all http://erlang.org/doc/reference_manual/typespec.html
- The Erlang documentation of the abstract format has the following to say (http://erlang.org/doc/apps/erts/absform.html#types)
  > If T is an annotated type `A :: T_0`, where `A` is a *variable*, then Rep(T) = `{ann_type,LINE,[Rep(A),Rep(T_0)]}`.
- The Elixir documentation states (https://hexdocs.pm/elixir/typespecs.html#defining-a-specification)
  > You can also name your arguments in a typespec using `arg_name :: arg_type` syntax.

From all this I deduct that in an annotated type `A :: T` `A` is not a type variable or type, and `::` does not says that "A is a *subtype* of T", but rather "A is a *name* or *variable name*, the value of which has the type T". It is common practice in Erlang and Elixir code that A is the same as the variable name in the head of the function implementation.

Although Gradualizer can give any meaning to existing type/spec constructs but I think the recent tendency was more to support the existing usage patterns and to be able to support as much existing code as possible. (A good example is the practical usage of what looks like bounded polymorphic types)

Therefore I suggest to treat the annotation of annotated types as shear documentation and ignore it as soon in the process as possible. @josefs @zuiderkwast @Zalastax @UlfNorell please share your views in this thread!

zuiderkwast:
Another possibility would be to treat them as introducing a type alias, so that

```Erlang
-spec foo(A :: string(), A) -> ok.
```

means the same as

```Erlang
-spec foo(A, A) -> ok when A :: string().
```

I've never seen any code using them like this though, so I think just dropping them makes more sense.

What does dialyzer do with a spec such as the first one above? Is it even valid?

Zalastax:
@zuiderkwast annotated types sure look a lot like type variables used in specs but I'm not certain they are the same. It makes sense for them to be the same so that type variables can be expressed in `fun` types, but `fun` types don't support conjunctions so we can't expect feature parity.

Type variables are weird so your spec actually says that `foo` accepts two strings that have to be the same...

(from [typespec](http://erlang.org/doc/reference_manual/typespec.html)):

Type variables can be used in specifications to specify relations for the input and output arguments of a function. For example, the following specification defines the type of a polymorphic identity function:
```Erlang
  -spec id(X) -> X.
```
[...]

PR #227, opened 2019-10-23
Evade the erlc compile bug

erszcz:
It turns out the previous PR for #224 triggered an `erlc` bug. This one fixes the test to sidestep that. Interesting that [the `erlc` issue is at least 3 years old](http://erlang.org/pipermail/erlang-questions/2016-October/090668.html).

@NelsonVides, thanks for catching that!

`erlc` can't compile the spec:

    -spec named_record_arg(R :: r()) -> {ok, R}.

and reports the following error:

    annotated_types_problems.erl:8: type variable 'R' is only used once (is unbound)

Let's sidestep this issue for this gradualizer test.

### DIFF
--- a/test/known_problems/should_pass/annotated_types_problems.erl
+++ b/test/known_problems/should_pass/annotated_types_problems.erl
@@ -5,5 +5,5 @@
 -record(r, {}).
 -type r() :: #r{}.
 
--spec named_record_arg(R :: r()) -> {ok, R}.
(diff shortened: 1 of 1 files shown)


Zalastax:
For clarity, `R` is not a type variable, it's an annotated type. We've discussed this confusing topic in #180.

erszcz:
I think #228 solves this problem - I'm closing this one.

PR #229, opened 2019-10-27
Add tests for undefined things

Zalastax:
Should increase coverage somewhat!

How does one run into the remote_record error? I.e. how does one reach https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L342?

### DIFF
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -100,12 +100,12 @@ format_type_error({undef, Type, Anno, {Module, Name, Arity}}, Opts)
        Name,
        Arity,
        format_location(Anno, verbose, Opts)]);
-format_type_error({not_exported, remote_type, {{atom, Anno, _} = Module, Name, Arity}}, Opts) ->
+format_type_error({not_exported, remote_type, Anno, {Module, Name, Arity}}, Opts) ->
     io_lib:format(
       "~sThe type ~s:~s/~p~s is not exported~n",
       [format_location(Anno, brief, Opts),
-       pp_expr(Module, Opts),
-       pp_expr(Name, Opts),
+       Module,
+       Name,
        Arity,
        format_location(Anno, verbose, Opts)]);
 format_type_error({illegal_pattern, Pat}, Opts) ->
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -169,18 +169,6 @@ format_type_error({type_error, call_intersect, Anno, FunTy, Name}, Opts) ->
        pp_expr(Name, Opts),
        format_location(Anno, verbose, Opts),
        pp_intersection_type(FunTy, Opts)]);
-format_type_error({type_error, mfa, Anno, M, F, A, ResTy, FunTy}, Opts) ->
-    io_lib:format(
-      "~sThe mfa ~p:~p/~p~s is expected to have type : ~n~s~n"
-      "but has type : ~n"
-      "~s~n",
-      [format_location(Anno, brief, Opts),
-       M,
-       F,
-       A,
-       format_location(Anno, verbose, Opts),
-       pp_type(ResTy, Opts),
-       pp_intersection_type(FunTy, Opts)]);
 format_type_error({type_error, expected_fun_type, Anno, Func, FunTy}, Opts) ->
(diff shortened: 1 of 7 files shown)


zuiderkwast:
Nice!!

> How does one run into the remote_record error?

In `undefined_errors_helper`, write a spec or a type containing an undefined record, e.g.

```Erlang
-spec f() :: #undefined_record{}.
```

In another module, use this type in some way, e.g. make the type checker check some expression against this type. If the type checker is smart enough (I don't remember if it is or not), it only needs to actually look up `#undefined_type{}` in the remote module if it is a compared against a record of the same name (or a tuple) defined in the currently checked module (or in a third module), because only then it can actually be type correct, e.g.

```Erlang
-record(undefined_record, {a, b, c}).
-spec foo() -> #undefined_record{}.
foo() -> undefined_errors_helper:f().
```

Zalastax:
Are we okay with not_exported tests being in undefined_errors? Should I rename/split? Suggestions welcome if current status is not OK!

zuiderkwast:
> Are we okay with the not_exported test being in undefined_errors? Should I rename/split? Suggestions welcome if current status is not OK!

Totally fine by me, but you could name it `not_exported` or so, instead of `h`.

Zalastax:
I noticed that several error printing clauses dealt with printing some combination of module/name/arity so I introduced the `#mna{}` record. Once that was done, the clauses folded onto each other and only one clause remained. I'm considering whether a map would be better since this becomes part of the external API.

src/mna.hrl
```diff
@@ -0,0 +1,4 @@
+-record(mna, {
+    module = undefined :: atom() | undefined,
+    name               :: atom(),
+    arity  = undefined :: integer() | undefined}).
```
zuiderkwast: By unifying the errors for undefined local and remote functions, records and types, the module and arity becomes optional in all of these type errors. That's expanding the type by allowing a lot more cases which don't exist in reality. For some errors we always have an arity and in other cases we don't. Therefore, I don't like this approach.

Zalastax: To me it's worse that the formatting of M:N/A is spread out in several cases since that both makes our code more verbose and it makes for an API that is more difficult to use. We can always add an intersection type when time comes that would remove the impossible combinations.  That's how I would do it in Agda for example - a fairly simple type combined with propositions that restrict it to make it exact.

Zalastax: I'll move this commit to another PR if you still don't agree.

zuiderkwast: Good idea. :-)

Looking at the types, it means anyone implementing this will need to be able to handle records like `mod:#rec{}/3`.

The error message where we refer to `mod:#rec{}` isn't even correct Erlang syntax, so we should probably write this as "`#rec{}` in the `mod` module" instead.

Another way is to represent the remote module as the file annotation in the type, e.g. as `{undef_record, `{record, [{location, {3,5}}, {file, "mod.erl"}], rec}}`. That's how gradualizer_db returns them anyway.

The interface for these type errors is quite large anyway, so I think a well-documented interface on this form would be hard to maintain. Instead, we can have an opaque error type and they can supply pretty-printing funs for expressions an types. That's a very clear interface IMO.

src/gradualizer_fmt.erl
```diff
@@ -82,37 +82,30 @@ format_type_error({undef, record_field, FieldName}, Opts) ->
...
     io_lib:format(
       "~sUndefined user type ~p/~p~s~n",
       [format_location(Anno, brief, Opts),
        Name,
        Arity,
        format_location(Anno, verbose, Opts)]);
-format_type_error({undef, user_type, Anno, {Module, Name, Arity}}, Opts) ->
+format_type_error({undef, Type, Anno, {Module, Name, Arity}}, Opts)
+  when Type =:= user_type; Type =:= remote_type ->
+    TypeS = case Type of user_type -> "user type"; remote_type -> "remote type" end,
```
zuiderkwast: I think "user-defined type" is the proper term, but "Undefined user-defined type" sounds really silly, since it's not defined in this case.

So, better just say "type" for user_type.

josefs:
Looks good to me! @Zalastax, are you ready to merge this?

PR #233, opened 2019-10-31
Add a test for a known crash on map update

erszcz:
We've just found a bug:

```
$ gradualizer update_map_type_error.erl
escript: exception error: no function clause matching
                 typechecker:update_map_type({var,25,'Acc'},
                                             [{type,
                                               {8,31},
                                               map_field_exact,
                                               [{atom,{8,31},id},
                                                {var,0,
                                                 "_TyVar-576460752303423485"}]},
                                              {type,
                                               {9,31},
                                               map_field_exact,
                                               [{atom,{9,31},name},
                                                {var,0,
                                                 "_TyVar-576460752303423453"}]},
                                              {type,
                                               {10,31},
                                               map_field_exact,
                                               [{atom,{10,31},members},
                                                {type,0,nonempty_list,
                                                 [{type,406,list,
                                                   [{type,406,byte,[]}]}]}]}]) (src/typechecker.erl, line 2866)
  in function  typechecker:do_type_check_expr_in/3 (src/typechecker.erl, line 2202)
  in call from typechecker:type_check_expr_in/3 (src/typechecker.erl, line 2076)
  in call from typechecker:check_clause/4 (src/typechecker.erl, line 3242)
  in call from typechecker:'-check_clauses/4-fun-0-'/4 (src/typechecker.erl, line 3204)
  in call from lists:foldl/3 (lists.erl, line 1263)
  in call from typechecker:check_clauses/4 (src/typechecker.erl, line 3202)
  in call from typechecker:do_type_check_expr_in/3 (src/typechecker.erl, line 2326)
```

### DIFF
--- a/test/known_problems/should_pass/update_map_type_error.erl
+++ b/test/known_problems/should_pass/update_map_type_error.erl
@@ -1,0 +1,15 @@
+-module(update_map_type_error).
+
+-compile(export_all).
+
+-spec t(list()) -> map().
+t(Rows) ->
+    lists:foldl(fun({Id, Name, Member}, #{members := []} = List) ->
+                        List#{id := Id,
+                              name := Name,
+                              members := [binary_to_list(Member)]};
+                   ({_, _, Member}, #{members := Members} = List) ->
+                        List#{members := [binary_to_list(Member) | Members]}
+                end, #{id => undefined,
+                       name => undefined,
+                       members => []}, Rows).


Zalastax:
For reference:
```Erlang
t([{1, "marco", <<"POLO">>}, {2, "foo", <<"BAR">>}]).
%Output:
#{id => 1,members => ["BAR","POLO"],name => "marco"}
```

erszcz:
I've minimised the problem case, traced the exception, and extended `update_map_type` to accept the current env - it's all in my fork in the same branch.

It seems that `{var,25,'Acc'}` is [the `Acc` from `lists:foldl/3` from `priv/prelude/lists.specs.erl`](https://github.com/josefs/Gradualizer/blob/efe8487e289e9426c1d3107464b1035ec23c4898/priv/prelude/lists.specs.erl#L25). Taking the example into account:

```erlang
-spec t(list()) -> map().
t(Rows) ->
    lists:foldl(fun(Id, List) ->
                        List#{id := Id}
                end, #{id => undefined}, Rows).
```

It also seems that `{var,25,'Acc'}` should not reach `update_map_type`, but instead its actual type determined as the type of `#{id => undefined}`, which is the initial value of `Acc`. At least that would seem logical to me. I'm not sure - should it be handled higher in the call chain?

Here's the trace for reference:

```
{trace,<0.308.0>,call,
    {typechecker,update_map_type,
        [{env,
             #{{t,1} =>
                   [{type,
                        {5,8},
                        bounded_fun,
                        [{type,
                             {5,8},
                             'fun',
                             [{type,{5,8},product,[{type,{5,9},list,[]}]},
                              {type,{5,20},map,any}]},
                         []]}]},
             #{},
             #{'Id' => {var,25,'T'},
               'List' => {var,25,'Acc'},
               'Rows' => {type,{5,9},list,[{type,{5,9},any,[]}]}},
             {tenv,update_map_type_error,#{},#{}},
             false,false,true},
         {var,25,'Acc'},
         [{type,{8,31},map_field_exact,[{atom,{8,31},id},{var,25,'T'}]}]]}}

{{2019,11,1},{11,21,7}}
{trace,<0.308.0>,exception_from,
       {typechecker,update_map_type,3},
       {error,function_clause}}
```

The pretty printed env is:

```erlang
#env{
    fenv =
        #{{t,1} =>
              [{type,
                   {5,8},
                   bounded_fun,
                   [{type,
                        {5,8},
                        'fun',
                        [{type,
                             {5,8},
                             product,
                             [{type,{5,9},list,[]}]},
                         {type,{5,20},map,any}]},
                    []]}]},
    imported = #{},
    venv =
        #{'Id' => {var,25,'T'},
          'List' => {var,25,'Acc'},
          'Rows' => {type,{5,9},list,[{type,{5,9},any,[]}]}},
    tenv =
        #tenv{
            module = update_map_type_error,types = #{},
            records = #{}},
    infer = false,verbose = false,exhaust = true}
```

zuiderkwast:
Nice minimising the case! It can be merged right away.

For the implementation, I looked at your branch and I can't see what `Env` is used for in `update_map_type/3`. Probably, a case for the type `{var, _, Var}` should be added. The type variable should matches any type but a constraint can be added.

erszcz:
> For the implementation, I looked at your branch and I can't see what `Env` is used for in `update_map_type/3`. I guess it's not ready yet?

Not ready at all. The env is not used, I was just tinkering to see if any information about `Acc` can be found within and tracing with `Env` present is more convenient.

> a function clause for the type {var, _, Var} should be added. The type variable should match any type but a constraint can be added.

Thanks! I'll give writing it a go.

Issue #231, opened 2019-10-28
Infinite loop when checking call from gradualizer_type:abstract_expr()


Zalastax:
Checking this causes Gradualizer to hang:
```Erlang
-spec describe_expr(gradualizer_type:abstract_expr()) -> io_lib:chars().
describe_expr({call, _, _, _})            -> "function call".
```

This is the cause of Travis running out of memory in #230.

zuiderkwast:
Because `abstract_expr()` is a very large type and also a recursive type, probably. A gradualizer_cache was introduced already for solving some performance issue, which perhaps can used in more cases. If it's not enough, perhaps we need to do pruning at some point, i.e. stop checking at some maximum depth. But first we need to pinpoint the bottle neck.

zuiderkwast:
I'm not sure it's an infinite loop; just inefficient. The efficient way would be to represent the types on Disjunctive Normal Form, using binary decision trees, as described in *G. Castagna: Covariance and Contravariance: a fresh look at an old issue (a primer in advanced type systems for learning functional programmers* ([contravarianceagain.pdf](https://www.irif.fr/~gc/papers/covcon-again.pdf)) and lifted to gradual types in *G. Castagna and V. Lanvin: Gradual Typing with Union and Intersection Types* ([icfp17.pdf](https://www.irif.fr/~gc/papers/icfp17.pdf)).

josefs:
I've had a look at the first paper that you mentioned @zuiderkwast. This paper uses a pretty big hammer (BDDs and other optimizations) to make testing for subtyping efficient. I think that's overkill for us. But that doesn't mean that there aren't ideas in there that we can use. 

Starting on page 33 the author shows a representation of unions which separates the different types of the unions into buckets so that, for instance, atoms would live in their own bucket and lists would live in another bucket. I think if we had different sized tuples live in different buckets then we could speed up typechecking of the example @Zalastax gave above, because the typechecker wouldn't have to go through all tuples in the union, only those with the same size.

zuiderkwast:
Really nice to see that you're following up on this!

Yes! Keeping them separate and in particular storing tuples by size sounds like a good idea! We'd have to change the code in all the places where we pattern match on unions.

Another thing they do in that paper is lazy unions.

Another thing we could do is to add caching in `subtype/3` and/or `normalize/2` as we have in `glb/3`. (Caching without copying can be done using the process dictionary instead of an ETS table. Not sure if it's worth the ugliness.)

[...]

Issue #222, opened 2019-10-21
Quoting expressions in error messages


josefs:
I think it'd be nice to quote expressions in the error message to distinguish them from the natural language text and improve readability. 

For example, right now we have:
> The expression X and X on line 8 at column 17 is expected to have type tuple() but it has type boolean()

But we could do something like: 

> The expression 'X and X' on line 8 at column 17 is expected to have type tuple() but it has type boolean()

Exactly how to quote is clearly up for bikeshedding.

zuiderkwast:
Shouldn't we quote types as well then?

In a terminal emulator, we could use colours...

Another way would be to print it below the message, with a little more context, e.g.

```
The logic expression on line 8 at column 17 is expected to have type tuple() but it has type boolean()

f(X, Y) ->
    tuple_to_list(X and Y).
                  ~~~~~~~
```

josefs:
> Shouldn't we quote types as well then?

That would make sense, yes.

> In a terminal emulator, we could use colours...

Nice!

Your example looks quite nice. I think there is room for experimenting a bit with how to best print these errors.

PR #234, opened 2019-11-02
Fancy error messages

zuiderkwast:
* Colors
* Expressions highlighted in their context (#222)
* A performance hack to fix a bottle neck (#231)

### DIFF
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -4,6 +4,8 @@
 -include("typelib.hrl").
 
 -define(FMT_LOCATION_DEFAULT, verbose).
+-define(color_type, "\e[35m"). % 35=magenta
+-define(color_end, "\e[0m").
 
 -spec format_location(any(), brief | verbose, any()) -> io_lib:chars().
 format_location(Expr, FmtType, Opts) ->
@@ -481,9 +483,11 @@ pp_intersection_type([Ty|Tys], Opts) ->
     pp_type(Ty, Opts) ++ ["\n" || Tys /= []] ++ pp_intersection_type(Tys, Opts).
 
 pp_type(Ty, Opts) ->
-    case proplists:get_value(fmt_type_fun, Opts) of
-        Fun when is_function(Fun) ->
-            Fun(Ty);
-        _ ->
-            typelib:pp_type(Ty)
+    PP = case proplists:get_value(fmt_type_fun, Opts) of
+             Fun when is_function(Fun) -> Fun(Ty);
+             _                         -> typelib:pp_type(Ty)
+         end,
+    case use_color(Opts) of
+        true  -> [?color_type, PP, ?color_end];
+        false -> PP
     end.
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -42,7 +42,7 @@ format_location(Expr, verbose) ->
 -spec format_type_error(any(), any()) -> io_lib:chars().
 format_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
   when is_tuple(Expression) ->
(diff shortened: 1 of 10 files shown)


zuiderkwast:
Btw, for `{color, auto}` I'm using `io:columns()` to detect if stdout is a TTY. It doesn't work from an escript, where it always `{error, enotsup}`. Is there any other way?

One way would be to start the CLI by calling `erl` from a shell script instead of an escript.

src/gradualizer_hilite.erl
```diff
@@ -0,0 +1,233 @@
...
+%% Highlights a node in the AST in a piece of source code. The node must
+%% annotated with line and column information for this to work. Returns
+%% an inlist().
+highlight_in_source(AstNode, Source, Color) ->
+    {Start, End} = find_start_and_end_location_in_source(AstNode, Source),
+    highlight_text(Source, Start, End, Color).
+
+%% Finds the start and end location of an abstract node in source code.
+%% This is done by pretty-printing, tokenizing, counting the tokens and
+%% then performing finding the corresponding tokens in the tokenized
```
Zalastax: then finding (remove performing)

test/gradualizer_hilite_tests.erl
```diff
@@ -0,0 +1,20 @@
+-module(gradualizer_hilite_tests).
+
```
Zalastax: gradualizer_highlight_tests?

src/gradualizer_hilite.erl
```diff
@@ -0,0 +1,233 @@
+-module(gradualizer_hilite).
+
```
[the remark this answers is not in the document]
zuiderkwast: it's too long :-) but I can change if you want

Zalastax:
I think some discussion is needed about the performance hack. Do we have problems with all unions? Is our approach too naive? How problematic is it to replace with any?

zuiderkwast:
> I think some discussion is needed about the performance hack. Do we have problems with all unions? Is our approach too naive? How problematic is it to replace with any?

I agree, but I would like to merge it first and discuss afterwards. Running out of memory is not OK. Let's discuss and/or find a proper solution in #231.

gomoripeti:
awesome work!

(I tried to understand how rebar3 escript does tty detection. I kindof remembered escripts use the old-shell. I couldn't find any auto-color detection snippet. I only found that `rebar3 shell` basically stops the old-shell and starts a new-type shell https://github.com/erlang/rebar3/blob/master/src/rebar_prv_shell.erl#L140 - the horror)

gomoripeti:
Just bumped into this erlang issue ticket about `io:columns` in escript fixed in OTP 22.1. Not too relevant but interesting https://bugs.erlang.org/browse/ERL-717

zuiderkwast:
> 
> 
> Just bumped into this erlang issue ticket about `io:columns` in escript fixed in OTP 22.1. Not too relevant but interesting https://bugs.erlang.org/browse/ERL-717

That's great! It means that automatic colors will work.

src/gradualizer_fmt.erl
```diff
@@ -371,20 +373,64 @@ format_type_error({bad_type_annotation, TypeLit}, Opts) ->
...
+             {ok, SourceBin} = file:read_file(SourceFile),
+             Source = binary_to_list(SourceBin),
+             gradualizer_highlight:highlight_in_source(AstNode, Source, Color)
+     end,
+     $\n]. % blank line after
+
+-spec use_color(gradualizer:options()) -> boolean().
+use_color(Opts) ->
+    case proplists:get_value(color, Opts, auto) of
+        auto -> case io:columns() of {ok, _} -> true; _ -> false end;
```
josefs: Can you explain the use of `io:columns/0` here? It's be nice to have a comment in the code detailing what's going on here.

PR #238, opened 2019-12-19
Returning incorrectly typed map should fail

erszcz:
This fixes the common and simple case of constructing a new map and returning it from a function. If the map does not typecheck a previously silent error is now reported.

All existing tests, including "should fail / should pass" tests for covariant map keys still pass.

Preview:

```
$ cat test/should_fail/map_entry.erl
-module(map_entry).
-export([f/0, g/0]).

-spec f() -> #{bepa := apa}.
f() -> #{apa => bepa}.

-type typed_map() :: #{field_a := atom()}.

-spec g() -> typed_map().
g() ->
    #{field_a => <<"ala ma kota">>}.
$ gradualizer -- test/should_fail/map_entry.erl
The map on line 5 at column 8 is expected to have type #{bepa := apa} but it has type #{apa => bepa}

-spec f() -> #{bepa := apa}.
f() -> #{apa => bepa}.
       ^^^^^^^^^^^^^^

The map on line 11 at column 5 is expected to have type typed_map() but it has type #{field_a => <<_:88>>}

-spec g() -> typed_map().
g() ->
    #{field_a => <<"ala ma kota">>}.
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```


erszcz:
Ahh, it's not that easy. Commit e83a161 adds a test case which fails now... so it's a WIP.

erszcz:
Ooops, it seems the last commit also fixed https://github.com/josefs/Gradualizer/pull/236. I'm rebasing onto master and pushing the changes again.

src/typechecker.erl
```diff
@@ -2181,16 +2181,16 @@ do_type_check_expr_in(Env, ResTy, {tuple, _, TS} = Tup) ->
     end;
 
 %% Maps
-do_type_check_expr_in(Env, ResTy, {map, _, Assocs} = Map) ->
+do_type_check_expr_in(Env, ResTy, {map, _, Assocs} = MapCreation) ->
     {AssocTys, VBs, Cs2} = type_check_assocs(Env, Assocs),
-    _MapTy = update_map_type(type(map, any), AssocTys),
-    case subtype(type(map, any), ResTy, Env#env.tenv) of
+    MapTy = update_map_type(type(map, []), AssocTys),
+    case subtype(MapTy, ResTy, Env#env.tenv) of
```
zuiderkwast: Is this correct? I guess you're aware that `{type, _, map, any}` represents `map()` = *any map* and `{type, _, map, []}` represents `#{}` = the empty map type (the type containing only the empty map).

I don't understand exactly what update_map_type does. A description (and possibly a spec) would be useful.

src/typechecker.erl
```diff
@@ -275,8 +275,8 @@ compat_ty({type, _, AssocTag2, [Key2, Val2]},
              AssocTag2 == map_field_exact, AssocTag1 == map_field_exact;
              AssocTag2 == map_field_assoc, AssocTag1 == map_field_exact ->
     %% For M1 <: M2, mandatory fields in M2 must be mandatory fields in M1
-    {A1, Cs1} = compat_ty(Key1, Key2, A, TEnv),
-    {A2, Cs2} = compat_ty(Val1, Val2, A1, TEnv),
+    {A1, Cs1} = compat(Key1, Key2, A, TEnv),
+    {A2, Cs2} = compat(Val1, Val2, A1, TEnv),
```
zuiderkwast: Good!

zuiderkwast:
It seems very good, but I don't completely understand the update_map_type code. Does it infer a map type, by adding associations starting from the empty map type?

src/typechecker.erl
erszcz: I wasn't sure myself what `update_map_type`'s purpose was, but having analysed the map creation and update expressions I now understand it's inferring type of the resulting (created or updated) map, before it can be type-checked.

> I guess you're aware that {type, _, map, any} represents map() = any map and {type, _, map, []} represents #{} = the empty map type (the type containing only the empty map).

Indeed. That's why I renamed `Map` to `MapCreation` and `MapUpdate` accordingly, to make the two clauses more distinct. In case of map creation, we know that the exact type can be fully determined from the analysed expression. Moreover, in case of map creation we know that the inferred type has only non-optional associations (i.e. the fields are already there when we build the map), so the expression's assocs are rewritten (by `update_assocs`).

src/typechecker.erl
erszcz: There's one more case like this - `compat_ty` calls `all_type` which (probably incorrectly) calls `compat_ty` again. I have not touched that case since I don't have a test case (or even an idea for one) which could be fixed or would verify if it's intended or not. Maybe worth a separate issue?

src/typechecker.erl
zuiderkwast: Great. Put some of this as comments in the code. :-)

If the clauses of update_map_type serve different purposes, perhaps it's better to split it into two different functions?

src/typechecker.erl
zuiderkwast: Good, keep the PR small.

I you want you can add a small comment such as `%% TODO: call compat/4 instead of compat_ty/4 here?`.

src/typechecker.erl
erszcz: > If the clauses of update_map_type serve different purposes, perhaps it's better to split it into two different functions?

I've refactored one of the other clauses somewhat - I think all the clauses do something close enough to each other to keep it a single function.

zuiderkwast: Nice! I'm convinced. Good code comments added. 👍

josefs:
LGTM

Issue #232, opened 2019-10-28
Intersection with any()


Zalastax:
There's lots of interesting discussions in https://github.com/josefs/Gradualizer/pull/206#discussion_r335621822 and I want to bring some of that out to a separate issue. In particular, I want to discuss the meaning of intersection with `any()`.

To summarize, @josefs argues that "`any() & integer()` is the same as `integer()`" whereas @zuiderkwast argues that "`any() & integer()` is not the same as `integer()`".

My personal take is that I agree with @zuiderkwast. To me, `T :: any() & integer()` means that `T` is of unknown type but it's at least an `integer()`. If we have `-spec f(5) -> integer().`, I believe that @josefs disallows `f(T)` whereas @zuiderkwast and I don't. My intuition is that `any()` behaves in the same manner as a new type variable (which in this example would be satisfied if it's type is `5`). I think you told me this was wrong during our last fika but I didn't follow. An example for why this thinking gets us to the wrong place would be very helpful!

As an extension to my thinking, I believe that if we also have `-spec g(6) -> integer()`, then we should be allowed to reject `f(T), g(T)` since `T` can't be both `5` and `6`. 

Relevant papers (I believe that the first few pages deal with this issue already):
https://www.irif.fr/~gc/papers/icfp17.pdf
https://www.irif.fr/~gc/papers/popl19.pdf

gomoripeti:
I did not follow the discussion, but for me also `any & integer()` _feels_ like an "unknown subset of all integers". But even if intersection with `any()` should not mean this, I think there would be value if internally Gradualizer could represent such notion, ie "an unknown subtype of a static type" or rather "a dynamic type that is although unknown, it is not a subtype of anything else than a <static type>".  This way Gradualizer could represent a partial information about a type. Sorry about the sidetrack.

zuiderkwast:
@josefs
> The reason I ask is because your ideas sound compelling. They are also different from the existing literature: https://www.irif.fr/~gc/papers/icfp17.pdf which makes them extra intriguing. And I don't think your ideas are necessarily wrong just because they're different from a published paper.

No need. The paper exactly confirms my ideas.

> Let me just give a tldr of the paper: any() & integer() is the same as integer() and any() | integer() is any().

This statement is completely wrong. I suggest you actually read the paper instead guessing. It explains "my ideas" pretty well.

josefs:
Yeah, I need to read the paper properly. It's on my todo list.

[...]

PR #241, opened 2019-12-26
Guard support

FrankBro:
Simple support for guard BIFs to allow constraining `term()` to something more specific.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3469,7 +3469,7 @@ check_guard(_Env, {call, _, {remote,_,_,{atom, _, Fun}}, Vars}) ->
 check_guard(Env, {op, _OrElseAnno, 'orelse', Call1, Call2}) ->
     G1 = check_guard(Env, Call1),
     G2 = check_guard(Env, Call2),
-    union_var_binds_lub([G1, G2], Env#env.tenv);
+    union_var_binds_symmetrical([G1, G2], Env#env.tenv);
 check_guard(Env, {op, _AndAlsoAnno, 'andalso', Call1, Call2}) ->
     G1 = check_guard(Env, Call1),
     G2 = check_guard(Env, Call2),
@@ -3495,7 +3495,7 @@ check_guards(Env, Guards) ->
     VarBinds = lists:map(fun (GuardSeq) ->
         check_guards_sequence(Env, GuardSeq)
     end, Guards),
-    VB = union_var_binds_lub(VarBinds, Env#env.tenv),
+    VB = union_var_binds_symmetrical(VarBinds, Env#env.tenv),
     VB.
 
 type_check_function(Env, {function,_, Name, NArgs, Clauses}) ->
@@ -3991,16 +3991,44 @@ is_power_of_two(N) when N rem 2 == 0 ->
     is_power_of_two(N div 2);
 is_power_of_two(_) -> false.
 
--spec union_var_binds_lub(list(), #tenv{}) -> map().
-union_var_binds_lub([], _) ->
+-spec union_var_binds_symmetrical(list(), #tenv{}) -> map().
+union_var_binds_symmetrical([], TEnv) ->
     #{};
-union_var_binds_lub(VarBindsList, TEnv) ->
+union_var_binds_symmetrical(VarBindsList, TEnv) ->
     Lub = fun(_K, Ty1, Ty2) ->
         UTy = type(union, [Ty1, Ty2]),
         NTy = normalize(UTy, TEnv),
         NTy
-          end,
-    union_var_binds_help(VarBindsList, Lub).
+    end,
+    union_var_binds_symmetrical_help(VarBindsList, Lub).
+
+-spec union_var_binds_symmetrical_help(list(), function()) -> map().
+union_var_binds_symmetrical_help([VB1, VB2 | Rest], Lub) ->
+    VB =
+        case maps:size(VB1) < maps:size(VB2) of
+            true ->
+                maps:fold(fun (K, Ty1, VB) ->
+                    case maps:is_key(K, VB2) of
+                        false -> VB;
+                        true ->
+                            Ty2 = maps:get(K, VB2),
+                            Ty = Lub(K, Ty1, Ty2),
+                            maps:put(K, Ty, VB)
+                    end
+                end, #{}, VB1);
+            false ->
+                maps:fold(fun (K, Ty2, VB) ->
+                    case maps:is_key(K, VB1) of
+                        false -> VB;
+                        true ->
+                            Ty1 = maps:get(K, VB1),
+                            Ty = Lub(K, Ty1, Ty2),
+                            maps:put(K, Ty, VB)
+                    end
+                end, #{}, VB2)
+        end,
+    union_var_binds_symmetrical_help([VB | Rest], Lub);
+union_var_binds_symmetrical_help([VB], _) ->  VB.
 
 union_var_binds(VB1, VB2, TEnv) ->
     union_var_binds([VB1, VB2], TEnv).
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3469,12 +3469,7 @@ check_guard(_Env, {call, _, {remote,_,_,{atom, _, Fun}}, Vars}) ->
 check_guard(Env, {op, _OrElseAnno, 'orelse', Call1, Call2}) ->
     G1 = check_guard(Env, Call1),
     G2 = check_guard(Env, Call2),
-    Lub = fun(_K, Ty1, Ty2) ->
-        UTy = {type, erl_anno:new(0), union, [Ty1, Ty2]},
-        NTy = normalize(UTy, Env#env.tenv),
-        NTy
-    end,
-    union_var_binds_help([G1, G2], Lub);
+    union_var_binds_lub([G1, G2], Env#env.tenv);
 check_guard(Env, {op, _AndAlsoAnno, 'andalso', Call1, Call2}) ->
     G1 = check_guard(Env, Call1),
     G2 = check_guard(Env, Call2),
@@ -3497,15 +3492,10 @@ check_guards_sequence(Env, GuardSeq) ->
 -spec check_guards(#env{}, list()) -> map().
 check_guards(Env, []) -> #{};
 check_guards(Env, Guards) ->
-    Lub = fun(_K, Ty1, Ty2) ->
-        UTy = {type, erl_anno:new(0), union, [Ty1, Ty2]},
-        NTy = normalize(UTy, Env#env.tenv),
-        NTy
-    end,
     Tys = lists:map(fun (GuardSeq) ->
         check_guards_sequence(Env, GuardSeq)
     end, Guards),
-    Ty = union_var_binds_help(Tys, Lub),
+    Ty = union_var_binds_lub(Tys, Env#env.tenv),
     Ty.
 
 type_check_function(Env, {function,_, Name, NArgs, Clauses}) ->
@@ -4000,6 +3990,17 @@ is_power_of_two(1) -> true;
 is_power_of_two(N) when N rem 2 == 0 ->
     is_power_of_two(N div 2);
 is_power_of_two(_) -> false.
+
+-spec union_var_binds_lub(list(), #tenv{}) -> map().
+union_var_binds_lub([], _) ->
+    #{};
+union_var_binds_lub(VarBindsList, TEnv) ->
+    Lub = fun(_K, Ty1, Ty2) ->
+        UTy = type(union, [Ty1, Ty2]),
+        NTy = normalize(UTy, TEnv),
+        NTy
+          end,
+    union_var_binds_help(VarBindsList, Lub).
 
 union_var_binds(VB1, VB2, TEnv) ->
     union_var_binds([VB1, VB2], TEnv).
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3441,33 +3441,28 @@ no_guards({clause, _, _, Guards, _}) ->
     Guards == [].
 
 
--spec check_guard_bif(erlang:anno(), atom(), list()) -> map().
-check_guard_bif(P, is_atom, [{var, _, Var}]) -> #{Var => {type, P, atom, []}};
-check_guard_bif(P, is_binary, [{var, _, Var}]) -> #{Var => {type, P, binary, []}};
-check_guard_bif(P, is_bitstring, [{var, _, Var}]) -> #{Var => {type, P, bitstring, []}};
-check_guard_bif(P, is_boolean, [{var, _, Var}]) -> #{Var => {type, P, boolean, []}};
-check_guard_bif(P, is_float, [{var, _, Var}]) -> #{Var => {type, P, float, []}};
-check_guard_bif(P, is_integer, [{var, _, Var}]) -> #{Var => {type, P, integer, []}};
-check_guard_bif(P, is_number, [{var, _, Var}]) -> #{Var => {type, P, number, []}};
-check_guard_bif(P, is_list, [{var, _, Var}]) -> #{Var => {type, P, list, []}};
-check_guard_bif(P, is_map, [{var, _, Var}]) -> #{Var => {type, P, map, []}};
-check_guard_bif(P, is_pid, [{var, _, Var}]) -> #{Var => {type, P, pid, []}};
-check_guard_bif(P, is_port, [{var, _, Var}]) -> #{Var => {type, P, port, []}};
-check_guard_bif(P, is_reference, [{var, _, Var}]) -> #{Var => {type, P, reference, []}};
-check_guard_bif(P, is_tuple, [{var, _, Var}]) -> #{Var => {type, P, tuple, []}};
-check_guard_bif(P, is_function, [{var, _, Var}]) -> #{Var => {type, P, 'fun', []}};
-check_guard_bif(P, is_record, [{var, _, Var}, {atom, _, Record}]) ->
-    #{Var => {type, P, record, [{atom, P, Record}]}};
-check_guard_bif(_P, _Fun, _Vars) ->
-    #{}.
-
 -spec check_guard_call(erlang:anno(), atom(), list()) -> map().
-check_guard_call(P, Name, Args) ->
-    Arity = length(Args),
-    case erl_internal:bif(Name, Arity) of
-        true -> check_guard_bif(P, Name, Args);
-        false -> #{}
-    end.
+check_guard_call(P, is_atom, [{var, _, Var}]) -> #{Var => {type, P, atom, []}};
+check_guard_call(P, is_binary, [{var, _, Var}]) -> #{Var => {type, P, binary, []}};
+check_guard_call(P, is_bitstring, [{var, _, Var}]) -> #{Var => {type, P, bitstring, []}};
+check_guard_call(P, is_boolean, [{var, _, Var}]) -> #{Var => {type, P, boolean, []}};
+check_guard_call(P, is_float, [{var, _, Var}]) -> #{Var => {type, P, float, []}};
+check_guard_call(P, is_function, [{var, _, Var}]) -> #{Var => {type, P, 'fun', []}};
+check_guard_call(P, is_function, [{var, _, Var}, _]) -> #{Var => {type, P, 'fun', []}};
+check_guard_call(P, is_integer, [{var, _, Var}]) -> #{Var => {type, P, integer, []}};
+check_guard_call(P, is_list, [{var, _, Var}]) -> #{Var => {type, P, list, []}};
+check_guard_call(P, is_map, [{var, _, Var}]) -> #{Var => {type, P, map, []}};
+check_guard_call(P, is_number, [{var, _, Var}]) -> #{Var => {type, P, number, []}};
+check_guard_call(P, is_pid, [{var, _, Var}]) -> #{Var => {type, P, pid, []}};
+check_guard_call(P, is_port, [{var, _, Var}]) -> #{Var => {type, P, port, []}};
+check_guard_call(P, is_record, [{var, _, Var}, {atom, _, Record}]) ->
+    #{Var => {type, P, record, [{atom, P, Record}]}};
+check_guard_call(P, is_record, [{var, _, Var}, {atom, _, Record}, _]) ->
+    #{Var => {type, P, record, [{atom, P, Record}]}};
+check_guard_call(P, is_reference, [{var, _, Var}]) -> #{Var => {type, P, reference, []}};
+check_guard_call(P, is_tuple, [{var, _, Var}]) -> #{Var => {type, P, tuple, []}};
+check_guard_call(_P, _Fun, _Vars) ->
+    #{}.
 
 -spec check_guard(#env{}, term()) -> map().
 check_guard(_Env, {call, P, {atom, _, Fun}, Vars}) ->
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3442,27 +3442,24 @@ no_guards({clause, _, _, Guards, _}) ->
 
 
 -spec check_guard_call(atom(), list()) -> map().
-check_guard_call(is_atom, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), atom, []}};
-check_guard_call(is_binary, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), binary, []}};
-check_guard_call(is_bitstring, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), bitstring, []}};
-check_guard_call(is_boolean, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), boolean, []}};
-check_guard_call(is_float, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), float, []}};
-check_guard_call(is_function, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), 'fun', []}};
-check_guard_call(is_function, [{var, _, Var}, _]) -> #{Var => {type, erl_anno:new(0), 'fun', []}};
-check_guard_call(is_integer, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), integer, []}};
-check_guard_call(is_list, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), list, []}};
-check_guard_call(is_map, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), map, []}};
-check_guard_call(is_number, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), number, []}};
-check_guard_call(is_pid, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), pid, []}};
-check_guard_call(is_port, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), port, []}};
-check_guard_call(is_record, [{var, _, Var}, {atom, _, Record}]) ->
-    #{Var => {type, erl_anno:new(0), record, [{atom, erl_anno:new(0), Record}]}};
-check_guard_call(is_record, [{var, _, Var}, {atom, _, Record}, _]) ->
-    #{Var => {type, erl_anno:new(0), record, [{atom, erl_anno:new(0), Record}]}};
-check_guard_call(is_reference, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), reference, []}};
-check_guard_call(is_tuple, [{var, _, Var}]) -> #{Var => {type, erl_anno:new(0), tuple, []}};
-check_guard_call(_Fun, _Vars) ->
-    #{}.
+check_guard_call(is_atom, [{var, _, Var}]) -> #{Var => type(atom)};
+check_guard_call(is_binary, [{var, _, Var}]) -> #{Var => type(binary)};
+check_guard_call(is_bitstring, [{var, _, Var}]) -> #{Var => type(bitstring)};
+check_guard_call(is_boolean, [{var, _, Var}]) -> #{Var => type(boolean)};
+check_guard_call(is_float, [{var, _, Var}]) -> #{Var => type(float)};
+check_guard_call(is_function, [{var, _, Var}]) -> #{Var => type('fun')};
+check_guard_call(is_function, [{var, _, Var}, _]) -> #{Var => type('fun')};
+check_guard_call(is_integer, [{var, _, Var}]) -> #{Var => type(integer)};
+check_guard_call(is_list, [{var, _, Var}]) -> #{Var => type(list)};
+check_guard_call(is_map, [{var, _, Var}]) -> #{Var => type(map)};
+check_guard_call(is_number, [{var, _, Var}]) -> #{Var => type(number)};
+check_guard_call(is_pid, [{var, _, Var}]) -> #{Var => type(pid)};
+check_guard_call(is_port, [{var, _, Var}]) -> #{Var => type(port)};
+check_guard_call(is_record, [{var, _, Var}, {atom, _, Record}]) -> #{Var => type_record(Record)};
+check_guard_call(is_record, [{var, _, Var}, {atom, _, Record}, _]) -> #{Var => type_record(Record)};
+check_guard_call(is_reference, [{var, _, Var}]) -> #{Var => type(reference)};
+check_guard_call(is_tuple, [{var, _, Var}]) -> #{Var => type(tuple)};
+check_guard_call(_Fun, _Vars) -> #{}.
 
 -spec check_guard(#env{}, term()) -> map().
 check_guard(_Env, {call, _, {atom, _, Fun}, Vars}) ->
@@ -3988,6 +3985,9 @@ top() ->
 
 type_var(Name) ->
     {var, erl_anno:new(0), Name}.
+
+type_record(Name) ->
+    {type, erl_anno:new(0), record, [{atom, erl_anno:new(0), Name}]}.
 
 return(X) ->
     { X, #{}, constraints:empty() }.
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3461,29 +3461,29 @@ check_guard_call(is_reference, [{var, _, Var}]) -> #{Var => type(reference)};
 check_guard_call(is_tuple, [{var, _, Var}]) -> #{Var => type(tuple)};
 check_guard_call(_Fun, _Vars) -> #{}.
 
--spec check_guard(#env{}, term()) -> map().
-check_guard(_Env, {call, _, {atom, _, Fun}, Vars}) ->
+-spec check_guard_expression(#env{}, term()) -> map().
+check_guard_expression(_Env, {call, _, {atom, _, Fun}, Vars}) ->
     check_guard_call(Fun, Vars);
-check_guard(_Env, {call, _, {remote,_, {atom, _, erlang},{atom, _, Fun}}, Vars}) ->
+check_guard_expression(_Env, {call, _, {remote,_, {atom, _, erlang},{atom, _, Fun}}, Vars}) ->
     check_guard_call(Fun, Vars);
-check_guard(Env, {op, _OrElseAnno, Op, Call1, Call2}) when Op == 'orelse'; Op == 'or' ->
-    G1 = check_guard(Env, Call1),
-    G2 = check_guard(Env, Call2),
+check_guard_expression(Env, {op, _OrElseAnno, Op, Call1, Call2}) when Op == 'orelse'; Op == 'or' ->
+    G1 = check_guard_expression(Env, Call1),
+    G2 = check_guard_expression(Env, Call2),
     union_var_binds_symmetrical([G1, G2], Env#env.tenv);
-check_guard(Env, {op, _AndAlsoAnno, Op, Call1, Call2}) when Op == 'andalso'; Op == 'and' ->
-    G1 = check_guard(Env, Call1),
-    G2 = check_guard(Env, Call2),
+check_guard_expression(Env, {op, _AndAlsoAnno, Op, Call1, Call2}) when Op == 'andalso'; Op == 'and' ->
+    G1 = check_guard_expression(Env, Call1),
+    G2 = check_guard_expression(Env, Call2),
     union_var_binds([G1, G2], Env#env.tenv);
-check_guard(Env, Guard) ->
+check_guard_expression(Env, Guard) ->
     {_Ty, VB, _Cs} = type_check_expr(Env, Guard), % Do we need to thread the Env?
     VB.
 
 %% The different guards use glb
--spec check_guards_sequence(#env{}, list()) -> map().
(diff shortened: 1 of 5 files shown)


FrankBro:
Fairly happy with this right now. Anything else you guys would want? I feel like it's simpler than the previous PR.

src/typechecker.erl
```diff
@@ -3408,14 +3408,54 @@ refinable(_) ->
...
+    #{Var => {type, P, record, [{atom, P, Record}]}};
+check_guard_bif(_P, _Fun, _Vars) ->
+    #{}.
+
+-spec check_guard_call(erlang:anno(), atom(), list()) -> map().
+check_guard_call(P, Name, Args) ->
+    Arity = length(Args),
+    case erl_internal:bif(Name, Arity) of
+        true -> check_guard_bif(P, Name, Args);
+        false -> #{}
```
josefs: If the call is not to a bif then we should type check it as usual. Call `type_check_expr` here.

gomoripeti: I think that only guard bifs are allowed in guards. So a false would yield a compiler error. Gradualizer could either emit an error as well, or just ignore the false case (and leave it to other tools).

josefs: Ah, good point @gomoripeti! I'm fine leaving the code as it is, but it'd be good to add a comment to document the fact that this case should never happen.

Personally, I would very much prefer if arbitrary functions were allowed in guards, except perhaps in receive guards. But that's a debate for another time.

src/typechecker.erl
```diff
@@ -3408,14 +3408,54 @@ refinable(_) ->
...
+    check_guard_call(P, Fun, Vars);
+check_guard(Env, {op, _OrElseAnno, 'orelse', Call1, Call2}) ->
+    union_var_binds([check_guard(Env, Call1), check_guard(Env, Call2)], Env#env.tenv);
+check_guard(Env, Guard) ->
+    {_Ty, VB, _Cs} = type_check_expr(Env, Guard), % Do we need to thread the Env?
+    VB.
+
 %% TODO: implement proper checking of guards.
+-spec check_guards(#env{}, term()) -> map().
 check_guards(Env, Guards) ->
```
zuiderkwast: The Guards parameter is a guard sequence in abstract format, nothing else, so `term()` is wrong here (because it is currently the top type), but `any()` would be fine.

src/typechecker.erl
```diff
@@ -3408,14 +3408,54 @@ refinable(_) ->
...
     union_var_binds(
       lists:map(fun (GuardSeq) ->
                         union_var_binds(
                           lists:map(fun (Guard) ->
-                                                {_Ty, VB, _Cs} = type_check_expr(Env, Guard), % Do we need to thread the Env?
-                                                VB
+                              check_guard(Env, Guard)
                                     end, GuardSeq), Env#env.tenv)
                 end, Guards), Env#env.tenv).
```
zuiderkwast: There the nested semicolon-separated list of comma-separated guards is traversed. I can't see that the or/and logic for guards is handled here. I'm just worried about OR-guard sequences like `when is_integer(X); is_atom(X)` and `when is_integer(X); is_integer(Y)`. In the first case, it's possible to refine X to `integer() | atom()` but in the second case we can't refine X nor Y. In cases where we can't be sure, its better to not refine at all IMO, so perhaps you should limit the refinement to guard sequences without semicolons (i.e. singleton list on the outer level) for now?

@NelsonVides made #206 and it had the same issue initially. Don't know if it was fixed later.

FrankBro:
Will try to implement all of these asap.

src/typechecker.erl
NelsonVides: I think this is what I tried implementing here: https://github.com/josefs/Gradualizer/blob/691a3bc4d15ef877a16b6c546290c22d398a2199/src/typechecker.erl#L3431-L3450

src/typechecker.erl
FrankBro: I did not know only bifs are allowed in guards. I took a look at all the available ones (Seen here, not sure if up to date https://en.wikibooks.org/wiki/Erlang_Programming/guards#Guard_functions). Might try to expand this to support more of them.

FrankBro: This is probably a better list: http://erlang.org/doc/reference_manual/expressions.html#guard-sequences

src/typechecker.erl
FrankBro: I did some tests with this. I've added a comment at the end for questions on how to support `andalso` and guard sequences via comma.

src/typechecker.erl
FrankBro: Yea thanks for that. Used some of your logic in my update.

FrankBro:
One last thing I'm uncertain how to support is `andalso` and commas, also known as guard sequence.

I am kind of assuming they are similar: for example:

```erlang
test_andalso(N) when is_integer(N) andalso is_number(N) -> N;
test_andalso(_) -> not_integer.

test_seq(N) when is_integer(N), is_number(N) -> N;
test_seq(_) -> not_integer.
```

Technically, both of these cases should end up being `integer() | not_integer`. I guess using glb on `integer() | number()` would get rid of number at least.

FrankBro:
Added `andalso` support but for some reason, the failing test case passes. Not sure why it thinks that `number()` is valid if it must also pass `integer()`.

FrankBro:
@josefs It seems like the latest change that treats `term()` like `any()` as broken my tests that were supposed to fail. For example, this test would fail before:

```erlang
-spec test(term()) -> integer() | not_integer.
test(I) when is_atom(I) -> I;
test(_) -> not_integer.
```

I feel like this a valid code and spec example that should not pass. To make this fail with the current code, all I had to do was comment this: https://github.com/josefs/Gradualizer/commit/eac33e8a5de8d4ffc5fe57ff60dae4d209e48e83#diff-89cf8fff9d52a7e6459033763aaacb3aR146-R149

FrankBro:
> Can I ask you to update the wiki also? The page below has some text which explains when type refinement happens and this PR extends that. We want to make sure the documentation is correct.
> https://github.com/josefs/Gradualizer/wiki/Type-refinement

Seems like it's not possible to PR the wiki. Once the changes are done and everything is working, I'll gladly send an edited version.

josefs:
@FrankBro, my last change did indeed change the semantics of `term()`. We decided to make this change a while ago for a number of reasons. The type `term()` now represents the dynamic type and your test should pass. If you want to test the old behavior you should use the new type `gradualizer:top()` which behaves like the old `term()`.

As for the wiki, you just edit it in the browser. There are no PRs. I've added you as a collaborator on the project so you can now make edits to the wiki.

src/typechecker.erl
```diff
@@ -3438,16 +3438,76 @@ refinable(_) ->
...
+check_guard_call(P, is_reference, [{var, _, Var}]) -> #{Var => {type, P, reference, []}};
+check_guard_call(P, is_tuple, [{var, _, Var}]) -> #{Var => {type, P, tuple, []}};
+check_guard_call(_P, _Fun, _Vars) ->
+    #{}.
+
+-spec check_guard(#env{}, term()) -> map().
+check_guard(_Env, {call, P, {atom, _, Fun}, Vars}) ->
+    check_guard_call(P, Fun, Vars);
+check_guard(_Env, {call, P, {remote,_,_,{atom, _, Fun}}, Vars}) ->
+    check_guard_call(P, Fun, Vars);
```
zuiderkwast: I assume the module must be `erlang` here. Better match it explicitly. We shouldn't accept `arbitrary_module:is_atom(X)`.

src/typechecker.erl
```diff
@@ -3438,16 +3438,76 @@ refinable(_) ->
 no_guards({clause, _, _, Guards, _}) ->
     Guards == [].
 
+
+-spec check_guard_call(erlang:anno(), atom(), list()) -> map().
+check_guard_call(P, is_atom, [{var, _, Var}]) -> #{Var => {type, P, atom, []}};
+check_guard_call(P, is_binary, [{var, _, Var}]) -> #{Var => {type, P, binary, []}};
```
zuiderkwast: These are using the annotation of the function call for the created type, which we don't usually do. There are helper functions `type/1,2` to create types with an empty annotation, e.g. `type(atom)`.

src/typechecker.erl
```diff
@@ -3438,16 +3438,76 @@ refinable(_) ->
...
-                end, Guards), Env#env.tenv).
+    Lub = fun(_K, Ty1, Ty2) ->
+        UTy = {type, erl_anno:new(0), union, [Ty1, Ty2]},
+        NTy = normalize(UTy, Env#env.tenv),
+        NTy
+    end,
+    Tys = lists:map(fun (GuardSeq) ->
+        check_guards_sequence(Env, GuardSeq)
+    end, Guards),
+    Ty = union_var_binds_help(Tys, Lub),
```
zuiderkwast: This line and the definition of Lub is the same in the code for `orelse`, so I think the common code could be reused, perhaps wrapped in a function named something like `union_var_binds_using_lub`?

Also, the variable name `Ty` and `Tys` here are confusing, because these are not types. They are var binds (maps) and list of such.

----

I'm not convinced though that `union_var_binds_help(Tys, Lub)` correctly merges `X :: integer()` with `Y :: integer()` into the empty set of var binds. (`union_var_binds_help` calls `gradualizer_lib:merge_with/3` which is using `maps:update_with/4`, which if I'm reading correctly would return `X :: integer(), Y :: integer()` in this case.)

Thus, I think `union_var_binds_using_lub` (or with a better name) needs to be implemented differently, so that only variables which are defined in all var bind maps are merged using Lub. Otherwise, they should be omitted.

To verify this, I'd like to see a *should fail* test case like this:

```Erlang
-spec f(integer() | atom(), integer() | atom()) -> integer().
f(X, Y) when is_integer(X); is_integer(Y) -> X + Y;
f(_, _) -> 42.
```

A similar guard which should prevent refinement is P in the following:

```Erlang
-spec g(integer() | atom(), boolean()) -> integer().
g(X, P) when is_integer(X); P -> X.
```

zuiderkwast:
> @josefs It seems like the latest change that treats `term()` like `any()` as broken my tests that were supposed to fail. For example, this test would fail before:
> 
> ```erlang
> -spec test(term()) -> integer() | not_integer.
> test(I) when is_atom(I) -> I;
> test(_) -> not_integer.
> ```
> 
> I feel like this a valid code and spec example that should not pass. To make this fail with the current code, all I had to do was comment this: [eac33e8#diff-89cf8fff9d52a7e6459033763aaacb3aR146-R149](https://github.com/josefs/Gradualizer/commit/eac33e8a5de8d4ffc5fe57ff60dae4d209e48e83#diff-89cf8fff9d52a7e6459033763aaacb3aR146-R149)

The reason that this passes is that `glb(term(), atom()) -> term()` in our type system. We can get higher precision if we adopt the semantic subtyping combined with gradual typing as described in https://www.irif.fr/~gc/papers/icfp17.pdf, and then we can probably detect that there is no overlap between the types `atom() & any()` and `integer() | not_integer` and thus report a type error here. See #232.

FrankBro:
> @FrankBro, my last change did indeed change the semantics of `term()`. We decided to make this change a while ago for a number of reasons. The type `term()` now represents the dynamic type and your test should pass. If you want to test the old behavior you should use the new type `gradualizer:top()` which behaves like the old `term()`.
> 
> As for the wiki, you just edit it in the browser. There are no PRs. I've added you as a collaborator on the project so you can now make edits to the wiki.

So users would have to put `gradualizer:top()` instead of `term()`? I don't think it's going to work for a lot of external libraries. Perhaps we should expose an option to support the old behavior?

src/typechecker.erl
FrankBro: So apparently, you can only have a few BIFs as guards. So there is no need to check for this because otherwise, erlang would complain at compilation and it wouldn't work. Hence why I removed the BIF check. http://erlang.org/doc/reference_manual/expressions.html#guard-sequences

josefs:
> So users would have to put gradualizer:top() instead of term()? I don't think it's going to work for a lot of external libraries. Perhaps we should expose an option to support the old behavior?

The reason we decided to change the semantics of `term()` is that the old semantics, the top of the subtyping hierarchy, is a rather useless type. There aren't really any practical uses for it. That's why we're relegated the top of the subtyping hierarchy to the rather obscure type `gradualizer:top()`. 

> Isn't the point of Gradualizer to be a drop-in replacement for Dialyzer which is pretty bad and let way too many things go through?

Gradualizer and dialyzer tries to achieve different things. Dialyzer tries to prove the presence of bugs while gradualizer is a more conventional type system and tries to prove the absense of a certain restricted set of bugs. Gradualizer doesn't supersede dialyzer. Dialyzer can do useful things that gradualizer will never be able to do. They are complementary.

FrankBro:
Not sure what is the expected behaviour for this.

```erlang
-spec wrong_andalso(gradualizer:top()) -> atom() | not_integer.
wrong_andalso(N) when is_integer(N) andalso is_atom(N) -> N;
wrong_andalso(_) -> not_integer.
```

I would expect an error, considering something cannot be both an integer and an atom. Using `union_var_binds` and `glb`, this reduces it to `none`. Should it be more aggressive with reporting errors?

zuiderkwast:
No, we don't expect people to use `gradualizer:top()`.

The problem with using the top type (or `term()` until recently) as a result type of various functions is that when this is later passed to a function such as `integer_to_list/1`, it's a type error.

```Erlang
my_counter_as_string() ->
    case get_value(my_counter) of
        undefined -> "Nothing";
        N         -> integer_to_list(N)
    end.

-spec get_value(Key :: term()) -> Value :: term() | undefined.
get_value(...) -> ...
```

Many times, the programmer "knows" that e.g. N above is an integer for some reason, but the type checker doesn't. The type checker cannot guarantee that it's safe to call `integer_to_list` with this value and must therefore report a type error.

With the unknown type AKA the dynamic type though, this is not statically checked, so no type error is reported at compile time. This is what the programmer wants in situations where an exact enough type can't be specified.

If you want statically checked key-value storage, you would need to use a type parameterized with the key and value types, such as `dict:dict(Key, Value)` and the lookup function would need to have a type such as `-spec fetch(Key, Dict) -> Value when Dict :: dict(Key, Value)`. Map types can be specified in a very exact way.

Unfortunately, polymorphic types can't be used for ETS tables, message passing, calls to gen servers and in a lot of other situations, because there is no way to write types for these situations.

With refinement, we can do something, but we can't refine the unknown type to `integer()` using a guard like `is_integer(N)`. (I've explained why in another comment.)

FrankBro:
@zuiderkwast I think your example actually shows that the developer is wrong in this situation and he should change the code to remove assumptions. `my_counter_as_string()` should've been implemented like this:

```erlang
my_counter_as_string() ->
    case get_value(my_counter) of
        N when is_integer(N) -> integer_to_list(N);
        _ -> "Nothing"
    end.
```

Not sure why trying to accept assumptions that might become wrong at any time is a good idea. I know there is a lot of erlang that is impossible to typecheck, such as message passing, especially considering you can upgrade live code and change what true at compile time at some point.

Then why not treat these possibilities as always requiring a `term()` path as well?

I worked on a row polymorphism toy compiler (https://github.com/FrankBro/ordo) and one thing I thought was nice is the distinction of matching with or without an underscore case. Why can't messages be forced to have a type ending with `term()` if that limitation is forced on us by erlang?

```erlang
-type message() :: {message_a, integer()} | {message_b, atom()} | term().
```

src/typechecker.erl
zuiderkwast: You're assuming that we only want to type check modules which compile.

It's fine with me, although we have lots code for checking code which doesn't compile. 

Anyway, withing `{atom, _, erlang}` here instead of `_` would make it easier to understand the code IMO.

FrankBro:
Should I just change all my tests to use more specific tests then? So

```erlang
-spec test_is_atom(term()) -> atom() | not_atom.
test_is_atom(A) when is_atom(A) -> A;
test_is_atom(_) -> not_atom.
```

To

```erlang
-spec test_is_atom(atom() | integer()) -> atom() | not_atom.
test_is_atom(A) when is_atom(A) -> A;
test_is_atom(_) -> not_atom.
```

zuiderkwast:
> @zuiderkwast I think your example actually shows that the developer is wrong in this situation and he should change the code to remove assumptions. `my_counter_as_string()` should've been implemented like this:
> 
> ```erlang
> my_counter_as_string() ->
>     case get_value(my_counter) of
>         N when is_integer(N) -> integer_to_list(N);
>         _ -> "Nothing"
>     end.
> ```

The programmer may know that the counter is always an integer, because all the functions for setting the value sets it to an integer and he understands the program. Thus, he may think that the extra guard is a waste of effort.

Either way, the `is_integer` guard doesn't change anything if we don't refine `term()` to `integer()`, and we don't because the dynamic type of N here may be known to be an even more specific type, such as an even number or a positive number or a prime. Refining unknown to `integer()` will cause a type error to be reported if e.g. `lists:nth(N, SomeList)` is called.

A gradual type system reports type errors for statically typed code and doesn't complain about dynamically typed, to allow mixing of the two styles.

> Then why not treat these possibilities as always requiring a `term()` path as well?
> 
> I worked on a row polymorphism toy compiler (https://github.com/FrankBro/ordo) and one thing I thought was nice is the distinction of matching with or without an underscore case. Why can't messages be forced to have a type ending with `term()` if that limitation is forced on us by erlang?
> 
> ```erlang
> -type message() :: {message_a, integer()} | {message_b, atom()} | term().
> ```

I don't understand when you want to force `... | term()`. What does "these possibilities" refer to?

You can certainly write a type like that. Intuitively, it means "message a or message b or possibly but not certainly something else".

FrankBro:
>Thus, he may think that the extra guard is a waste of effort.

And he'd be wrong

>Either way, the is_integer guard doesn't change anything if we don't refine term() to integer()

Indeed, I was discussing if we were to try

>I don't understand when you want to force ... | term(). What does "these possibilities" refer to?

Fairly new to erlang and gradual type systems. Maybe trying to do too much and it doesn't actually apply. A lot of my blabbering is just me thinking out loud :)

zuiderkwast:
@FrankBro 

>> Thus, he may think that the extra guard is a waste of effort.
>
> And he'd be wrong

Who are we to judge? It's an example. Isn't this reasoning very similar to where I suggested you should match explicitly on `erlang` as the module in the remote call in a guard, while you thought it's superfluos because if it passes the compiler, then it has to be a call to the erlang module.

That's the kind of knowledge the programmer has that the type checker doesn't.

> A lot of my blabbering is just me thinking out loud :)

Fair enough! :smiley:

As I mentioned earlier, I think it would be very nice if we can refine any() to "any() & integer()" intuitively meaning "some unknown subtype of integer()" (#232). The theory and also how to implement it efficiently is all in the paper.

zuiderkwast:
> Not sure what is the expected behaviour for this.
> 
> ```erlang
> -spec wrong_andalso(gradualizer:top()) -> atom() | not_integer.
> wrong_andalso(N) when is_integer(N) andalso is_atom(N) -> N;
> wrong_andalso(_) -> not_integer.
> ```
> 
> I would expect an error, considering something cannot be both an integer and an atom. Using `union_var_binds` and `glb`, this reduces it to `none`. Should it be more aggressive with reporting errors?

Interesting! `glb/3` probably returns `none()`, which is a subtype of the expected return value, thus no type error. This is sound, so it's fine for now.

The problem with this clause though is that it will never match, obviously. We could report a "can never match" error if any type is refined to `none()`. Let's make it a separate improvement?

src/typelib.erl
```diff
@@ -91,8 +91,8 @@ remove_pos({Type, _, Value})
 remove_pos({user_type, Anno, Name, Params}) when is_list(Params) ->
     {user_type, anno_keep_only_filename(Anno), Name,
      lists:map(fun remove_pos/1, Params)};
-remove_pos({type, Anno, record, Params = [_Name]}) ->
-    {type, anno_keep_only_filename(Anno), record, Params};
+remove_pos({type, Anno, record, Params = [{atom, AtomAnno, Name}]}) ->
+    {type, anno_keep_only_filename(Anno), record, [{atom, anno_keep_only_filename(AtomAnno), Name}]};
```
FrankBro: Omg we were not fixing the anno for the record's atom name @_@ lol.

FrankBro:
Nice, everything passes again. Not sure if I wanna keep the `andalso` tests in this PR. Anything else you guys would want?

src/typechecker.erl
FrankBro: Can we close this? Considering we're only looking for function calls and only specific BIFs are allowed.

FrankBro:
I think a better name for this PR and branch would be `guard-BIF-support`. I've supported every functions listed here: http://erlang.org/doc/reference_manual/expressions.html#guard-sequences

If we read the length of everything supported, it mentions:

- ~The atom true~ Useless?
- ~Other constants (terms and bound variables), all regarded as false~ Useless as well? Will just never be hit
- ~Calls to the BIFs specified in table Type Test BIFs~ This PR fixes this.
- Term comparisons
- Arithmetic expressions
- Boolean expressions
- ~Short-circuit expressions (andalso/orelse)~ This PR handles this mostly, maybe need more testing

The arithmetic stuff are not handle and seems more complicated a bit. I haven't entered the constraint part of gradualizer but we only extract variable bindings from `check_guards`. If we wanted to support things like

```erlang
-spec is_zero(integer()) -> no | 0.
is_zero(N) when N == 0 -> N;
is_zero(_) -> no.
```

We would constraints I think. Sounds like another PR to me ;)

src/typechecker.erl
FrankBro: Can we resolve this? I have some `orelse` and `andalso` tests.

src/typechecker.erl
FrankBro: Can we resolve this? Handled both `orelse` and `andalso` in tests.

src/typechecker.erl
FrankBro: Do we still need to support the `remote` clause if we use erlang there?

src/typechecker.erl
FrankBro: Fixed the first part, thanks for the feedback. Looking into the second part now.

FrankBro: To make things clear, `;` is `orelse`.  I will add test cases for both to make sure everything works.

FrankBro: Thanks for that man, good finding. Managed to make sure both `orelse` and `andalso` correctly bind stuff. Check the new test cases. Resolving this.

FrankBro:
I guess codecov doesn't include the `should_pass`/`should_fail` folder? :(

src/typelib.erl
zuiderkwast: Good. :-)

src/typechecker.erl
```diff
@@ -3438,16 +3438,63 @@ refinable(_) ->
...
 
+
+-spec check_guard_call(atom(), list()) -> map().
+check_guard_call(is_atom, [{var, _, Var}]) -> #{Var => type(atom)};
+check_guard_call(is_binary, [{var, _, Var}]) -> #{Var => type(binary)};
+check_guard_call(is_bitstring, [{var, _, Var}]) -> #{Var => type(bitstring)};
+check_guard_call(is_boolean, [{var, _, Var}]) -> #{Var => type(boolean)};
+check_guard_call(is_float, [{var, _, Var}]) -> #{Var => type(float)};
+check_guard_call(is_function, [{var, _, Var}]) -> #{Var => type('fun')};
+check_guard_call(is_function, [{var, _, Var}, _]) -> #{Var => type('fun')};
```
josefs: We can actually return a more precise type here if the second argument to `is_function` is a literal number. We would then know the arity of the function. For example, if we encountered `is_function(F,2)` we can give `F` the type  `fun(any(), any()) -> any()`.

It's fine if you don't want to make this improvement in this PR, but I think it's a natural thing to include.

src/typechecker.erl
```diff
@@ -3933,6 +3983,45 @@ is_power_of_two(N) when N rem 2 == 0 ->
...
+    #{};
+union_var_binds_symmetrical(VarBindsList, TEnv) ->
+    Lub = fun(_K, Ty1, Ty2) ->
+        UTy = type(union, [Ty1, Ty2]),
+        NTy = normalize(UTy, TEnv),
+        NTy
+    end,
+    union_var_binds_symmetrical_help(VarBindsList, Lub).
+
+-spec union_var_binds_symmetrical_help(list(), function()) -> map().
```
josefs: Please use `union_var_binds_help` here instead of defining a new function. You can just pass the `LUB` function to `union_var_binds_help`.

zuiderkwast: No. This was done initially, but it doesn't work. See my comment on previous patch.

src/typechecker.erl
```diff
@@ -3438,16 +3438,63 @@ refinable(_) ->
...
+
+%% The different guards use glb
+-spec check_guards_sequence(#env{}, list()) -> map().
+check_guards_sequence(Env, GuardSeq) ->
+    RefTys = union_var_binds(
+        lists:map(fun (Guard) ->
+            check_guard(Env, Guard)
+                  end, GuardSeq),
+        Env#env.tenv),
+    maps:merge(Env#env.venv, RefTys).
```
zuiderkwast: This should be the same as for andalso, i.e. `union_var_binds`. Otherwise the last guard overwrites the types refined by previous guards, instead of combining them.

Btw, the terminology according to erlang docs:

* Guard sequences ::= semicolon-separated list of guards
* Guard ::= comma-separated list of guard expressions
* Guard Expression ::= Expression, with limitations

src/typechecker.erl
josefs: The pattern should be `{remote, _, {atom, _, erlang}, {atom, _, Fun}}`.

zuiderkwast: If `erlang:is_atom(X)` is allowed in guards, we should better support it, because it would be represented as `remote` in the AST.

zuiderkwast: It's not very important though, but since it's already implemented, better keep it.

src/typechecker.erl
josefs: I can't find the comment you're referring to. How is this function different from `union_var_binds_help`?

zuiderkwast: https://github.com/josefs/Gradualizer/pull/241#discussion_r363033009

zuiderkwast:
> I think a better name for this PR and branch would be `guard-BIF-support`. I've supported every functions listed here: http://erlang.org/doc/reference_manual/expressions.html#guard-sequences

Makes sense.

> If we read the length of everything supported, it mentions:
> 
>     * ~The atom true~ Useless?

It may be a constant defined to true using conditional preprocessor code, which can be useful in some cases.

>     * ~Other constants (terms and bound variables), all regarded as false~ Useless as well? Will just never be hit

Variables bound to true are regarded as true. Literal false may be from constants defined using compiler options, which is not useless.

In gradualizer, we sometimes have a guard like `when Env#env.infer` which is an expression which evaluates to true or false.

> The arithmetic stuff are not handle and seems more complicated a bit. I haven't entered the constraint part of gradualizer but we only extract variable bindings from `check_guards`. If we wanted to support things like
> 
> ```erlang
> -spec is_zero(integer()) -> no | 0.
> is_zero(N) when N == 0 -> N;
> is_zero(_) -> no.
> ```
> 
> We would constraints I think. Sounds like another PR to me ;)

Another PR definitely.

Since there is a total ordering of terms in Erlang, I think a guard like `X > 2` can be turned into a type, although not always an exact type. With the combination `X > 2, X < 10`, we could be able to merge these types into `X :: 3..9`.

I think constraints solving is quite difficult, considering the *future work* section in the paper from 2019, where they combine gradual typing with polymorphism.

src/typechecker.erl
FrankBro: Nice. Yea I wasn't sure how to do this so I didn't implement it. Will do.

src/typechecker.erl
FrankBro: Basically, instead of merging all var binds, we merge them when they are present in every var binds.

```
union_var_binds : [{ x: integer }, {x: integer, y: integer}] = {x: integer, y: integer}
union_var_binds_symmetrical : [{ x: integer }, {x: integer, y: integer}] = {x: integer}
```

src/typechecker.erl
zuiderkwast: We can resolve this, because now `term()` is the dynamic type. The top type would be wrong here. That's what my comment was about. I don't know why you were talking about andalso and orelse here.

src/typechecker.erl
zuiderkwast: Yeap, good.

src/typechecker.erl
josefs: I see. That makes sense.

FrankBro:
Quick question in relation to the anno stuff. I used to put the anno of the expression in the type but was asked to change it to `erl_anno:new(0)` because it wasn't actually the anno of the type. If I check this link, https://github.com/zuiderkwast/erlang_abstract_format#expressions, all the annos in types seem to point to where the expression would be. Am I understanding correctly?

zuiderkwast:
> Quick question in relation to the anno stuff. I used to put the anno of the expression in the type but was asked to change it to `erl_anno:new(0)` because it wasn't actually the anno of the type. If I check this link, https://github.com/zuiderkwast/erlang_abstract_format#expressions, all the annos in types seem to point to where the expression would be. Am I understanding correctly?

The anno of the type points to where the type is written in the code, using -type or -spec. (A type definition is not an expression.)

FrankBro:
Also, is there a difference between `{type, _, any}` and `{type, _, any, []}`? It seems the former is only used for functions but I seem to get a lot of `INVALID-FORM` when I use it.

zuiderkwast:
> Also, is there a difference between `{type, _, any}` and `{type, _, any, []}`? It seems the former is only used for functions but I seem to get a lot of `INVALID-FORM` when I use it.

I had to search for `"any}"` on the page https://github.com/zuiderkwast/erlang_abstract_format to find out:

* `{type, _, any}` is the `...` in types like `fun((...) -> integer())` (function of any arity)
* `{type, _, map, any}` is `map()`
* `{type, _, tuple, any}` is `tuple()`

`any()` is represented as `{type, _, any, []}`.

FrankBro:
I added support for `is_function` with arity but the tests that should fail don't seem to. I'm wondering if it's because of all the `any()` types basically skipping everything. It should at least check arity, right?

src/typechecker.erl
FrankBro: Can we come up with a test case that would fail but need to pass? Would love to confirm I fix it.

FrankBro:
Wiki I would change. 

When can type refinement happen?
Type refinement can only be applied when the type checker knows exactly why a clause mismatches and those values can be excluded from the types. That means that we can refine types that consists of unions of singleton types. Singleton types are single integers, atoms, the empty list, the empty tuple and tuples consisting of singletons. Empty maps and empty binaries are singleton types too, but currently Gradualizer is not capable of using these for refinement. Free variables and _ (match all) can also be part of pattern. We know that these can never cause a pattern to mismatch.
***Type refinement can also occur for some guards. For example, all the BIFs guards are supported as well as `andalso`/`and`/`,` and `orelse`/`or`/`;` . [The full list is show here](http://erlang.org/doc/reference_manual/expressions.html#guard-sequences)***

What prevents type refinement?
Guards. If there is any uncertainty to why a clause mismatches, such as a guard **that use term comparison**, we can not exclude any values. We cannot exclude the possibility that the clause mismatches just because of the guard. Currently, Graudalizer can only do type refinement for clauses without guards **or clauses using BIF guards**.
%% This gives a type error
-spec guard_prevents_refinement(apa | bepa) -> bepa.
guard_prevents_refinement(X) ->
    Y = receive Message -> Message end,
    case X of
        apa when Y /= yellow -> bepa;
        Other                -> Other  %% Here, Gradualizer cannot exclude apa
    end.

zuiderkwast:
The wiki change ideas look fine, but I'd like to suggest a restructuring of the page. The old stuff is all about refining when a clause *mismatches*, which is maybe not the first thing you'd think of, so why not put your stuff first? How about sections like these:

1. Type refinement based on guards, for the body of the matching clause (the new stuff)
2. Type refinement based on mismatching pattern, for the following clauses (most of the old content, including when it can happen and what prevents it)

Feel free to edit the wiki multiple times. It's just W.I.P. documentation anyway.

src/typelib.erl
```diff
@@ -104,6 +104,10 @@ remove_pos({type, _, 'fun', [{type, _, any}, RetTy]}) ->
     %% the only place where `{type, _, any}` can occure
     {type, erl_anno:new(0), 'fun', [{type, erl_anno:new(0), any}
                                    ,remove_pos(RetTy)]};
+remove_pos({type, _, 'fun', [{type, _, product, Args}, RetTy]}) ->
+    {type, erl_anno:new(0), 'fun', [
+        {type, erl_anno:new(0), product, [remove_pos(Arg) || Arg <- Args]}
+        , remove_pos(RetTy)]};
```
zuiderkwast: The reason this wasn't here before is that `fun()` is normalized to `fun((...) -> any())`, which is equivallent. @gomoripeti did this some time ago. Maybe there are more places where `fun()` is not handled because it's expected to be represented as `fun((...) -> any())`?

I don't know, but if we only create new types on the form returned by normalize/2, we wouldn't need to add new clauses such as this.

zuiderkwast: Sorry, my comment was wrong, so I made it hidden. I didn't read the code carefully.

I think this case should be covered by the function clause below: `remove_pos({type, _, Type, Params})`. Type would match 'fun' and then 'product' when called recursively. Doesn't it?

src/typechecker.erl
zuiderkwast:
```Erlang
-spec f(number()) -> integer().
f(N) when is_integer(N), is_number(N) -> N;
f(_) -> 42.
```

FrankBro: The test passes as it is. I put some printing to make sure it didn't end up being `type(none)` but it was `type(integer)` in the end.

src/typelib.erl
FrankBro: Hum, I commented it and everything passes. I'm wondering if this is related to the fact I can't seem to get different arity functions with `is_function/2` tests to fail properly. I know I added this code during that time.

FrankBro: I will remove it since looking at it, I'm pretty sure you're right.

src/typechecker.erl
zuiderkwast: Ah... My bad. I see now that `union_var_binds` is used to combine the refined types, just like for `andalso`. Afterwards, `maps:merge/2` is used just to add the refined var binds to then environment. I guess it's fine to let the refined types overwrite the original types, so I was wrong here.

zuiderkwast: Just repeating the terminology, because I think using the proper official terminology is crucial for understanding and maintaining the code.

A comma-separated list of guard expressions is called "guard", so I think the function "check_guards_sequence" should be renamed to "check_guard".

The function "check_guard" should better be renamed to "check_guard_expression".

Also, the variable names used within these functions should better be renamed to reflect this terminology.

(A semicolon-separated list of *guards*  is called "guard sequence", which is handled by "check_guards". The name "check_guards" is fine, I think, but the variable GuardSeq used inside it should be renamed to Guard.)

src/typechecker.erl
```diff
@@ -3438,16 +3438,71 @@ refinable(_) ->
...
+check_guard(_Env, {call, _, {remote,_,_,{atom, _, Fun}}, Vars}) ->
+    check_guard_call(Fun, Vars);
+check_guard(Env, {op, _OrElseAnno, 'orelse', Call1, Call2}) ->
+    G1 = check_guard(Env, Call1),
+    G2 = check_guard(Env, Call2),
+    union_var_binds_symmetrical([G1, G2], Env#env.tenv);
+check_guard(Env, {op, _OrAnno, 'or', Call1, Call2}) ->
+    G1 = check_guard(Env, Call1),
+    G2 = check_guard(Env, Call2),
+    union_var_binds_symmetrical([G1, G2], Env#env.tenv);
```
zuiderkwast: This duplicated code can easily be combined into

```Erlang
check_guard(Env, {op, _Anno, Op, Call1, Call2}) when Op == 'orelse';
                                                     Op == 'or' ->
    G1 = check_guard(Env, Call1),
    G2 = check_guard(Env, Call2),
    union_var_binds_symmetrical([G1, G2], Env#env.tenv);
```

and similarily for 'and' and 'andalso'.

src/typechecker.erl
```diff
@@ -3438,16 +3438,71 @@ refinable(_) ->
...
+check_guard_call(is_record, [{var, _, Var}, {atom, _, Record}]) -> #{Var => type_record(Record)};
+check_guard_call(is_record, [{var, _, Var}, {atom, _, Record}, _]) -> #{Var => type_record(Record)};
+check_guard_call(is_reference, [{var, _, Var}]) -> #{Var => type(reference)};
+check_guard_call(is_tuple, [{var, _, Var}]) -> #{Var => type(tuple)};
+check_guard_call(_Fun, _Vars) -> #{}.
+
+-spec check_guard(#env{}, term()) -> map().
+check_guard(_Env, {call, _, {atom, _, Fun}, Vars}) ->
+    check_guard_call(Fun, Vars);
+check_guard(_Env, {call, _, {remote,_,_,{atom, _, Fun}}, Vars}) ->
```
zuiderkwast: Please change the pattern to `{call, _, {remote, _, {atom, _, erlang}, {atom, _, Fun}}`. Then I think this PR can be submitted.

src/typechecker.erl

src/typechecker.erl

src/typechecker.erl
FrankBro: I agree the naming convention should be respected. I will change those names.

FrankBro: Renamed. Thank you.

src/typechecker.erl
FrankBro: Done with caveats. Added a known_issue because I can't seem to get tests to fail but I emit the proper types.

src/typechecker.erl

FrankBro:
Anything else this PR needs? Not sure I have anything else to do.

FrankBro:
I like @zuiderkwast suggestion for the wiki. Both this PR and my other one could be in their own section. I suggest we open another issue and put in there what we feel it would look like to unblock the two PRs so they can be merged.

src/typechecker.erl
```diff
@@ -3921,6 +3968,15 @@ top() ->
 type_var(Name) ->
     {var, erl_anno:new(0), Name}.
 
+type_record(Name) ->
+    {type, erl_anno:new(0), record, [{atom, erl_anno:new(0), Name}]}.
+
+type_fun(0) ->
```
josefs: There is no need for this special case. `lists:seq(1,0)` returns `[]`.

FrankBro: Fixed

josefs:
Great work @FrankBro! Feel free to open a separate issue to track the wiki changes.

FrankBro:
Ouch, should've squashed 😅 My next PR will also affect the wiki so I will wait for it to be complete as well.

zuiderkwast:
Why wait? Incremental updates to the wiki is better IMO.

PR #244, opened 2019-12-30
Let term() also be the dynamic type, just like any()

### LINKED ISSUE #217: Treat term() as the dynamic type
The discussion in https://github.com/josefs/Gradualizer/pull/205 cumulated in a decision to treat `term()` as the dynamic type, i.e. as `any()`.

@josefs writes
> Cool! It seems that we've reached an agreement on treating `term()` as the dynamic type as well.
> 
> We still need an internal representation for the top of the subtyping hierarchy, for example in `glb`. But that's an implementation detail for whoever implements this change.

josefs:
Closes #217. See the discussion in #205 for the reasons for this
change.

I've introduced the new type gradualizer:top() which now represents
the top of the subtyping hierarchy. Few programmers should ever need
to use this type but it's important in a couple of places in the
typechecker.


src/gradualizer.erl
```diff
@@ -39,10 +39,13 @@
          type_check_forms/2
         ]).
 
--export_type([options/0]).
+-export_type([options/0, top/0]).
 
 -type options() :: proplists:proplist().
 
+%% This type is the top of the subtyping lattice.
+-type top() :: any().
```
zuiderkwast: Could be -opaque or defined explicitly as a union of all types

src/typechecker.erl
```diff
@@ -136,14 +138,20 @@ compat(T1, T2, A, TEnv) ->
             compat_ty(Ty1, Ty2, sets:add_element({Ty1, Ty2}, A), TEnv)
     end.
 
-%% any() is used as the unknown type in the gradual type system
+%% any() and term() are used as the unknown type in the gradual type system
```
zuiderkwast: If we normalize term to any, we don't need to handle term here and in some other places.

src/gradualizer.erl
josefs: Thanks! Using opaque makes sense so I've used that. Defining it as a union of all types doesn't seem to be very practical, as I don't think that's possible without resorting to using `any()` in lots of places, which defeats the purpose.

src/typechecker.erl
josefs: That's a good point. We should start moving to a point where the internal representation we have for types is different from the type representation used in the absform. Normalizing away `term()` in favor of `any()` is one such thing, and we'll also want to use a more efficient representation of unions to solve the efficiency problems with have with it.
I'll keep the code as it is for now, and we can clean it up as part of being more stringent about the separation between our internal representation and the absform representation.

Issue #248, opened 2020-01-09
Crash in :typechecker.unary_op_arg_type/2


lukaszsamson:
Gradualizer crashes on latest master (862356582afb4f0e8377111125122c5819c0a7d9) when run on https://github.com/elixir-lsp/elixir_sense project.

Erlang/OTP 22 [erts-10.6] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
Elixir 1.9.4 (compiled with Erlang/OTP 22)

```
** (FunctionClauseError) no function clause matching in :typechecker.unary_op_arg_type/2

The following arguments were given to :typechecker.unary_op_arg_type/2:

    # 1
    :not

    # 2
    {:type, 0, :term, []}

    (gradualizer) /Users/elixir_sense/deps/gradualizer/src/typechecker.erl:2771: :typechecker.unary_op_arg_type/2
    (gradualizer) /Users/elixir_sense/deps/gradualizer/src/typechecker.erl:2764: :typechecker.type_check_unary_op_in/5
    (gradualizer) /Users/elixir_sense/deps/gradualizer/src/typechecker.erl:2095: :typechecker.type_check_expr_in/3
    (gradualizer) /Users/elixir_sense/deps/gradualizer/src/typechecker.erl:3268: :typechecker.check_clause/4
    (gradualizer) /Users/elixir_sense/deps/gradualizer/src/typechecker.erl:3230: anonymous fn/4 in :typechecker.check_clauses/4
    (stdlib) lists.erl:1263: :lists.foldl/3
    (gradualizer) /Users/elixir_sense/deps/gradualizer/src/typechecker.erl:3228: :typechecker.check_clauses/4
    (gradualizer) /Users/elixir_sense/deps/gradualizer/src/typechecker.erl:2349: :typechecker.do_type_check_expr_in/3
```

zuiderkwast:
Just some notes.

For `any()` there is a clause for "switching direction" in the type checking here:

```Erlang
do_type_check_expr_in(Env, {type, _, any, []}, Expr) ->
    {_Ty, VB, Cs} = type_check_expr(Env, Expr),
    {VB, Cs};
```

For `term()`, there is no such clause, so we get term() in places where it's not expected. I think this will be solved by normalizing term() to any().

josefs:
@lukaszsamson I believe my diff fixed your problem. Would you care to check?

lukaszsamson:
@josefs It did. Thanks.

PR #251, opened 2020-01-11
Normalize away term() to any()

josefs:
Removing `term()` during normalization will enable us to ignore
this type and simplify the typechecker. Currently, we have
a few odd bugs because we don't remove `term()`.

Thanks to @zuiderkwast for identifying the problem in #248

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -657,6 +657,8 @@ normalize(Type, _TEnv) ->
 -spec expand_builtin_aliases(type()) -> type().
 expand_builtin_aliases({var, Ann, '_'}) ->
     {type, Ann, any, []};
+expand_builtin_aliases({type, Ann, term, []}) ->
+    {type, Ann, any, []};
 expand_builtin_aliases({type, Ann, binary, []}) ->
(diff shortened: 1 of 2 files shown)


src/typechecker.erl
```diff
@@ -595,6 +595,8 @@ glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
 %% * Replace built-in type synonyms
 %% * Flatten unions and merge overlapping types (e.g. ranges) in unions
 -spec normalize(type(), TEnv :: #tenv{}) -> type().
+normalize({type, _, term, []}, TEnv) ->
+    type(any);
```
zuiderkwast: Now that I'm thinking about it: This could be placed in `expand_builtin_aliases`, since term() is listed as an alias for any() in http://erlang.org/doc/reference_manual/typespec.html. `expand_builtin_aliases` almost exactly follows this list.

josefs: Yeah, that makes total sense. I've changed it. Thanks!

PR #252, opened 2020-01-11
Catch errors in highlighter and fallback to non-fancy error message

zuiderkwast:
Some abstract code compiles, even if it can't be pretty-printed and parsed again.

This commit makes sure that error messages can be generated when typechecking programs containing such code.

An example of such code is the Elixir equivalent of

    -type mytype() :: {foo :: integer(), bar :: atom()}.


src/gradualizer_fmt.erl
```diff
@@ -404,6 +419,9 @@ highlight_in_context(AstNode, Opts) ->
      end,
      $\n]. % blank line after
 
+%% Determines if colors (escape sequences) should be used in the output.  If the
+%% option for colors is 'auto' (the default), io:columns/0 is used to detect if
+%% standard output is a TTY or not.
```
josefs: 👍

PR #253, opened 2020-01-12
Remove position and annotations as early as possible

zuiderkwast:
This commit sets the location annotation in types to 0 as early as possible using `typelib:remove_pos/1`, i.e. when the types are loaded from beam or erl code, fetched from gradualizer_db or created.

Annotated types (Name :: Type) are also removed in `typelib:remove_pos/1`, so those don't need to be handled anywhere else.

### DIFF
--- a/src/gradualizer_db.erl
+++ b/src/gradualizer_db.erl
@@ -26,6 +26,11 @@
 
 %% Types for the Erlang Abstract Format
 -type type() :: gradualizer_type:abstract_type().
+-type typed_record_field() :: {typed_record_field,
+                               {record_field, erl_anno:anno(),
+                                Name :: atom(),
+                                Default :: erl_parse:abstract_expr()},
+                               type()}.
 
 %% Compiled regular expression
 -type regexp() :: {re_pattern, _, _, _, _}.
@@ -78,7 +83,7 @@ get_opaque_type(M, T, A) ->
 
 %% @doc Fetches a record type defined in the module.
 -spec get_record_type(Module :: module(),
-                      Name :: atom()) -> {ok, [type()]} | not_found.
+                      Name :: atom()) -> {ok, [typed_record_field()]} | not_found.
 get_record_type(Module, Name) ->
     call({get_record_type, Module, Name}).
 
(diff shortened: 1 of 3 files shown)


src/typechecker.erl
```diff
@@ -129,8 +137,10 @@ any_subtype([Ty1|Tys], Ty, TEnv) ->
 %% The function compat_ty is just a convenience function to be able to
 %% pattern match on types in a nice way.
 compat(T1, T2, A, TEnv) ->
-    Ty1 = typelib:remove_pos(normalize(T1, TEnv)),
-    Ty2 = typelib:remove_pos(normalize(T2, TEnv)),
+    Ty1 = normalize(T1, TEnv),
+    Ty2 = normalize(T2, TEnv),
+    ?assert_normalized_anno(Ty1),
+    ?assert_normalized_anno(Ty2),
```
josefs: I was expecting the assertions to be checked before normalization. Isn't the invariant that you're trying to establish that all types have their annotations normalized **before** calling `normalize`?

zuiderkwast: OK. I was thinking that the only place where the annotations need to be zeroed is for the call to `sets:is_element({Ty1, Ty2}, A)`, but `remove_pos` was done before `normalize` before, so I'll keep that order. I don't think normalize/2 adds any non-zero annotations.

The assertion just check the top-level anno though, so it's not bullet proof. Just something for the transition.

src/gradualizer_db.erl
```diff
@@ -26,6 +26,11 @@
 
 %% Types for the Erlang Abstract Format
 -type type() :: gradualizer_type:abstract_type().
+-type typed_record_field() :: {typed_record_field,
```
Zalastax: Does it matter that the type uses a tuple instead of the equivalent record?

zuiderkwast:
Thanks for the meme! I still use the old-school ascii smileys. ;-)

src/gradualizer_db.erl
zuiderkwast: Which record? This is an abstract form. They are defined as tuples, not records.

This change fixed the dialyzer error btw.

josefs:
We use memes and gifs at work. It takes reviewing to a whole new level :-)

Issue #25, opened 2018-07-22
Type refinement in conditional expressions


gomoripeti:
(this ticket was split from https://github.com/josefs/Gradualizer/issues/23 which also contain some of the early comments)

let's see the following correct function with spec
```
-spec f(atom() | integer()) -> true | integer().
f(AI) ->
    if is_atom(AI) ->    %% Expr1
         true;
       true ->
          AI + 1         %% Expr2
    end.
```
- the type of variable `AI` in `Exrp2` must be the subtype of `number() :: integer() | float()` as the first argument of `+`
- the type of variable `AI` in `Exrp2` is `atom() | integer()` based on the spec which is not a subtype of `integer() | float()`

How could Gradualizer realise that `Expr2` is only executed when type of `AI` is `integer()`?

- if type checking was supersmart it would realise that `if NOT is_atom(AI) => typeof(AI) == (atom() | integer()) & (not atom()) == integer()` - however this is not always possible

Let's see a more generic example
```
-spec f(input()) -> any().
f(A) ->
  case cond(A) of
    true -> true;
    false -> body(A)
  end.

-spec cond(input()) -> boolean().
-spec body(required()) -> any().
```
In this case what the type checking could do is realise that `body(A)` is only executed in a subset of cases therefor type of `A` in `body(A)` is just a subset or subtype of `input()`. In case `input() & required() == none()` (the intersection of the two types is empty) then we can say that `body(A)` will always fail, but otherwise we cannot know (and have to leave type checking at runtime). This method requires to be able to calculate the **intersection** of two types.

What is the right way to handle these kind of constructs? (`if`, `case`, multiple function clause)

josefs:
@gomoripeti, since you opened this ticket we've implemented type refinement. It doesn't handle the cases that you imagined though. Are you still keen on pursuing the kind of refinement that you outlined or should we let the machinery we have now suffice?

zuiderkwast:
The code for `if` clauses is already unified with the code handling case, function, receive, try-of and catch clauses (with an empty list as the patterns), so we get the type refinement implemented for the other cases also in `if` automatically.

The remaining part for the first example above is to implement refinement based on the `is_TYPE` guards for **mismatching** clauses, i.e. to subtract those types in certain cases. (This can only be done in a much more limited set of situations, since we need to be sure that the clause hasn't mismatched for some other reason. We must be able to "blame" the specific guard function for the mismatch.)

gomoripeti:
[...]

PR #255, opened 2020-01-13
Refinement of bound vars using guard BIFs for mismatching clauses

zuiderkwast:
This now passes:

```Erlang
-spec refine_bound_var_by_guard_bifs(ok | integer() | pid() |
                                     #{x => y} | {x, y}) -> ok.
refine_bound_var_by_guard_bifs(X) ->
    if
        is_map(X)     -> ok;
        is_tuple(X)   -> ok;
        is_integer(X) -> ok;
        is_pid(X)     -> ok;
        true          -> X
    end.
```

Fixes some of the examples in #25.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3208,18 +3208,18 @@ check_clauses(Env, any, ResTy, [{clause, _, Args, _, _} | _] = Clauses) ->
     %% 'any' is the ... in the type fun((...) -> ResTy)
     ArgsTy = lists:duplicate(length(Args), type(any)),
     check_clauses(Env, ArgsTy, ResTy, Clauses);
-check_clauses(Env, [], {var, _, TyVar}, Clauses) ->
-    %% This case was added for 'if' clauses.
-    {Ty, VarBinds, Cs} = infer_clauses(Env, Clauses),
-    {VarBinds, constraints:combine(constraints:upper(TyVar, Ty), Cs)};
-check_clauses(Env, ArgsTy, ResTy, Clauses) ->
-    {VarBindsList, Css, RefinedArgsTy} =
-        lists:foldl(fun (Clause, {VBs, Css, RefinedArgsTy}) ->
+check_clauses(Env = #env{tenv = TEnv}, ArgsTy, ResTy, Clauses) ->
+    %% Clauses for if, case, functions, receive, etc.
+    {VarBindsList, Css, RefinedArgsTy, _VEnvJunk} =
+        lists:foldl(fun (Clause, {VBs, Css, RefinedArgsTy, VEnvIn}) ->
                             {NewRefinedArgsTy, VB, Cs} =
-                                check_clause(Env, RefinedArgsTy, ResTy, Clause),
-                            {[VB | VBs], [Cs | Css], NewRefinedArgsTy}
+                                check_clause(Env#env{venv = VEnvIn},
+                                             RefinedArgsTy, ResTy, Clause),
+                            VEnvOut =
+                                refine_vars_by_mismatching_clause(Clause, VEnvIn, TEnv),
+                            {[VB | VBs], [Cs | Css], NewRefinedArgsTy, VEnvOut}
                     end,
-                    {[], [], ArgsTy},
+                    {[], [], ArgsTy, Env#env.venv},
                     Clauses),
     % Checking for exhaustive patternmatching
     case {Env#env.exhaust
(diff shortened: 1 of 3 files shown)


src/typechecker.erl
```diff
@@ -3423,6 +3431,47 @@ refinable(_) ->
...
+                                catch no_refinement -> Ty
+                                end,
+                    VEnv#{Var := RefinedTy};
+                _ ->
+                    %% No refinement
+                    VEnv
+            end;
+        _OtherGuards ->
+            %% No refinement
+            VEnv
```
FrankBro: Great work. Would it be hard to add `andalso`/`and` and `orelse`/`or`?

zuiderkwast:
> My only concern is that the rules for when refinement happens become increasingly complex. Please update the wiki so we can have a good sense of how we communicate to the programmer when they can rely on refinement.

I agree. @FrankBro can you update with what's already been merged? Then I can add stuff from this PR. It's not good to let documentation lag behind.

zuiderkwast:
Documentation draft:

### Refinement of bound variables after mismatching clauses

The type of a variable `Var` bound already before an `if`, `case` or other construct with clauses, can be refined using information about why the previous clauses have failed to match. The following conditions must be fulfilled for this refinement to happen:

1. The patterns of a clause must match all possible input (meaning that every pattern is either `_` or a free variable)
    * Note: `if` clauses have no patterns, so this condition is automatically fulfilled
2. There is a single guard on the form `is_TYPE(Var)` for the clause and no other guards are present

If these conditions are fulfilled, the only reason for the clause to mismatch is the guard. Thus, for the following clauses within the same construct (`if` or `case` etc.), the type of Var can be refinement by *excluding* TYPE from the type of Var.

Example:

```Erlang
-spec f(integer() | atom()) -> atom().
f(Var) ->
    case Var of
        _ when is_integer(Var) -> hello;
        _                      -> Var       % here Var :: atom()
    end.
```

src/typechecker.erl
zuiderkwast: Possibly for in limited cases, but I'd prefer to leave it to a separate PR.

For a guard sequence like `is_atom(X); is_integer(X)`, where all the guards are on the same variable X, the type of X can be refined by excluding both atom() and integer().

Do you have any other example in mind?

FrankBro:
Yes, will try to come up with something asap.

src/typechecker.erl
FrankBro: Totally fine to split it in multiple PRs.

This is a hair-pulling example a bit. Only useful if we want to have it as a test I guess.

```erlang
-spec test(integer() | atom() | list()) -> list().
test(X) ->
    if
        is_atom(X) orelse is_integer(X) -> [];
        true -> X
    end.
```

It also reminded me of the record part I did. Could also be another PR.

```erlang
-record(r1, {f :: integer()}).
-record(r2, {f :: integer()}).

-spec record(#r1{} | #r2{}) -> integer().
record(#r1{f = F}) -> F;
record(R2) -> R2#r2.f.
```

FrankBro:
Can we have a copy of the wiki in a folder somewhere so we can do PRs on it? I feel like this PRs changes to the wiki and the one for guards is kind of related.

Zalastax:
@FrankBro, GitHub doesn't provide an easy way to use that workflow. There are workarounds though: https://www.growingwiththeweb.com/2016/07/enabling-pull-requests-on-github-wikis.html. It's okay if you just make the edits you want to do. Someone can make additional changes afterwards if they have any comments.

FrankBro:
Which is why I was suggesting having a copy in like doc or something.

I tried to merge all three PRs, idk if it's good. I suck at documentation 😭 

### Refinement of bound variables
The type of a variable Var bound already before an if, case or other construct with clauses, can be refined using information such as guards, record matching as well as why the previous clauses have failed to match. The following conditions must be fulfilled for this refinement to happen:

The patterns of a clause must match all possible input (meaning that every pattern is either _ or a free variable)
Note: if clauses have no patterns, so this condition is automatically fulfilled
There is a single guard on the form is_TYPE(Var) for the clause and no other guards are present
If these conditions are fulfilled, we can refine the variable. Additionally, if the only reason for the clause to mismatch is the guard, we can refine the type for the following clauses. Thus, for the following clauses within the same construct (if or case etc.), the type of Var can be refinement by excluding TYPE from the type of Var.

Example:
```erlang
-record(r, {f :: atom()}).
-spec f(integer() | atom() | #r{}) -> atom().
f(Var) ->
    case Var of
        R = #r{}                         -> R#r.f;
        _ when is_integer(Var) -> hello;
        _                                     -> Var       % here Var :: atom()
    end.
```

zuiderkwast:
@FrankBro, the clause `R = #r{} -> R#r.f` is not about refining *already bound variables* for *mismatching* clauses. R is a variable *bound in the clause* if the clause is *matching*. Thus, I don't think it belongs in the same section.

zuiderkwast:
I updated the [wiki page](https://github.com/josefs/Gradualizer/wiki/Type-refinement) and added stubs for the matching clauses as well. The two main sections are for *matching clauses* and for *mismatching clauses*.

Feel free to add stuff!

Another possibility for work-in-progress documentation is to create a temporary wiki page for the draft and/or discussions. (Compare to Wikipedia's discussion tab of any page. It's just another wiki page.)

PR #257, opened 2020-01-15
Refine pattern using unsatisfied guards

zuiderkwast:
Solving one remaining kind of refinements:

```Erlang
-spec f(integer() | atom()) -> atom().
f(N) when is_integer(N) -> ok;
f(A) -> A. % here A :: atom()
```

Solves the known problem by moving it to should pass: https://github.com/josefs/Gradualizer/compare/master...zuiderkwast:refine_pattern_using_unsatisfied_guards?expand=1#diff-125dcd766b91e71dfbd59fffda480511


zuiderkwast:
```
 _____ _          _ 
|_   _| |__ __  _| |
  | | | '_ \\ \/ / |
  | | | | | |>  <|_|
  |_| |_| |_/_/\_(_)
```

PR #259, opened 2020-01-18
Increase coverage

Zalastax:
Added some tests guided by coverage. Also aligned the return for the send operator `!`.

### DIFF
--- a/test/known_problems/should_fail/map_expr.erl
+++ b/test/known_problems/should_fail/map_expr.erl
@@ -1,0 +1,8 @@
+-module(map_expr).
+
+-export([foo/0]).
+
+-spec foo() -> ok.
+foo() ->
+    A = #{ok => 42},
+    A.
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1648,10 +1648,11 @@ do_type_check_expr(Env, {'receive', _, Clauses, _After, Block}) ->
 
 %% Operators
 do_type_check_expr(Env, {op, _, '!', Proc, Val}) ->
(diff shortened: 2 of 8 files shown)


test/known_problems/should_fail/map_expr.erl
```diff
@@ -0,0 +1,8 @@
+-module(map_expr).
+
+-export([foo/0]).
+
+-spec foo() -> ok.
+foo() ->
+    A = #{ok => 42},
+    A.
```
zuiderkwast: I think this would fail with `-gradualizer(infer)` but maybe it should not fail without infer, since in general literals are considered untyped. (Then, we may consider changing this, but that's another story.)

zuiderkwast:
```
   _____       __               ___     ____
  / ___/____  / /__  ____  ____/ (_)___/ / /
  \__ \/ __ \/ / _ \/ __ \/ __  / / __  / / 
 ___/ / /_/ / /  __/ / / / /_/ / / /_/ /_/  
/____/ .___/_/\___/_/ /_/\__,_/_/\__,_(_)   
    /_/                                     

```

test/known_problems/should_fail/map_expr.erl
Zalastax: It doesn't fail with infer either 😢. Should I move this to `known_problems/should_fail_infer` (and add support for that folder)?

zuiderkwast: Just add `-gradualizer(infer).` to the file, perhaps. No need for separate folder IMO.

PR #260, opened 2020-01-19
Increase coverage

Zalastax:
Added some tests guided by coverage. Also add new folders ` test/should_pass_infer` and ` test/should_fail_infer`.


Zalastax:
@josefs, do you mean that you disagree with the naming? I'm up for suggestions!

josefs:
What I mean is that I don't really see the need for `should_fail_infer`, the files in that directory could just as well go into `should_fail`. But again, I don't think it's super important. And it's easy to change.

Zalastax:
Aha!

Do you have a suggestion for how to know that the infer option should be used?

zuiderkwast:
I also think the new directories are unnecessary.

Just add `-gradualizer(infer).` to in the files which should be checked with infer enabled. Doesn't it work? It is supposed to be implemented and I think there is some test case using this already.

zuiderkwast:
* https://github.com/josefs/Gradualizer/blob/master/test/should_fail/infer_enabled.erl
* https://github.com/josefs/Gradualizer/blob/master/test/known_problems/should_fail/infer_any_pattern.erl

Zalastax:
I'm now using `-gradualizer(infer).` instead! I was not aware of it.

Zalastax:
I think it's likely they all would be solved. The tricky thing is to do that without reducing performance and without losing precision. I got stuck with #163 last time I tried. Do you think we can salvage the idea?

zuiderkwast:
> Do you think we can salvage the idea?

We can keep it in mind. I think it's doable but we should sort out the performance issues first and maybe other things as well. For now, it's good to have the known problems stored.

Zalastax:
Added some commits after your approvals. Please take a quick look again. Feel free to merge.

josefs:
> Do you think we can salvage the idea?

We would have to give up some precision in the types. Right now we can compute rather accurate types for ranges of integers, e.g. if we have `X :: 1..2` and `Y :: 3..4` then `X + Y :: 3..6`. That's not something we can do if we change the typechecking of operators from specialized code to just using type specs.

zuiderkwast:
Ideally, we can do both, i.e. keep the special logic for arithmetic operations, `record_info`, etc. and for the rest of the operators, they can work just like specs and we can define their specs in the prelude.

I think the special logic is good for precision, not only for binary operators but we also for functions like `element` and `set_element`. It can increase precision, just like refinements.

Zalastax:
Alright! Then we should still look into why some operators behave strangely. 

'll look into reviving #163 .

src/typechecker.erl
```diff
@@ -153,6 +153,7 @@ compat_ty({type, _, any, []}, _, A, _TEnv) ->
     ret(A);
 compat_ty(_, {type, _, any ,[]}, A, _TEnv) ->
     ret(A);
+%% The clauses for term should never happen since term() is an alias of any().
 compat_ty({type, _, term, []}, _, A, _TEnv) ->
     ret(A);
 compat_ty(_, {type, _, term ,[]}, A, _TEnv) ->
```
zuiderkwast: Now that term() is normalized to any(), these clauses should be deleted. We only need to handle types on head normal form.

test/should_pass/operator_subtypes.erl
```diff
@@ -104,3 +104,18 @@ unary_op4(X) -> -X.
 unary_op5(X) ->
     A = bnot X,
     A.
+
+-spec unary_op6(pos_integer()) -> neg_integer().
+unary_op6(X) ->
+    A = -X,
+    A.
```
zuiderkwast: Why not just `unary_op6(X) -> -X.`? Does that cause code coverage to be different?

Zalastax: Yes. It happens a fair bit due to bidirectionality. I think the code you suggested is already covered elsewhere. If it's not, I'll find it in another coverage hunt!

josefs: An alternative to using the trick with variables is to use `?annotate_type(Expr, any())`. It's also a bit obscure. Perhaps it's worth defining a new macro with a helpful name to use in these types of test where we want to make sure we test inference (inference, in the sense that we don't know what the final type should be, not in the `--infer` sense) rather than checking.

src/typechecker.erl
Zalastax: We don't want to keep these as safety? Then I'll remove them. At all places where I commented?

test/should_pass/operator_subtypes.erl
Zalastax: A new macro to be used in the tests is a great idea! Can you show how `?annotate_type(Expr, any())` would be used here to verify that `-X :: neg_integer()`? Can the new macro be crafted to verify the exact type instead of just compatibility?

josefs: Hmmm, `?annotate_type` wouldn't help to verify `-X :: neg_integer()`. 

I feel a little uneasy about the `X = <bla>, X.` pattern because ideally we would like gradualizer to actually typecheck this better. In the future I would like the typechecking of variable bindings to go like this: create a type variable, let's call it `T` and give `X` the type `T`. Then typecheck `<bla>` with the expected type `T`. In the rest of the expression, use `X :: T` in the environment. This will propagate types better that the way we're doing it now, but relies on having a constraint solver in place. 

I think a more robust syntax would be to say `case <bla> of X -> X.` That could be made into a macro.

zuiderkwast: Another idea would be to unfold the variables, effectively rewriting the function to `unary_op6(X) -> -X`. To do this, we can bind the variable to the expression instead of a type. Then we have `A = -X` in the environment and whenever we run into `type_check_expr_ty(A,  neg_integer())` it expands to `type_check_expr_ty(-X, neg_integer())` (pseudo-code).

I'm trying to avoid relying on constraint-solving because, after reading the paper POPL19 which covers polymorphism with set-theoretic gradual types, it seems very hard. They solve it (by completely redefining stuff, replacing the compatibility relation with a "materialization" rule, "frame variables", etc etc), but then instead, intersection types for function types are not supported in that system. It's listed under *future work*:

*"This work lays a foundation for integrating gradual typing and full set-theoretic types and, as such,
it opens many new questions and issues. There are in particular two practical issues that we want
to address in the near future.*

*The first is to address a restriction we imposed to our system: namely, that it is not possible
to assign intersection types to a function. Forbidding that (other than by subsumption) was an
early design choice of this work, motivated by several reasons: its absence would complicate the
dynamic semantics of the cast calculus (see Castagna and Lanvin [2017], where this restriction is
not present); it would make type reconstruction and constraint solving much more difficult, and
it would have probably hindered completeness even for simple systems; a system without this
restriction would have been interesting only if the language had a type-case construct, which we
wanted to avoid for simplicity and for sticking as close as possible to ML."*

src/typechecker.erl
zuiderkwast: Yes, remove all.

It it doesn't work, it means something is not normalized where it should be.

Issue #258, opened 2020-01-16
Type refinement using comparison guards


zuiderkwast:
Refinement using comparisons using `<, >, =<, >=, ==, /=, =:=, =/=` in guard expressions on the form `Var <op> Const` and `Const <op> Var`.

```Erlang
%% Matching clause
-spec p1(integer()) -> pos_integer().
p1(N) when N > 0 -> N;
p1(_) -> 42;

%% Mismatching clause
-spec p2(integer()) -> pos_integer().
p2(N) when N =< 0 -> 42;
p2(N) -> N; %% Here the range neg_inf..0 is excluded, so N :: pos_integer().

%% More complex mismatching clause
-spec p3(integer() | undefined) -> pos_integer().
p2(N) when N < 0; is_atom(N) -> 42;
p2(N) -> N.
```

Erlang's term ordering

    number < atom < reference < fun < port < pid < tuple < map < nil < list < bit string

can be used to turn an expression like X < 10 into the refined var binding `X :: neg_integer() | 0..9 | float()`.

Likewise, from a guard like X > 2 we get `X :: 3..pos_inf | float() | atom() | reference() | ... | bitstring()`.

Comma-separated guard expressions like `when X > 2, X < 10` are combined using GLB to produce  `X :: 3..9`.

I suggest dividing the work into 3 PRs:

### 1. Matching clauses

I think implementation is straight-forward.

### 2. Mismatching clauses with only one guard expression

We must be able to prove that the combination of guards is the reasons for the clause to mismatch, so I suggest the following limitations:

1. The patterns must be match-all (free vars and `_`)
2. Only one guard expression exists for the clause

### 3. Mismatching clauses with andalso/orelse

The 2nd limitation for mismatching clauses is replaced by these three:

* All guard expressions in the same clause must constrain the same variable
* Guards on the form is_TYPE(X) are also allowed
* Combinations of comma, semicolon, and, or, andalso, orelse allowed
    * comma, and, andalso => GLB
    * semicolon, or, orelse => LUB

PR #263, opened 2020-01-21
Refinement using comparison guards in matching clauses

zuiderkwast:
Step 1 in #258

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3270,8 +3270,8 @@ refine_ty({Tag1, _, M}, {Tag2, _, N}, _TEnv)
     if M == N -> type(none);
        M /= N -> throw(disjoint)
     end;
-refine_ty(Ty1, Ty2, TEnv) when ?is_int_type(Ty1),
-                               ?is_int_type(Ty2) ->
+refine_ty(Ty1, Ty2, _TEnv) when ?is_int_type(Ty1),
+                                ?is_int_type(Ty2) ->
     gradualizer_int:int_type_diff(Ty1, Ty2);
 refine_ty(Ty1, Ty2, TEnv) ->
     case glb(Ty1, Ty2, TEnv) of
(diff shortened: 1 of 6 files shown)


src/typechecker.erl
```diff
@@ -3388,11 +3388,60 @@ check_guard_call(is_reference, [{var, _, Var}]) -> #{Var => type(reference)};
...
+type_comp_op('=:=', I) when is_integer(I) ->
+    {integer, erl_anno:new(0), I};
+type_comp_op(Op, A) when Op =:= '==' orelse Op =:= '=:=',
+                         is_atom(A) ->
+    {atom, erl_anno:new(0), A};
+type_comp_op(Neq, I) when Neq =:= '/=' orelse Neq =:= '/=',
+                          is_integer(I) ->
+    {type, _, union, U1} = type_comp_op('<', I),
+    {type, _, union, U2} = type_comp_op('>', I),
+    type(union, U1 ++ U2);
```
josefs: Do you need to normalize the union here?

test/should_pass/refine_comparison.erl
```diff
@@ -0,0 +1,26 @@
...
+comp4(_) -> 1.
+
+-spec compatom1(a | b | c) -> a | b | ok.
+compatom1(X) when X == a -> X;
+compatom1(X) when X =:= b -> X;
+compatom1(_) -> ok.
+
+-spec compatom2(a | b | c | pid()) -> a | c | pid().
+compatom1(X) when X =/= b -> X;
+compatom1(_) -> a.
```
josefs: I suggest adding the faculty function here as well, written using guards, rather than pattern matching.

```
-spec fac(non_neg_integer()) -> pos_integer().
fac(N) when N == 0 ->
    1;
fac(N) ->
    N * fac(N-1).
```

FrankBro:
Duuude nice.

test/should_pass/refine_comparison.erl
zuiderkwast: Good idea. This PR doesn't solve this though. The refinement done here only concerns the matching clause. Excluding 0 in subsequent clauses is another PR.

src/typechecker.erl
zuiderkwast: It should already contain only normalized types here, or at least it would be better that way. Now, that I think about it, `type('fun')` , `type(list)` and `type(bitstring)` should return the normalized forms `fun((...) -> any())`, `[any()]` and `<<_:0, _:_*1>>`. I think specific clauses in `type/1` should do that.

zuiderkwast:
> The diff has a bit low coverage.

@Zalastax, yes, I'll add a few more tests. I didn't look at the coverage before posting the PR.

> It would also be nice to spell out what type_comp_op can deal with. Let's add it to the wiki after the merge?

Yes... and then there will be part 2 and 3 of #258. Perhaps the wiki changes can cover all at once.

> I suspect there is a simpler implementation of type_comp_op lurking somewhere which generalizes to handle the whole term order.

@josefs, how would we do that?

PR #264, opened 2020-01-22
Add specs in constraints module

Zalastax:
I'm not sure that these new specs are correct. Notice that `atom()` is replaced with `string()` in `lower_bounds` and `upper_bounds`.

This worries me:

https://github.com/josefs/Gradualizer/blob/cabde8b7a74b8f2194e806783df2427882e29a11/src/typechecker.erl#L2475-L2480

`new_type_var() :: string()` but `type_var(atom()) -> af_type_variable()`. Should we instead have `-type var() :: atom().` and change the implementation of `new_type_var/0`?

The sets module is also [troublesome](https://github.com/blackberry/Erlang-OTP/blob/3ec06ce97019eed9472a2cf8950ed23d7a2e6b9b/lib/stdlib/src/sets.erl#L70-L75):

```
%% A declaration equivalent to the following one is hard-coded in erl_types.
%% That declaration contains hard-coded information about the #set{}
%% record and the types of its fields.  So, please make sure that any
%% changes to its structure are also propagated to erl_types.erl.
%%
%% -opaque set() :: #set{}.
```

I was therefore not able to override `sets:add_element/2` in a sets prelude I was about to make.
```
-module(sets).

-spec add_element(A, sets:set(A)) -> sets:set(A).
````

I managed to get this type error (for `constraints:add_var/2`) running `make gradualize`
> The record field on line 24 is expected to have type sets:set(A) but it has type sets:set(var())

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -88,6 +88,8 @@
              %,tyvenv  = #{}
              }).
 
+-include("gradualizer.hrl").
+
 %% Two types are compatible if one is a subtype of the other, or both.
 compatible(Ty1, Ty2, TEnv) ->
     case {subtype(Ty1, Ty2, TEnv), subtype(Ty2, Ty1, TEnv)} of
@@ -1127,7 +1129,7 @@ expect_record_type(Record, {type, _, union, Tys}, TEnv) ->
     expect_record_union(Record, Tys, TEnv);
 expect_record_type(Record, {var, _, Var}, _TEnv) ->
     {ok, constraints:add_var(Var,
-           constraints:upper(Var, {record, erl_anno:new(0), Record}))};
+           constraints:upper(Var, type_record(Record)))};
 expect_record_type(_, _, _) ->
     type_error.
 
@@ -1141,6 +1143,7 @@ expect_record_union(Record, [Ty | Tys], TEnv) ->
 expect_record_union(_Record, [], _TEnv) ->
     type_error.
 
+-spec new_type_var() -> constraints:var().
 new_type_var() ->
     I = erlang:unique_integer(),
     "_TyVar" ++ integer_to_list(I).
--- a/src/constraints.erl
+++ b/src/constraints.erl
@@ -7,28 +7,35 @@
 -type type() :: gradualizer_type:abstract_type().
 
 -record(constraints, {
-          lower_bounds = #{}        :: #{ atom() => [type()] },
-          upper_bounds = #{}        :: #{ atom() => [type()] },
-          exist_vars   = sets:new() :: sets:set(string())
+          lower_bounds = #{}        :: #{ var() => [type()] },
+          upper_bounds = #{}        :: #{ var() => [type()] },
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/typechecker.erl [hunk 1]
```diff
@@ -1127,7 +1129,8 @@ expect_record_type(Record, {type, _, union, Tys}, TEnv) ->
     expect_record_union(Record, Tys, TEnv);
 expect_record_type(Record, {var, _, Var}, _TEnv) ->
     {ok, constraints:add_var(Var,
-           constraints:upper(Var, {record, erl_anno:new(0), Record}))};
+           constraints:upper(Var, ?annotate_type({record, erl_anno:new(0), Record},
+                                                 type())))};
```
zuiderkwast: A record type is written `{type,_,record,[{atom,_,my_record}]}`. That's the error.

(The reason `?annotate_type` passes without error is that the literal is considered untyped, which makes it possible to hide the error like this. With infer, a type error would be reported for `?annotate_type`, I think.)

zuiderkwast:
> Should we instead have -type var() :: atom(). and change the implementation of new_type_var/0?

Yeah, that's the reason a type containing these are not regular `erl_parse:abstract_type()`s.

How do you create atoms without exploding the atom table?

josefs:
We should probably always represent type variables as strings (or perhaps binaries, for efficiency) and get rid of the atoms internally. We could use `normalize` for that.

josefs:
I don't understand why you couldn't override `sets:add_element/2` but from the comment in the OTP module it sounds like we may also have to hard-wire the definition of sets.

zuiderkwast:
I don't understand how overriding the spec of `sets:add_element/2` would change anything. The original one below is expanded to your version without `when` by Ulf's unfolding, isn't it?

```Erlang
-spec add_element(Element, Set1) -> Set2 when
      Set1 :: set(Element),
      Set2 :: set(Element).
```

I'm not sure I completely understand the problem. If we can pin-point it, we should add a minimal equivalent as a `known_problem`.

zuiderkwast:
> We should probably always represent type variables as strings (or perhaps binaries, for efficiency) and get rid of the atoms internally. We could use `normalize` for that.

This could simplify some handling, but keep in mind that normalize only normalizes the outer level (head normal form) so variables inside tuples etc. would still be atoms, so ... yes, as long as we know what we're doing...

Or we could just type them as `var() :: atom() | string()` and keep the implementation as is (or possibly replacing string with binary, but that's a side track). How about that?

josefs:
> This could simplify some handling, but keep in mind that normalize only normalizes the outer level (head normal form) so variables inside tuples etc. would still be atoms, so ... yes, as long as we know what we're doing...
> 
> Or we could just type them as `var() :: atom() | string()` and keep the implementation as is (or possibly replacing string with binary, but that's a side track). How about that?
> 
Good point about normalization. Having `var() :: atom() | string()` feels a bit icky but it might be the simplest solution overall.

Zalastax:
Let me know if there's anything else to change. I will autosquash before merging.

PR #266, opened 2020-02-01
Add known problems regarding intersection with any

zuiderkwast:
We could perhaps regard this as a feature request rather than a known problem. Anyway, it is highlighting what we currently do for intersection with any.

Since it's not possible use `&` in spec, other ways are used to represent this, e.g. type refinement.


zuiderkwast:
These are examples I think @NelsonVides wanted to handle in #206 though, which spawned a lot of discussion.

@josefs, I agree it would be a lot of work: Handling intersection types in normalize, subtype, glb and lots of other places. But I can't see why it would requiring extending the type language beyond intersection types. Anyway, that's not important.

Issue #268, opened 2020-02-01
Integrate gradualizer into the erl repl


josefs:
It'd be really nice to run gradualizer automatically from the repl, whenever a module is loaded into the beam. Ultimately, I think this is where we want gradualizer to head. Though integrating gradualizer in the main OTP distribution is not likely to happen any time soon it would be nice to create a fork which has the functionality and shows how it can be done.

This is not likely to happen in the near future but I'm creating this ticket as a reminder.

zuiderkwast:
It's possible to override `c/1` and the other shell commands as described on http://erlang.org/doc/man/shell_default.html:

> To add your own commands to the shell, create a module called user_default and add the commands you want. Then add the following line as the first line in your .erlang file in your home directory.
>
>     code:load_abs("$PATH/user_default").

It's a bit easier than forking OTP. :-)

josefs:
That's amazing! I had no idea that functionality existed.

zuiderkwast:
This could be activated when people do `make shell` or the equivalent in their projects.

For example, the rebar3 plugin code set up the code path to provide this in `rebar3 shell` and gradualixir could provide it to the Elixir REPL.

josefs:
With #269 we get gradualizer functionality when using `c/1` in the repl. (But only when the argument is a filepath represented as a string). So that's a start.

PR #269, opened 2020-02-02
A first stab at typechecking automatically from the repl

josefs:
When the module 'user_default' is loaded it will override
the compile command in the repl so that files are typechecked
before compiled.

There are plenty of ways this implementation can be improved.
Right now the command only accepts strings as filepaths pointing
to source code files.
Secondly, it parses the file and passes it on to gradualizer.
After typechecking the absform is completely discarded and the
compiler starts parsing from scratch.

### DIFF
--- a/src/user_default.erl
+++ b/src/user_default.erl
@@ -2,7 +2,7 @@
 
 -export([c/1]).
 
-c(File) ->
+c(File) when not is_atom(File) ->
     % TODO: we can be a lot more clever about recognizing
     % if the argument is a file or a module, just like
     % shell_default:c/1.
@@ -11,4 +11,7 @@ c(File) ->
             shell_default:c(File);
         Err ->
             Err
-    end.
+    end;
+c(File) ->
+    shell_default:c(File).
+
--- a/src/user_default.erl
+++ b/src/user_default.erl
@@ -1,0 +1,14 @@
+-module(user_default).
+
+-export([c/1]).
+
+c(File) ->
+    % TODO: we can be a lot more clever about recognizing
+    % if the argument is a file or a module, just like
+    % shell_default:c/1.
+    case gradualizer:type_check_file(File) of
+        ok ->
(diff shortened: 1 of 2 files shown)


src/user_default.erl
```diff
@@ -0,0 +1,14 @@
+-module(user_default).
```
zuiderkwast: Is it safe to include a `user_default` in the repo? I mean, we can think about what happens if people are using their own user_defaults and don't want Gradualizer's one to override it. If the one loaded first takes priority, it's safe, right?

src/user_default.erl
```diff
@@ -0,0 +1,14 @@
...
+
+-export([c/1]).
+
+c(File) ->
+    % TODO: we can be a lot more clever about recognizing
+    % if the argument is a file or a module, just like
+    % shell_default:c/1.
+    case gradualizer:type_check_file(File) of
```
zuiderkwast: We could skip gradualizing if File is an atom. Then, `c(my_module)` is still usable at least.

src/user_default.erl
josefs: I'm not quite sure what the best solution is here. I just included it under `src/` because it was convenient but I agree it's not ideal. I suppose we can put it under `examples/` if you think that's better.

zuiderkwast: I don't know. Maybe it's no problem, but I haven't looked into it enough. Maybe src/ is fine but we need to change -pa to -pz in `make shell` or something like that. If there are multiple modules with the same name in the code path, the order in the code path matters...

src/user_default.erl
josefs: Yes.

src/user_default.erl
josefs: Ok, I've changed it to '-pz' which I think makes sense.

josefs:
Just for reference, this PR is a result of discussion in #268

PR #271, opened 2020-02-06
Record refinement

### LINKED ISSUE #245: Record refinement via function clauses
Here is what I'm trying to get working:

```erlang
-module(test).

-export([args/1, tuple/1, record/1]).

-spec args(integer() | undefined, integer()) -> integer().
args(undefined, I) -> I;
args(I, _) -> I.

-spec tuple({integer() | undefined, integer()}) -> integer().
tuple({undefined, I}) -> I;
tuple({I, _}) -> I.

-record(test, {
    field :: integer() | undefined
}).

-spec record(#test{}, integer()) -> integer().
record(#test{field = undefined}, I) -> I;
record(#test{field = I}, _) -> I.
```

The last test does not pass because we don't do refinement for records it seems. In fact, in `add_type_pat`, the type returned is always `type(none)`. 

I'm trying to implement this but not sure how to do it. Looking at the tuple, one thing that is nice is that we'll always have every fields and the type for a record is pretty easy to figure out:
`{type, _, tuple, [Field1, Field2]}`. The only type I've ever constructed for records had its name and that's it: `{type, _, record, [{atom, _, Name}]}`. How can I inject some fields being eliminated? 

Here is another examples with steps:

```erlang
-spec test({undefined | integer(), undefined | integer()}) -> integer().
test({undefined, undefined}) -> 0;
test({undefined, B}) -> B;
test({A, undefined}) -> A;
test({A, B}) -> A + B.
```

The type will be stepped this way:

```erlang
{type,0,tuple, [{type,0,union, [{atom,0,undefined}, {type,0,integer,[]}]}, {type,0,union, [{atom,0,undefined}, {type,0,integer,[]}]}]}
{type,0,union, [{type,0,tuple, [{type,0,integer,[]}, {type,0,union, [{atom,0,undefined},{type,0,integer,[]}]}]}, {type,0,tuple, [{type,0,union, [{atom,0,undefined},{type,0,integer,[]}]}, {type,0,integer,[]}]}]}
{type,0,union, [{type,0,tuple, [{type,0,integer,[]},{type,0,integer,[]}]}, {type,0,tuple, [{type,0,integer,[]}, {type,0,union, [{atom,0,undefined}, {type,0,integer,[]}]}]}]}
{type,0,tuple, [{type,0,integer,[]},{type,0,integer,[]}]}
```

- First line is the full type. 
- Second line created a union for the two possibilities where `undefined` is defined on either the left or right side.
- Third line eliminated the possibility that the left side would have `undefined`
- Forth line has seen every instances of `undefined` removed.

Technically, a record is a tuple so I assume this should be supported. How can I encode something similar? It would start as

```erlang
{type,0,tuple, [{type,0,record,[{atom,0,test}], {type,0,integer,[]}]}
???
```

FrankBro:
This is still a work in progress but I managed to have test cases passing so I thought I'd create it. It resolves #245.

Some code paths are not covered because I have yet to make test cases for everything.

The way I've solved this problem is by expanding the fields for refinement when needed. For example:

```erlang
-record(r, {f :: integer() | undefined}).
-spec test(#r{}, integer()) -> integer().
test(#r{f = undefined}, I) -> I;
test(#r{f = I}, _) -> I.
```

This is now supported. The `ArgsTy` basically goes like this

```erlang
% First call
[{type, _, record, [{atom, _, r}]}, integer()]
% After refinement of first call
[{type, _, record, [{atom, _, r}, {type, _, integer, []}]}, integer()]
```

Trying to find test cases that break stuff now. A lot of the code is straight up a ripoff of the tuple logic (if you check `expect_record_type`).

Also, we have to make sure we squash all those garbage commits when we merge it.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1176,16 +1176,10 @@ expect_record_type(Union = {type, _, union, UnionTys}, Record, TEnv) ->
             {fields_tys, Tyss, Cs}
     end;
 expect_record_type({var, _, Var}, Record, #tenv{records = REnv}) ->
-%%    Not sure how to make this work
-%%    TyVar = new_type_var(),
-%%    {fields_ty
-%%        ,{var, erl_anno:new(0), TyVar}
-%%        ,constraints:add_var(Var,
-%%            constraints:upper(Var, type_record(Record)))
-%%    };
     case REnv of
         #{Record := Fields} ->
-            {fields_ty, Fields, constraints:empty()};
+            Cs = constraints:add_var(Var, constraints:upper(Var, type_record(Record))),
+            {fields_ty, Fields, Cs};
         _NotFound ->
             {type_error, Record}
     end;
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1713,9 +1713,15 @@ type_check_fields(Env, Rec, [{record_field, _, {var, _, '_'}, Expr} | Fields]
                                      ,{atom, erl_anno:new(0), Field}, Expr}
                                      || Field <- UnAssignedFields]
                                   ,should_not_be_inspected),
-    {VB2, Cs2} = type_check_fields(Env, Rec, Fields, UnAssignedFields),
+    {VB1, Cs1};
+type_check_fields(_Env, _Rec, [], should_not_be_inspected) ->
+    {#{}, constraints:empty()};
+type_check_fields(Env, Rec, [], [UnAssignedField|UnAssignedFields]) ->
+    FieldTy = get_rec_field_type({atom, erl_anno:new(0), UnAssignedField}, Rec),
+    {VB1, Cs1} = type_check_expr_in(Env, FieldTy, {atom, erl_anno:new(0), undefined}),
+    {VB2, Cs2} = type_check_fields(Env, Rec, [], UnAssignedFields),
     {union_var_binds(VB1, VB2, Env#env.tenv), constraints:combine(Cs1,Cs2)};
-type_check_fields(_Env, _Rec, [], _U) ->
+type_check_fields(_Env, _Rec, [], []) ->
     {#{}, constraints:empty()}.
 
 get_unassigned_fields(Fields, All) ->
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -251,10 +251,6 @@ compat_ty({type, Anno1, record, [{atom, _, Name}|Fields1]},
 compat_ty({type, _, record, _}, {type, _, tuple, any}, A, _TEnv) ->
     ret(A);
 
-%% Record fields that have been refined
-compat_ty(?type_field_type(Name, Ty1), ?type_field_type(Name, Ty2), A, TEnv) ->
-    compat(Ty1, Ty2, A, TEnv);
-
 %% Lists
 compat_ty(Ty1, Ty2, A, TEnv) when ?is_list_type(Ty1), ?is_list_type(Ty2) ->
     {Empty1, Elem1, Term1} = list_view(Ty1),
@@ -315,7 +311,7 @@ compat_tys(_Tys1, _Tys2, _, _) ->
 
 compat_record_tys([], [], A, _TEnv) ->
     ret(A);
-compat_record_tys([Field1|Fields1], [Field2|Fields2], A, TEnv) ->
+compat_record_tys([?type_field_type(Name, Field1)|Fields1], [?type_field_type(Name, Field2)|Fields2], A, TEnv) ->
     {A1, Cs1} = compat(Field1, Field2, A, TEnv),
     {A2, Cs2} = compat_record_tys(Fields1, Fields2, A1, TEnv),
     {A2, constraints:combine(Cs1, Cs2)};
@@ -622,10 +618,9 @@ glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
 %% * Flatten unions and merge overlapping types (e.g. ranges) in unions
 -spec normalize(type(), TEnv :: #tenv{}) -> type().
 normalize({type, _, record, [{atom, _, Name}|Fields]}, TEnv) when length(Fields) > 0 ->
-    NormFields = [normalize(Field, TEnv) || Field <- Fields],
+    NormFields = [type_field_type(FieldName, normalize(Type, TEnv)) 
+        || ?type_field_type(FieldName, Type) <- Fields],
     type_record(Name, NormFields);
-normalize(?type_field_type(_Name, Type), TEnv) ->
-    normalize(Type, TEnv);
 normalize({type, _, union, Tys}, TEnv) ->
     Types = flatten_unions(Tys, TEnv),
     case merge_union_types(Types, TEnv) of
@@ -1160,7 +1155,7 @@ expect_record_type({type, _, record, [{atom, _, Name}|RefinedTypes]}, Record, #t
                     [] -> Fields;
                     _ ->
                         [ {typed_record_field, RecordField, RefinedType}
-                        || {{typed_record_field, RecordField, _}, RefinedType} <- lists:zip(Fields, RefinedTypes)]
+                        || {{typed_record_field, RecordField, _}, ?type_field_type(_, RefinedType)} <- lists:zip(Fields, RefinedTypes)]
                 end,
             {fields_ty, Tys, constraints:empty()};
         _NotFound ->
@@ -2225,7 +2220,8 @@ do_type_check_expr_in(Env, ResTy, {record_field, Anno, Expr, Name, FieldWithAnno
                             {type, _, record, [{atom, _, Name}]} ->
                                 {VarType, FieldTy};
                             {type, _, record, [{atom, _, Name}|RefinedFields]} ->
-                                {VarType, lists:nth(FieldIndex-1, RefinedFields)};
+                                ?type_field_type(_, RefinedField) = lists:nth(FieldIndex-1, RefinedFields),
+                                {VarType, RefinedField};
                             {var, _, _} ->
                                 {type_record(Name), FieldTy};
                             ?type(any) ->
@@ -3352,16 +3348,17 @@ refine_ty(?type(record, [{atom, Anno, Name}]), Refined = ?type(record, [{atom, _
     refine_ty(expand_record(Name, Anno, TEnv), Refined, TEnv);
 refine_ty(?type(record, [{atom, _, Name} | _]), ?type(record, [{atom, _, Name}]), _TEnv) ->
     type(none);
-refine_ty(?type(record, [Name|Tys1]), ?type(record, [Name|Tys2]), TEnv)
-  when length(Tys1) > 0, length(Tys1) == length(Tys2) ->
+refine_ty(?type(record, [Name|FieldTys1]), ?type(record, [Name|FieldTys2]), TEnv)
+  when length(FieldTys1) > 0, length(FieldTys1) == length(FieldTys2) ->
     % Record without just the name
+    Tys1 = [Ty || ?type_field_type(_, Ty) <- FieldTys1],
+    Tys2 = [Ty || ?type_field_type(_, Ty) <- FieldTys2],
     RefTys = [refine(Ty1, Ty2, TEnv) || {Ty1, Ty2} <- lists:zip(Tys1, Tys2)],
-    RecordsElems = pick_one_refinement_each(Tys1, RefTys),
+    RecordsTys = pick_one_refinement_each(Tys1, RefTys),
+    RecordsElems = [ [ type_field_type(FieldName, RecordTy) || {?type_field_type(FieldName, _), RecordTy} <- lists:zip(FieldTys1, RecordTys)]
+        || RecordTys <- RecordsTys],
     Records = [type(record, [Name|RecordElems]) || RecordElems <- RecordsElems],
     normalize(type(union, Records), TEnv);
-refine_ty(?type_field_type(Name, Ty1), ?type_field_type(Name, Ty2), TEnv) ->
-    RefTy = refine(Ty1, Ty2, TEnv),
-    normalize(type_field_type(Name, RefTy), TEnv);
 refine_ty(?type(union, UnionTys), Ty, TEnv) ->
     RefTys = lists:foldr(fun (UnionTy, Acc) ->
                              try refine(UnionTy, Ty, TEnv) of
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -2158,7 +2158,13 @@ do_type_check_expr_in(Env, ResTy, {record, Anno, Name, Fields} = Record) ->
             {VarBinds, Cs2} = type_check_fields(Env, Rec, Fields),
             {VarBinds, constraints:combine(Cs1, Cs2)};
         {fields_tys, Tyss, Cs1} ->
-            todo;
+            case type_check_record_union_in(Env, Tyss, Fields) of
+                none ->
+                    {Ty, _VB, _Cs} = type_check_expr(Env#env{infer = true}, Record),
+                    throw({type_error, Record, Ty, ResTy});
+                {VBs, Cs2} ->
+                    {union_var_binds(VBs, Env#env.tenv), constraints:combine(Cs1, Cs2)}
+            end;
         any ->
             Rec = get_record_fields(Name, Anno, Env#env.tenv),
             type_check_fields(Env, Rec, Fields);
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1695,6 +1695,14 @@ type_check_fields(Env, Rec, Fields) ->
     UnAssignedFields = get_unassigned_fields(Fields, Rec),
     type_check_fields(Env, Rec, Fields, UnAssignedFields).
 
+%% type_check_fields for multiple cases, split by function head
+%% 1. The field is present with a value, need to make sure that field type checks that value
+%% 2. There is a _ present with a value, need to make sure all the remaining fields type checks that value
+%% 3. If case 2 was present, we have type checked all the fields and we can stop
+%% 4. If case 2 was absent, we need to make sure all the fields that were not type checked support:
+%%      In the case of having a default value: putting the default value
+%%      In the case of not having a default value: that it supports `undefined`
+%% 5. If case 2 was absent, we have type checked all the unassigned fields.
 type_check_fields(Env, Rec, [{record_field, _, {atom, _, _} = FieldWithAnno, Expr} | Fields]
                  ,UnAssignedFields) ->
     FieldTy = get_rec_field_type(FieldWithAnno, Rec),
@@ -4310,6 +4318,7 @@ add_var_binds(VEnv, VarBinds, TEnv) ->
     Glb = fun(_K, Ty1, Ty2) -> {Ty, _C} = glb(Ty1, Ty2, TEnv), Ty end,
     gradualizer_lib:merge_with(Glb, VEnv, VarBinds).
 
+%% From a record field name, find the default value from the typed list of record field definitions
 get_rec_field_default({atom, _, FieldName},
                     [{typed_record_field,
                         {record_field, _, {atom, _, FieldName}, Default}, _}|_]) ->
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -51,6 +51,9 @@
 -define(type(T, A), {type, _, T, A}).
 -define(top(), {remote_type, _, [{atom,_,gradualizer}
 				,{atom,_,top},[]]}).
+-define(record_field(Name), {record_field, _, {atom, _, Name}, _}).
+-define(typed_record_field(Name), {typed_record_field, ?record_field(Name), _}).
+-define(typed_record_field(Name, Type), {typed_record_field, ?record_field(Name), Type}).
 
 %% Data collected from epp parse tree
 -record(parsedata, {
@@ -238,6 +241,12 @@ compat_ty({type, P1, record, [{atom, _, Name}]},
     Fields2 = get_maybe_remote_record_fields(Name, P2, TEnv),
     compat_record_fields(Fields1, Fields2, A, TEnv);
 
+%% Records that have been refined on one side or the other
+compat_ty({type, Anno1, record, [{atom, _, Name}|Fields1]},
+          {type, Anno2, record, [{atom, _, Name}|Fields2]}, A, TEnv) ->
+    AllFields1 = case Fields1 of [] -> get_record_fields_types(Name, Anno1, TEnv); _ -> Fields1 end,
+    AllFields2 = case Fields2 of [] -> get_record_fields_types(Name, Anno2, TEnv); _ -> Fields2 end,
+    compat_record_tys(AllFields1, AllFields2, A, TEnv);
 compat_ty({type, _, record, _}, {type, _, tuple, any}, A, _TEnv) ->
     ret(A);
 
@@ -296,6 +305,17 @@ compat_tys([Ty1|Tys1], [Ty2|Tys2], A, TEnv) ->
     {Aps, Css} = compat_tys(Tys1, Tys2, Ap, TEnv),
     {Aps, constraints:combine(Cs, Css)};
 compat_tys(_Tys1, _Tys2, _, _) ->
+    throw(nomatch).
+
+
+compat_record_tys([], [], A, _TEnv) ->
+    ret(A);
+compat_record_tys([Field1|Fields1], [Field2|Fields2], A, TEnv) ->
+    {A1, Cs1} = compat(Field1, Field2, A, TEnv),
+    {A2, Cs2} = compat_record_tys(Fields1, Fields2, A1, TEnv),
+    {A2, constraints:combine(Cs1, Cs2)};
+compat_record_tys(_, _, _, _) ->
+    %% Mismatching number of fields
     throw(nomatch).
 
 %% Two records are compatible if they have the same name (defined in different
@@ -596,6 +616,9 @@ glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
 %% * Replace built-in type synonyms
 %% * Flatten unions and merge overlapping types (e.g. ranges) in unions
 -spec normalize(type(), TEnv :: #tenv{}) -> type().
+normalize({type, _, record, [{atom, _, Name}|Fields]}, TEnv) when length(Fields) > 0 ->
+    NormFields = [normalize(Field, TEnv) || Field <- Fields],
+    type_record(Name, NormFields);
 normalize({type, _, union, Tys}, TEnv) ->
     Types = flatten_unions(Tys, TEnv),
     case merge_union_types(Types, TEnv) of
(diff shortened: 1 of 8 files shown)


src/typechecker.erl
```diff
@@ -2079,49 +2130,78 @@ do_type_check_expr_in(Env, ResTy, {map, _, Expr, Assocs} = MapUpdate) ->
...
+                            {type, _, record, [{atom, _, Name}|RefinedFields]} ->
+                                {VarType, lists:nth(FieldIndex-1, RefinedFields)};
+                            _ ->
+                                throw({type_error, mismatch, type_record(Name), Expr})
+                        end;
+                    _ ->
+                        throw({unknown_variable, P, Var})
+                end;
+            _ -> {type_record(Name), FieldTy}
+        end,
```
FrankBro: This is so ugly @_@

src/typechecker.erl
```diff
@@ -3664,12 +3778,23 @@ add_type_pat({bin, _P, BinElements} = Bin, Ty, TEnv, VEnv, Caps) ->
...
-        type_error -> throw({type_error, record_pattern, P, Record, Ty});
-        {ok, Cs1} ->
-            {VEnv2, Cs2} = add_type_pat_fields(Fields, Record, TEnv, VEnv, Caps),
-            RecTy = {type, erl_anno:new(0), record, [{atom, erl_anno:new(0), Record}]},
-            {type(none), RecTy, VEnv2, constraints:combine(Cs1, Cs2)}
+    case expect_record_type(Ty, Record, TEnv) of
+        any ->
+            %% FIXME: If record() is used and we try to bind, every field types
+            %% - should be any(). add_any_types_pat won't work, will need a
+            %% - add_any_type_pat_fields
```
josefs: I think it'd be good to get this fixed before merging.

src/typechecker.erl
josefs: Yeah, it'd be good to generalize this.

One way to perhaps solve this would be to change `type_check_expr_in` so that it could also return a type, a type that was a refinement, or a subtype, of the type that was passed as an argument. That way we could call `type_check_expr_in` on the expression and get back the refined type of the record and return the appropriate field.

I'm not asking you to fix this now, that's quite a big change. I'm just thinking out loud.

src/typechecker.erl
```diff
@@ -1115,33 +1135,68 @@ expect_fun_type_union(Env, [Ty|Tys]) ->
...
 
-expect_record_type(Record, {type, _, record, [{atom, _, Name}]}, _TEnv) ->
-    if Record == Name ->
-         {ok, constraints:empty()};
-       true ->
-         type_error
+-spec expect_record_type(type(), atom(), #tenv{}) -> Any | ElemTy | ElemTys | TypeError when
+    Any :: any,
+    ElemTy :: {elem_ty, [typed_record_field()], constraints:constraints()},
+    ElemTys :: {elem_tys, [[typed_record_field()]], constraints:constraints()},
```
josefs: The `elem_ty` name is perhaps not the best choice here. The reason I chose that name was to refer to the type of the elements of lists and tuples. But in this case I think it's more appropriate to call it `fields_ty`.

src/typechecker.erl
zuiderkwast: > One way to perhaps solve this would be to change type_check_expr_in so that it could also return a type

If we do that, can we also unify `type_check_expr` with `type_check_expr_in`? (with `any()` as the expected type)

src/typechecker.erl
FrankBro: For sure, I just wanted to create the PR for early review, there is still a bit of work that needs to be done.

src/typechecker.erl
josefs: Ha, yes of course! That's a really good point @zuiderkwast ! I think that would be a very worthwhile rewrite!

So one invariant would be that the type that is returned is always a subtype of the type passed in from the context. There's an interesting corner case wrt `any()`: if the context demands a concrete type that is not `any()` but we infer the type `any()` what should we return? It should always be safe to return the concrete type, I think.

It's a bit daunting to think about rewriting the typechecker but I think it'll be great to do it. I'm not going to jump on it right away because I'm working on the constraint solver right now (branch `constraint_solver` for anyone interested).  So if anyone wants to take this on right now, that'd be really cool.

zuiderkwast: If we keep the current behaviour, is should be fine, right? We have a lot of test cases so I think attempting this kind of refactoring is pretty safe.

josefs: Yes, I agree. We have a pretty good test suite at this point.

src/typechecker.erl
```diff
@@ -2079,49 +2134,78 @@ do_type_check_expr_in(Env, ResTy, {map, _, Expr, Assocs} = MapUpdate) ->
...
+                ,constraints:combine([Cs1, Cs2|Css])};
+        {elem_tys, Tyss, Cs1} ->
+            case type_check_record_union_in(Env, Tyss, Fields) of
+                none ->
+                    {Ty, _VB, _Cs} = type_check_expr(Env#env{infer = true}, Record),
+                    throw({type_error, Record, Ty, ResTy});
+                {VBs, Cs2} ->
+                    {union_var_binds(VBs, Env#env.tenv), constraints:combine(Cs1, Cs2)}
+            end;
+        any -> fuck;
```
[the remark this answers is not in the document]
FrankBro: Ouch that's a big typo from `todo` 😅 . Fixed until I actually implement it.

src/typechecker.erl
FrankBro: Renamed

src/typelib.erl
```diff
@@ -48,6 +48,9 @@ pp_type(Type = {type, _, bounded_fun, _}) ->
 pp_type({var, _, TyVar}) ->
     %% TODO: In type(), TyVar should be an atom but we use a string.
     TyVar;
+pp_type({type, TypeAnno, record, [{atom, AtomAnno, Name}|Fields]})
+  when length(Fields) /= 0 ->
+    pp_type({type, TypeAnno, record, [{atom, AtomAnno, Name}]});
```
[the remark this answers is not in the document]
FrankBro: Just removed it. Most likely was crashing when `pp_type` was used with malformed extra fields.

src/typechecker.erl
```diff
@@ -238,6 +241,12 @@ compat_ty({type, P1, record, [{atom, _, Name}]},
     Fields2 = get_maybe_remote_record_fields(Name, P2, TEnv),
     compat_record_fields(Fields1, Fields2, A, TEnv);
 
+%% Records that have been refined on one side or the other
+compat_ty({type, Anno1, record, [{atom, _, Name}|Fields1]},
+          {type, Anno2, record, [{atom, _, Name}|Fields2]}, A, TEnv) ->
+    AllFields1 = case Fields1 of [] -> get_record_fields_types(Name, Anno1, TEnv); _ -> Fields1 end,
+    AllFields2 = case Fields2 of [] -> get_record_fields_types(Name, Anno2, TEnv); _ -> Fields2 end,
```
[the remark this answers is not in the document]
FrankBro: Changed it in `get_record_fields_types`. `get_maybe_remote_record_fields`, along with `get_record_fields_types`, return `typed_record_field`s. All I want is the type, not all the extra info such as name, default value. Should support remote records now. 👍

src/typechecker.erl
```diff
@@ -298,6 +307,17 @@ compat_tys([Ty1|Tys1], [Ty2|Tys2], A, TEnv) ->
...
+
+compat_record_tys([], [], A, _TEnv) ->
+    ret(A);
+compat_record_tys([Field1|Fields1], [Field2|Fields2], A, TEnv) ->
+    {A1, Cs1} = compat(Field1, Field2, A, TEnv),
+    {A2, Cs2} = compat_record_tys(Fields1, Fields2, A1, TEnv),
+    {A2, constraints:combine(Cs1, Cs2)};
+compat_record_tys(_, _, _, _) ->
+    %% Mismatching number of fields
+    throw(nomatch).
```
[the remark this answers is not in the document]
FrankBro: `compat_record_fields` uses the `typed_record_field` form whereas this one only uses the `type` form.

test/known_problems/should_pass/record_refinement.erl
```diff
@@ -0,0 +1,9 @@
+-module(record_refinement).
+
+-record(one_field, { a :: integer() | undefined }).
+
+-spec one_field3(#one_field{}, integer()) -> integer().
+one_field3(#one_field{a = I}, I) -> I;
```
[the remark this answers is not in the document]
FrankBro: Moved that test in the `pattern_bind_reuse` one that me and @zuiderkwast are using.

src/typechecker.erl
```diff
@@ -1115,33 +1135,77 @@ expect_fun_type_union(Env, [Ty|Tys]) ->
...
-    {ok, constraints:empty()};
-expect_record_type(Record, {type, _, union, Tys}, TEnv) ->
-    expect_record_union(Record, Tys, TEnv);
-expect_record_type(Record, {var, _, Var}, _TEnv) ->
-    {ok, constraints:add_var(Var,
-           constraints:upper(Var, type_record(Record)))};
-expect_record_type(_, _, _) ->
-    type_error.
+expect_record_type(?top() = _TermTy, _Record, _TEnv) ->
+    any;
```
FrankBro: Not sure about this here. For tuples, we create a tuple of N elements of `TermTy`. Should I just use the base type of the record fields?

src/typechecker.erl
```diff
@@ -1115,33 +1135,77 @@ expect_fun_type_union(Env, [Ty|Tys]) ->
...
+%%        ,{var, erl_anno:new(0), TyVar}
+%%        ,constraints:add_var(Var,
+%%            constraints:upper(Var, type_record(Record)))
+%%    };
+    case REnv of
+        #{Record := Fields} ->
+            {fields_ty, Fields, constraints:empty()};
+        _NotFound ->
+            {type_error, Record}
+    end;
```
FrankBro: Not sure if this is okay. The tuple version creates a new type variable. Since we know the name of the record, I just return the base type of the record fields.

src/typechecker.erl
josefs: The record fields should have type `top()` in this case as well.

src/typechecker.erl
josefs: It's important that you constrain the type variable just as in the case with tuples. Using the declared types of the fields should be fine.

src/typechecker.erl
josefs: Actually, on second thought, you're completely correct to use the fields types of the record declaration.

src/typechecker.erl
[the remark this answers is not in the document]
josefs: Ah, right.

src/typechecker.erl
FrankBro: 👍

src/typechecker.erl
```diff
@@ -2079,49 +2146,82 @@ do_type_check_expr_in(Env, ResTy, {map, _, Expr, Assocs} = MapUpdate) ->
...
+                    #{Var := VarType} ->
+                        case VarType of
+                            {type, _, record, [{atom, _, Name}]} ->
+                                {VarType, FieldTy};
+                            {type, _, record, [{atom, _, Name}|RefinedFields]} ->
+                                {VarType, lists:nth(FieldIndex-1, RefinedFields)};
+                            {var, _, _} ->
+                                {type_record(Name), FieldTy};
+                            ?type(any) ->
+                                {type_record(Name), FieldTy};
```
FrankBro: This is becoming really ugly

src/typechecker.erl
```diff
@@ -3200,13 +3310,33 @@ refine(OrigTy, Ty, TEnv) ->
...
     %% because there is imprecision.
     throw(no_refinement);
 refine_ty(?type(T, A), ?type(T, A), _) ->
     type(none);
+refine_ty(?type(record, [{atom, _, Name}]), ?type(record, [{atom, _, Name}]), _) ->
+    type(none);
+refine_ty(?type(record, [{atom, Anno, Name}]), Refined = ?type(record, [{atom, _, Name} | _]), TEnv) ->
+    refine_ty(expand_record(Name, Anno, TEnv), Refined, TEnv);
+refine_ty(Refined = ?type(record, [{atom, _, Name} | _]), ?type(record, [{atom, Anno, Name}]), TEnv) ->
+    refine_ty(Refined, expand_record(Name, Anno, TEnv), TEnv);
```
zuiderkwast: Shouldn't this be none()? If not, why not?

src/typelib.erl
```diff
@@ -37,6 +37,8 @@ pp_type({type, _, bounded_fun, [FunType, []]}) ->
     %% Bounded fun with empty constraints gets printed with a trailing "when"
     %% when pretty-printed as a spec (next clause)
     pp_type(?assert_type(FunType, function_type()));
+pp_type({type, Anno, record, [Name|Fields]}) when length(Fields) > 0 ->
+    pp_type({type, Anno, record, [Name]});
```
zuiderkwast: Regarding the commit message "erl_pp does not like record fields": have you tried replacing erl_pp with erl_prettypr?

src/typechecker.erl
```diff
@@ -1115,33 +1138,77 @@ expect_fun_type_union(Env, [Ty|Tys]) ->
...
-    expect_record_union(Record, Tys, TEnv);
-expect_record_type(Record, {var, _, Var}, _TEnv) ->
-    {ok, constraints:add_var(Var,
-           constraints:upper(Var, type_record(Record)))};
-expect_record_type(_, _, _) ->
-    type_error.
+expect_record_type(?top() = _TermTy, _Record, _TEnv) ->
+    any;
+expect_record_type({ann_type, _, [_, Ty]}, Record, TEnv) ->
+    expect_record_type(Ty, Record, TEnv);
```
zuiderkwast: It shouldn't be possible to match ann_type here. I have tried to make sure they are removed as early as possible. (Maybe not inside record fields though?)

FrankBro: I just copied this code from the `expect_tuple_type` before it was removed I think. I'll make sure it can't happen and remove it.

src/typechecker.erl
FrankBro: So at this point, we might be the first function clause and the record type hasn't been expanded yet. An example:

```erlang
-record(r, {f :: integer() | undefined}).

-spec test(#r{}) -> integer().
test(#r{f = undefined}) -> 0;
test(#r{f = I}) -> I.
```

At this point, the argument type is `{type, _, record, [{atom, _, r}]}` but the pattern type would be `{type, _, record, [{atom, _, r}, {atom, _, undefined}]}`.

src/typelib.erl
FrankBro: No but I think it might boil down from the fact that there is another notation possible. If I use 

```erlang
1> typelib:parse_type("#r{f :: integer()}").
{type,1,record,
      [{atom,1,r},
       {type,1,field_type,[{atom,1,f},{type,1,integer,[]}]}]}
```

I think what `erl_pp` would like is for me to specify only the fields that have been modified from their base values? Not sure, could test it.

src/typechecker.erl
zuiderkwast: If the first argument to `refine_ty` is not expanded, I agree, it has to be expanded in order to refine it. (Lines 3329-3330)

If the second argument to `refine_ty` is not expanded, I don't agree. This means that the result is none() and the record type doesn't need to be expanded to compute that. Perhaps this second clause isn't even needed...? (Lines 3331-3332)

src/typelib.erl
zuiderkwast: `erl_pp` doesn't care what the record definition is, so it doesn't matter if all the fields are added. It is purely a formatting of the parse tree. It must be something else...

With two fields refined (even without a record definition), it works in the current master:

```Erlang
3> T2 = {type, 1, record, [{atom, 1, r}, {type, 1, field_type, [{atom,1,f},{type,1,integer,[]}]},{type,1,field_type,[{atom,1,g},{type,1,atom,[]}]}]}.
{type,1,record,
      [{atom,1,r},
       {type,1,field_type,[{atom,1,f},{type,1,integer,[]}]},
       {type,1,field_type,[{atom,1,g},{type,1,atom,[]}]}]}
4> typelib:pp_type(T2).                                                                                                                              
"#r{f :: integer(), g :: atom()}"
```

zuiderkwast: It has to be `field_type` here, not <del><code>typed_record_field</code></del> as in the record definition.

src/typechecker.erl
FrankBro: Ah right, that makes sense. First argument is always the least refined right? I will change this.

zuiderkwast: > First argument is always the least refined right?

Well, this is like "minus" for types. If the first argument is a subtype of the second, then the result should be none().

This function is described as a helper for `type_diff`. (I think it would be a good idea to rename it to e.g. `type_diff_help` or `do_type_diff` or so, but we'd better make such refactorings as separate PRs.)

src/typelib.erl
FrankBro: The problem was that it was just extra arguments for the `record` type. Like

```erlang
{type, _, record, [{atom, _, name}, {type, _, integer, _}, {type, _, atom, _}]}
```

FrankBro:
I wonder if instead of expanding, I should use `field_type`. I think I tried but I honestly am not sure cuz I've been working on this on and off.

So

```erlang
-record(r, {a :: integer() | undefined, b :: integer() | undefined}).

-spec test(#r{}) -> integer().
test(#r{b = undefined}) -> 0;
test(#r{b = B}) -> B.
```

Right it would go from

```erlang
{type, _, record, [{atom, _, r}]}
```

to

```erlang
{type, _, record, [{type, _, union, [{type, _, integer, _}, {atom, _, undefined}]}, {type, _, integer, _}]}
```

We had to add the `a` field in there.

If we used `field_type`, maybe we could just specify the fields that have actually been refined. Sounds more elegant but I feel like I tried doing this a gave up. Maybe cuz it required a refactor that was too great. Maybe another PR?

src/typelib.erl
zuiderkwast: OK, resolved then?

zuiderkwast:
> If we used `field_type`, maybe we could just specify the fields that have actually been refined.

Maybe, but I think we should definitely use the syntax produced by the parser and accepted by the pretty-printer, i.e. with `field_type`. It doesn't matter much if we include all fields or just the refined ones. They are semantically the same anyway.

FrankBro:
I think the only annoying thing is that if include all the fields it will display them. I guess we shouldn't if they are the default type.

FrankBro:
Could we keep the changing record expansion to using `field_type` for another PR? This is already getting pretty huge and it would require a non-trivial refactor.

src/typechecker.erl
FrankBro: Fixed 👍

src/typelib.erl
FrankBro: If we keep it like so instead of using `field_type` then yes.

zuiderkwast:
> 
> 
> Could we keep the changing record expansion to using `field_type` for another PR? This is already getting pretty huge and it would require a non-trivial refactor.

Since this is not yet in master, refactoring is not a problem IMO. Then squash and rebase and it will all look fine.

If it's easier, just expand all fields but using `field_type`. Then it's a fairly trivial change, isn't it?

I really don't like to add a representation which doesn't follow the normal representation and breaks pretty-printing in a case where it's not even necessary to extend the type language.

(Even if a clause for pretty-printing is added, it's still a problem if a nested structure is pretty-printed, for example an "expanded" record type inside a tuple.)

FrankBro:
Ok I will do it the proper way.

src/typechecker.erl

src/typelib.erl
FrankBro: Will be removed to use `field_type`

FrankBro:
Alright, every record field refinement is expanded to `{type, _, field_type, [{atom, _, FieldName}, Type]}`. I still expand all the fields for now. I'm trying to get it to lead to ugly error messages but I can't.

src/typechecker.erl
FrankBro: I think it was actually right the way I wrote it. Otherwise it makes other stuff fail. Resolved for now.

FrankBro:
I *think* I'm done. I took care of all my todos. I think the codecov is failing because I'm having a hard time getting in the `fields_tys` and `any` paths for `expect_record_type`. Not sure what other types of tests to add. Would love suggestions.

zuiderkwast:
> I still expand all the fields for now. I'm trying to get it to lead to ugly error messages but I can't.

I haven't tried, but how about something like this:

```Erlang
-record(r, {a :: atom(), b :: atom(),
            c :: c, d :: d, e :: e, f :: f, g :: g, h :: h, i :: i}).
-spec too_much_expanded_error_message(#r{a :: a}) -> #r{a :: a, b :: b}).
too_much_expanded_error_message(R) -> R.
```

FrankBro:
Funny enough, your example does not spit an error but I assumed it was because `a` *IS* an `atom()` so I switched it to `integer()` and got no error. 

I assume I just willy-nilly allow remapping of types. Not sure that's what we want. Maybe I should make sure a refinement is a subtype of the initial definition? Or can this even make sense?

src/typelib.erl
```diff
@@ -92,8 +92,6 @@ remove_pos({Type, _, Value})
 remove_pos({user_type, Anno, Name, Params}) when is_list(Params) ->
     {user_type, anno_keep_only_filename(Anno), Name,
      lists:map(fun remove_pos/1, Params)};
-remove_pos({type, Anno, record, Params = [{atom, AtomAnno, Name}]}) ->
-    {type, anno_keep_only_filename(Anno), record, [{atom, anno_keep_only_filename(AtomAnno), Name}]};
```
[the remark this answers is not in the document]
FrankBro: Because this will take care of it https://github.com/josefs/Gradualizer/blob/master/src/typelib.erl#L108-L109

```erlang
remove_pos({type, _, Type, Params}) when is_list(Params) ->
    {type, erl_anno:new(0), Type, lists:map(fun remove_pos/1, Params)};
```

This also does `field_type` for free.

zuiderkwast:
> Funny enough, your example does not spit an error but I assumed it was because `a` _IS_ an `atom()` so I switched it to `integer()` and got no error.

Ah, right, make `b :: integer()` in the record definition. Then, the incoming R is a record with `b :: integer()` so I expect an error where R is returned, because `b :: b` is expected.

> I assume I just willy-nilly allow remapping of types. Not sure that's what we want. Maybe I should make sure a refinement is a subtype of the initial definition? Or can this even make sense?

Remapping `b :: integer()` to `b :: b` is not very nice, but I don't know if we should forbid it. I wouldn't mind if we forbid it, but I don't think it's needed in this case. I think we should be able to detect the error above anyway.

src/typelib.erl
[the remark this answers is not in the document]
zuiderkwast: But no, there is a difference. For records, `anno_keep_only_filename/1` should be called on the Anno instead of `erl_anno:new(0)`, because we need to remember which module the record came from. (A remote record comes from a remote type which expands to a record type.)

FrankBro:
This is mainly a note for myself. Been very busy at work but we've reserved some times to finish this PR soon enough (along with the map one). Sorry for the lack of contribution recently.

It seems like a bug is still present. If no default value is provided for a record, gradualizer doesn't catch it. For example

```erlang
-record(test, {
    integers :: [integer()]
}).

-spec test(#test{}) -> ok.
test([]) -> ok;
test([_X|Xs]) -> test(Xs).

call() ->
    Test = #test{},
    test(Test).
```

This will pass gradualizer but fail at runtime. If we had provided a default of empty list, obviously this would be right. Gradualizer should've caught the default values.

So get to that future Francois 😄

src/typechecker.erl
```diff
@@ -596,6 +621,11 @@ glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
 %% * Replace built-in type synonyms
 %% * Flatten unions and merge overlapping types (e.g. ranges) in unions
 -spec normalize(type(), TEnv :: #tenv{}) -> type().
+normalize({type, _, record, [{atom, _, Name}|Fields]}, TEnv) when length(Fields) > 0 ->
+    NormFields = [normalize(Field, TEnv) || Field <- Fields],
+    type_record(Name, NormFields);
+normalize(?type_field_type(_Name, Type), TEnv) ->
+    normalize(Type, TEnv);
```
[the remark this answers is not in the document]
FrankBro: Done.

src/typechecker.erl
```diff
@@ -3208,13 +3325,36 @@ refine(OrigTy, Ty, TEnv) ->
...
+refine_ty(?type(record, [Name|Tys1]), ?type(record, [Name|Tys2]), TEnv)
+  when length(Tys1) > 0, length(Tys1) == length(Tys2) ->
+    % Record without just the name
+    RefTys = [refine(Ty1, Ty2, TEnv) || {Ty1, Ty2} <- lists:zip(Tys1, Tys2)],
+    RecordsElems = pick_one_refinement_each(Tys1, RefTys),
+    Records = [type(record, [Name|RecordElems]) || RecordElems <- RecordsElems],
+    normalize(type(union, Records), TEnv);
+refine_ty(?type_field_type(Name, Ty1), ?type_field_type(Name, Ty2), TEnv) ->
+    RefTy = refine(Ty1, Ty2, TEnv),
+    normalize(type_field_type(Name, RefTy), TEnv);
```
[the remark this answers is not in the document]
FrankBro: Also removed

src/typechecker.erl
```diff
@@ -298,6 +312,17 @@ compat_tys([Ty1|Tys1], [Ty2|Tys2], A, TEnv) ->
...
+
+compat_record_tys([], [], A, _TEnv) ->
+    ret(A);
+compat_record_tys([Field1|Fields1], [Field2|Fields2], A, TEnv) ->
+    {A1, Cs1} = compat(Field1, Field2, A, TEnv),
+    {A2, Cs2} = compat_record_tys(Fields1, Fields2, A1, TEnv),
+    {A2, constraints:combine(Cs1, Cs2)};
+compat_record_tys(_, _, _, _) ->
+    %% Mismatching number of fields
+    throw(nomatch).
```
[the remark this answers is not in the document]
FrankBro: What would you suggest be here? One thing is that subtype especially catches `nomatch`. I guess it could catch more?

src/typechecker.erl
```diff
@@ -3208,13 +3325,36 @@ refine(OrigTy, Ty, TEnv) ->
...
     %% because there is imprecision.
     throw(no_refinement);
 refine_ty(?type(T, A), ?type(T, A), _) ->
     type(none);
+refine_ty(?type(record, [{atom, _, Name}]), ?type(record, [{atom, _, Name}]), _) ->
+    type(none);
+refine_ty(?type(record, [{atom, Anno, Name}]), Refined = ?type(record, [{atom, _, Name} | _]), TEnv) ->
+    refine_ty(expand_record(Name, Anno, TEnv), Refined, TEnv);
+refine_ty(?type(record, [{atom, _, Name} | _]), ?type(record, [{atom, _, Name}]), _TEnv) ->
+    type(none);
```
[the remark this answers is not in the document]
FrankBro: Fixed. Good catch.

src/typelib.erl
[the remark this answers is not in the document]
FrankBro: One point where I'm not sure about this. Let's say you have a record from a header file. Right now this PR support this case

```erlang
% file.hrl
-record(record, { field :: integer() | undefined }).
```
```erlang
% file.erl
-include("file.hrl").

-spec safe(#record{ field :: integer() }) -> integer().
safe(#record{ field = I }) -> I.

-spec unsafe(#record{}) -> integer().
unsafe(#record{ field = undefined }) -> 0;
unsafe(R) -> safe(R).
```

This use case is pretty useful for me because there are placed where some checks are done upstream. If we keep the filename, wouldn't that not work anymore? I tried readding this (and supporting `field_type` for fields but it makes things fail because I guess I do an explicit file check. Should we remove the filename in `remove_pos` or I can try to find who looks for the specific file check in `typechecker.erl`.

src/typechecker.erl
```diff
@@ -4314,6 +4315,15 @@ add_var_binds(VEnv, VarBinds, TEnv) ->
     Glb = fun(_K, Ty1, Ty2) -> {Ty, _C} = glb(Ty1, Ty2, TEnv), Ty end,
     gradualizer_lib:merge_with(Glb, VEnv, VarBinds).
 
+get_rec_field_default({atom, _, FieldName},
```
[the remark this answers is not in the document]
FrankBro: I've added a comment, will let you judge if this is resolved.

src/typechecker.erl
```diff
@@ -1718,7 +1718,8 @@ type_check_fields(_Env, _Rec, [], should_not_be_inspected) ->
     {#{}, constraints:empty()};
 type_check_fields(Env, Rec, [], [UnAssignedField|UnAssignedFields]) ->
     FieldTy = get_rec_field_type({atom, erl_anno:new(0), UnAssignedField}, Rec),
-    {VB1, Cs1} = type_check_expr_in(Env, FieldTy, {atom, erl_anno:new(0), undefined}),
+    FieldDefault = get_rec_field_default({atom, erl_anno:new(0), UnAssignedField}, Rec),
```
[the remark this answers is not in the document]
FrankBro: I've added a comment, will let you judge if this is resolved. Also renamed like you wished.

src/typechecker.erl
FrankBro: Seems I found a test case that require implementing this. As soon as you have two records embedded, this fails because it only checks for a variable. This examples cannot pass now.

```erlang
-record(r2, {value :: undefined | integer()}).
-record(r1, {r2 :: undefined | #r2{}}).

-spec test(#r1{}) -> integer().
test(#r1{r2 = undefined}) -> 
    0;
test(#r1{r2 = #r2{value = undefined}}) -> 
    0;
% Works
% test(#r1{value = #r2{value = Value}}) ->
%     Value.
test(R1) ->
    R1#r1.r2#r2.value.
```

FrankBro: Alright I think I fixed this. I went the `type_check_expr` route instead of `type_check_expr_in`. Eveything passes and the new test case passes. It also fixed a case in our codebase 👍

FrankBro: If you guys are fine with it, I would resolve this comment. The idea of refactoring for something cleaner could be tackled later.

src/typechecker.erl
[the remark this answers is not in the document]
zuiderkwast: Very good :-)

src/typelib.erl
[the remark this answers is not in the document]
zuiderkwast: It would be quite good if the error messages would contain the hrl file as the filename as you say, yes. I don't know if it works or not. What I remember is that I once made gradualizer_db add "module.erl" as the file when fetching records from remote modules. This happens if an exported type expands to a record defined in the other module.

I guess the only way to test it would be to check the output of a test case where the error message is referring to a record defined in another module.

zuiderkwast:
Let's merge this whenever you feel you're done. :-)

FrankBro:
Once this is merged. Would you be okay with starting to tag Gradualizer? I feel this is a good start to start releasing tags so that specifying versions in rebar is easier. I would suggest tagging from this point on `1.0.0` and we can then discuss what should constitute a minor or fix.

zuiderkwast:
> Once this is merged. Would you be okay with starting to tag Gradualizer? I feel this is a good start to start releasing tags so that specifying versions in rebar is easier. I would suggest tagging from this point on `1.0.0` and we can then discuss what should constitute a minor or fix.

We have talked about tagging a first *beta*, i.e. 0.1.0. Sure, I'd be fine with that. For 1.0.0, i think there are more things that we should have in place, to be able to claim we have a sound gradual type system.

FrankBro:
I'm fine with 0.1.0 as well.

src/typelib.erl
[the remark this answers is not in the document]
FrankBro: I've readded the case but kept it commented. I think we'll keep it for future work because I want to start integrating gradualizer in our codebase to try and find more bugs. I will resolve this for now.

FrankBro:
Finally merging this 🙌 
Will create a tagged version of 0.1.0 on master for this.
Thanks for the help guys.

Issue #270, opened 2020-02-05
Smart matching doesn't influence type of variables


FrankBro:
This should work, erlang makes sure both are the same integer value in this situation. 

```erlang
-module(test).

-spec test(integer() | undefined, integer()) -> integer().
test(I, I) -> I;
test(undefined, I) -> I;
test(I, _) -> I.
```

```
The variable on line 4 at column 9 is expected to have type integer() but it has type integer() | undefined

-spec test(integer() | undefined, integer()) -> integer().
test(I, I) -> I;
        ^
```

josefs:
That's a nice example, I hadn't thought about this problem before. It's tricky to deal with given how the typechecker currently works, but I agree that we could handle this better. Please submit a PR that adds this example under known problems so we can keep track of it.

I'm curious, is this a problem you've come across in practice?

FrankBro:
Nah I just thought about testing it while finishing my record refinement branch (which I hope to create a PR for tomorrow 🤞)

PR #274, opened 2020-02-08
More known problems for bound variable patterns

zuiderkwast:
I looked into #270 and at first I thought it seemed fairly easy to fix.

The idea: When a pattern is matched against a bound variable, instead of checking that VarTy is a subtype of PatternTy, we should refine the variable type using `NewVarTy = glb(VarTy, PatternTy)`. Does this make sense?

(The current code handling this is here: https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L3530).

This however leads to two other problems:

1. Since the first occurrence of a free variable in a pattern is treated as a catch-all, the refinement kicks in an says that the second clause is unreachable.

   ```Erlang
   -spec test(integer() | undefined, integer()) -> integer().
   test(I, I) -> I + I;
   test(_, I) -> I.
   ```

   In @FrankBro's test case above, the first occurrence of `I` is treated as a free variable and the second occurrence of `I` is treated as a bound variable.

   This PR adds a test case (`is_same/2`) isolating this problem.

2. Since `glb(integer(), any()) => any()`, changing to GLB makes the following should_fail test case pass: (https://github.com/josefs/Gradualizer/blob/master/test/should_fail/var_pattern_bound.erl)

   ```Erlang
   -spec var_as_pattern(atom()) -> integer().
   var_as_pattern(Atom) ->
       case get_any() of
           Atom ->
	       %% at this point still Atom :: atom()
	       %% and not Atom :: any()
               Atom
       end.
   ```

   If we think that using GLB for refining the variable type is a good idea, perhaps we will have to accept that the above example passes. In this PR I have added `guess_the_die/1` which illustrates that any() is a better choice than integer() for the variable type in this situation.

### DIFF
--- a/test/known_problems/should_pass/pattern_bind_reuse.erl
+++ b/test/known_problems/should_pass/pattern_bind_reuse.erl
@@ -1,7 +1,29 @@
 -module(pattern_bind_reuse).
 
--compile(export_all).
+-export([test/2, guess_the_die/1, is_same/2]).
 
 -spec test(integer() | undefined, integer()) -> integer().
 test(I, I) -> I + I;
 test(_, I) -> I.
+
+%% Guess a number. Roll a die. If same, you win the points of the die.
+-spec guess_the_die(Guess :: integer()) -> Score :: 0..6.
+guess_the_die(Guess) ->
+    %% Match any() against a bound variable of type integer()
+    case roll_die() of
+        Guess ->
+            Guess; % :: integer(), but should be any() & integer()
+        _WrongGuess ->
+            0
+    end.
+
+%% Untyped helper; returns 1..6
+roll_die() ->
+    random:uniform(6).
+
+-spec is_same(integer(), integer()) -> boolean().
+is_same(N, N) ->
+    true;
+is_same(_, _) ->
+    %% False error: This clause can't be reached
+    false.


josefs:
I think your idea makes a lot of sense. Though I think it's possible to refine it a bit; see below.

My comments on your two point:

1. I think the fundamental problem here is that we impose a particular order on the way that we check variables in patterns. We should really process them "at the same time". We should be able to do that by setting the type of a variable to a typevariable. Then for each occurrence of the variable we would require that its typevariable was a subtype of the type imposed by the typesignature or scrutenee depending on what kind of pattern it is. 

2. It'd be good to look in the literature for what definition of glb they've used. It should fall out from the framework presented in the paper "Abstracting Gradual Typing". It's too late right now for me to work that out.

zuiderkwast:
1. Setting the type of a variable to a type variable doesn't solve the issue that it's treated as match-all and therefore the next clause is reported as "can't be reached", or does it?

   I have another idea: Before checking a pattern, we search though the pattern for variables which occurs more than once and set their types to top() in the TEnv. Then, all the occurrences would be treated as bound and their types are refined. This prevents them from being treated as catch-all. I will have a go on this another day.......

2. I can't find anything useful about GLB this in "Abstracting Gradual Typing". It doesn't cover union and intersection types. It covers precision though (more precise = less unknown).

   We already have the understanding that `glb(T1, T2) = T1 & T2`. Whether `integer() & any()` is `integer()`, `any()` or something in between the two is also something we have already talked about and concluded that `any()` is the best we can do for now. (I think the paper on gradual types with union and intersection types has the most exact type for this, but implementing it would be a huge effort. I still like the keep idea for the far future, but not until there is a paper which combines polymorphism with this approach.)

   Anyway, I refining any() with is_integer/1 (e.g. `case get_any() of X when is_integer(X) -> ... end`) should have the result as refining integer() with any() (like in `guess_the_die/1` in this PR).

PR #275, opened 2020-02-10
Pattern bind reuse

### LINKED ISSUE #270: Smart matching doesn't influence type of variables

zuiderkwast:
Fixes #270 in the way outlined in #274.

TL;DR: Look at the test cases.

The capture/bind vars logic is revamped again. It is moved to an earlier stage, where all variables are extracted from the list of patterns. Therefore, Caps doesn't need to be passed around in add_type_pat/4 anymore.

* If a free variable is only occuring once in the pattern, it is handled as a free variable, i.e. it is not included in VEnv. (For bind_vars, any such variable is deleted from VEnv, to make sure it's treated as free later.)
* If a free variable occurs more than once in a pattern, it is bound to top(), which is later refined in add_type_pat/4 and add_any_types_pat/2. (For bind_vars, such variables are re-assigned to top() if they are bound previously, which is later refined in add_type_pat/4.)


zuiderkwast:
Build fails because there is a crash in `make gradualize`. I'll look into it.

test/should_pass/intersection_with_any.erl
```diff
@@ -10,3 +11,16 @@ any_refined_using_guard(X) when is_integer(X) -> X.
...
+-spec guess_two_dice(integer(), integer()) -> 0..6.
+guess_two_dice(X, Y) ->
+    case roll_two_dice() of
+        {X, Y} ->
+            X;
+        _GuessFail ->
+            0
+    end.
+
+%% untyped heper returning {1..6, 1..6}
```
josefs: "helper"

test/known_problems/should_fail/intersection_with_any_fail.erl
```diff
@@ -14,3 +16,26 @@ any_refined_using_guard(X) when is_atom(X) ->
...
+        {Atom} ->
+            %% at this point Atom :: any()
+            %% but we want Atom :: atom() & any()
+            Atom
+    end.
+
+%% helper
+-spec get_any() -> any().
+get_any() ->
+    receive Any -> Any end.
```
josefs: Nice tests!

src/typechecker.erl
```diff
@@ -3499,52 +3499,69 @@ type_check_function(Env, {function,_, Name, NArgs, Clauses}) ->
...
 %% by Pat. UBound is Ty or a subtype such that Pat :: UBound.
 add_types_pats(Pats, Tys, TEnv, VEnv, Caps) ->
-    add_types_pats(Pats, Tys, TEnv, VEnv, Caps, [], [], []).
-
-add_types_pats([], [], _TEnv, VEnv, _Caps, PatTysAcc, UBoundsAcc, CsAcc) ->
+    NewVEnv = assign_types_to_vars_bound_more_than_once(Pats, VEnv, Caps),
+    do_add_types_pats(Pats, Tys, TEnv, NewVEnv).
+
+%% Used in recursive calls from add_type_pat/4
+-spec do_add_types_pats(Pats :: [gradualizer_type:abstract_pattern()],
```
josefs: As I understand it, this function should not be called directly, all code should go via `add_types_pats` right? Can I suggest you put a comment on `add_types_pats` to highlight this fact

src/typechecker.erl
```diff
@@ -3499,52 +3499,69 @@ type_check_function(Env, {function,_, Name, NArgs, Clauses}) ->
...
-		false ->
-		    throw({type_error, Var, VarTy, Ty})
+	#{A := VarTy} ->
+	    case glb(VarTy, Ty, TEnv) of
+                {?type(none), _Cs} ->
+                    %% TODO: Better type error (it's a pattern, not an
+                    %% expression)
+		    throw({type_error, Var, VarTy, Ty});
+		{RefinedTy, Cs} ->
+		    {type(none), RefinedTy, VEnv#{A := RefinedTy}, Cs}
```
josefs: It's nice to be able to get rid of the `capture_vars` logic here.

src/typechecker.erl
```diff
@@ -3863,6 +3880,51 @@ add_any_types_pat({op, _, _Op, _Pat}, VEnv) ->
...
     VEnv.
 
+-spec assign_types_to_vars_bound_more_than_once(Pats :: [gradualizer_type:abstract_pattern()],
+                                                VEnv :: map(),
+                                                Caps :: capture_vars | bind_vars) ->
+                                                       NewVEnv :: map().
+%% Assigns the type top() to variables occurring more than once in the list of
+%% patterns. This prevents the variables from being treated as match-all, thus
+%% affecting refinement.
+assign_types_to_vars_bound_more_than_once(Pats, VEnv, Caps) ->
```
josefs: Giving variables the type `top()` feels a bit like a sledgehammer. I have the sense that we should be able to give them better types and refine the match-all logic to treat these variables better. But I don't have a concrete idea of how to do that.

test/known_problems/should_fail/intersection_with_any_fail.erl
zuiderkwast: I just moved them from another test file. :-) They were failing as they should before but now they are known problems.

src/typechecker.erl
zuiderkwast: :+1:

src/gradualizer_lib.erl
```diff
@@ -119,3 +119,37 @@ pick_value(?type(range, [{_TagLo, _, neg_inf}, {_TagHi, _, Hi}])) ->
...
+fold_ast(_Fun, Acc, _LiteralEtc) ->
+    Acc.
+
+%% Returns the children of an AST node
+get_ast_children({clauses, Clauses}) ->
+    %% This one doesn't have an annotation
+    Clauses;
+get_ast_children(Node) ->
+    [_Tag, _Anno | Children] = tuple_to_list(Node),
+    Children.
```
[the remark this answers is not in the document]
zuiderkwast: You mean `{clauses, _}` is sneaky, right? Not that I sneakily copied this function from gradualizer_highlight? ;-) It should probably only be defined in one place....

src/typechecker.erl
zuiderkwast: Good idea.

Btw, do we put the description above or below the spec?

src/typechecker.erl
zuiderkwast: We known that these variables occurs in the pattern so we know tha they will be refined using `glb(Ty, top())`, which will be Ty in all cases. That's why top() is chosen here.

test/should_pass/intersection_with_any.erl
zuiderkwast: Done

FrankBro:
Wow that was fast. Awesome 🙌

PR #276, opened 2020-02-10
Refine type for opaque (remain user_type)

FrankBro:
Allow opaque types to further refinement.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3268,6 +3268,10 @@ refine_ty({Tag1, _, M}, {Tag2, _, N}, _TEnv)
 refine_ty(Ty1, Ty2, _TEnv) when ?is_int_type(Ty1),
                                 ?is_int_type(Ty2) ->
     gradualizer_int:int_type_diff(Ty1, Ty2);
+refine_ty({user_type, Anno, Name, Args}, {user_type, Anno, Name, Args}, _TEnv) ->
+    % After being normalized, it's because it's defined as opaque.
+    % If it has the same annotation, name and args, it's the same.
+    type(none);
 refine_ty(Ty1, Ty2, TEnv) ->
     case glb(Ty1, Ty2, TEnv) of
         {?type(none), _}  -> throw(disjoint);     %% disjoint
--- a/test/should_pass/opaque.erl
+++ b/test/should_pass/opaque.erl
@@ -1,0 +1,13 @@
+-module(opaque).
+
+-export([external/2, internal/2]).
+
+-spec external(user_types:my_opaque(), integer() | undefined) -> integer().
+external(_, undefined) -> 0;
+external(_, I) -> I.
+
+-opaque my_opaque() :: integer().
+
+-spec internal(my_opaque(), integer() | undefined) -> integer().
+internal(_, undefined) -> 0;
+internal(_, I) -> I.


FrankBro:
Not sure if I can create a failing test case. I expect it to fail if the `Anno`, `Name` or `Args` are different.

FrankBro:
Good to merge?

josefs:
> Not sure if I can create a failing test case. I expect it to fail if the Anno, Name or Args are different.

You could create a failing test case by trying to use `user_types:my_opaque()` as an `integer()`. I think that'd be a good addition to this PR.

test/should_pass/opaque.erl
```diff
@@ -0,0 +1,13 @@
...
+-export([external/2, internal/2]).
+
+-spec external(user_types:my_opaque(), integer() | undefined) -> integer().
+external(_, undefined) -> 0;
+external(_, I) -> I.
+
+-opaque my_opaque() :: integer().
+
+-spec internal(my_opaque(), integer() | undefined) -> integer().
+internal(_, undefined) -> 0;
```
Zalastax: I'd expect a test that the type is still visible from inside the module, i.e. that the first parameter is used in this function.

FrankBro:
Sorry I guess I was in a rush a bit 😅 . Will try to include these test cases tomorrow.

src/typechecker.erl
```diff
@@ -3268,6 +3268,10 @@ refine_ty({Tag1, _, M}, {Tag2, _, N}, _TEnv)
 refine_ty(Ty1, Ty2, _TEnv) when ?is_int_type(Ty1),
                                 ?is_int_type(Ty2) ->
     gradualizer_int:int_type_diff(Ty1, Ty2);
+refine_ty({user_type, Anno, Name, Args}, {user_type, Anno, Name, Args}, _TEnv) ->
+    % After being normalized, it's because it's defined as opaque.
+    % If it has the same annotation, name and args, it's the same.
+    type(none);
```
zuiderkwast: There is a clause like this above:

```Erlang
refine_ty(?type(T, A), ?type(T, A), _) ->
    type(none);
```

Perhaps both these could be replaced by simply one clause like this:

```Erlang
refine_ty(Same, Same, _TEnv) ->
    %% If it's the same, it's the same
    type(none);
```

What do you think? Future improvement?

zuiderkwast:
Very good! I think it's good that you merged, because it was already complete and working. More tests, refactoring, etc. can always be done later.

src/typechecker.erl
FrankBro: The only thing is this would also catch a few other cases I haven't put thoughts into. From looking at `gradualizer_type.erl`, The other stuff that would pass:

- `ann_type`: Are these supposed to be removed at that point? I think I saw a PR go by mentioning this but haven't looked into it enough.
- `var`: Honestly have no clue, I have barely touched type variables at this point.
- `remote_type`: Pretty sure normalize catches this and either expand it in a `type`, or a `user_type` if it's opaque.
- `user_type`: D'uh. After normalization, the only way it's still a `user_type` is an opaque type from another module. Seems fine and would remove the case I wrote.

I created my other PR #277 but not sure I wanna include them. I merged this because I'm using our existant codebase to test my record refinement branch and I needed this change 😄

zuiderkwast: OK!

Anything minus itself would be none() I guess, even a type variable. (Then, perhaps we also may want to add other cases such as `SomeType \ TypeVar` and vice versa and add constraints.)

PR #280, opened 2020-02-17
Keep vars in unfold bounded type

zuiderkwast:
The function `unfold_bounded_type/2` used to pre-process a spec does what the name says, except it also replaces unconstrained type variables with `any()`. This PR makes sure unconstrained type variables are kept.

This function transforms a spec like the one for lists:unzip/1 ...

```Erlang
-spec unzip(List1) -> {List2, List3} when
      List1 :: [{A, B}],
      List2 :: [A],
      List3 :: [B],
      A :: term(),
      B :: term().
```

... in the following way:

| Before this PR | After this PR |
|---------------------|-------------------|
| `fun(([{any(), any()}]) -> {[any()], [any()]})` | `fun(([{A, B}]) -> {[A], [B]})` |

This, together with the fact that now `any() = term()`, permits deleting most of the spec overrides in the prelude without any broken test cases.


zuiderkwast:
> This fits very nicely with the work I'm doing on the constraint solver at the moment!

Yes, I know. :-)

FrankBro:
Very nice!

PR #283, opened 2020-02-26
Bit syntax integer signedness

FrankBro:
According to https://erlang.org/doc/programming_examples/bit_syntax.html#defaults

The default signedness for integer is `unsigned`. Feed on the specifiers to control signedness and make the default `unsigned`, which leads to unspecified signedness going from `integer()` to `non_neg_integer()`.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3958,17 +3958,17 @@ count_var_occurrences(Exprs) ->
                                          default,
                            Specifiers :: [atom() | {unit, pos_integer()}] |
                                          default},
-                           Source     :: pattern | expr) -> type().
-type_of_bin_element({bin_element, Anno, Expr, Size, default}, Source) ->
-    type_of_bin_element({bin_element, Anno, Expr, Size, []}, Source);
-type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}, Source) ->
+                           OccursAs   :: pattern | expr) -> type().
+type_of_bin_element({bin_element, Anno, Expr, Size, default}, OccursAs) ->
+    type_of_bin_element({bin_element, Anno, Expr, Size, []}, OccursAs);
+type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}, OccursAs) ->
     %% String literal is syntactic sugar for multiple char literals,
     IsStringLiteral = case Expr of
                           {string, _, _} -> true;
                           _              -> false
                       end,
     IsSigned =
-        case Source of
+        case OccursAs of
             pattern -> lists:member(signed, Specifiers);
             expr -> true
         end,
@@ -3982,11 +3982,10 @@ type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}, Source) ->
                                     IsStringLiteral ->
                                         %% <<"ab"/utf8>> == <<$a/utf8, $b/utf8>>.
                                         {true, type(string)};
-                                    not IsStringLiteral ->
-                                        case IsSigned of
-                                            true -> {true, type(integer)};
-                                            false -> {true, type(non_neg_integer)}
-                                        end
+                                    IsSigned ->
+                                        {true, type(integer)};
+                                    true ->
+                                        {true, type(non_neg_integer)}
                                 end;
                             (float) when IsStringLiteral ->
                                 %% <<"abc"/float>> is integers to floats conversion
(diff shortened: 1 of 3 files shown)


FrankBro:
Actually I think there's still a problem.

```erlang
-spec receiver(binary()) -> non_neg_integer().
receiver(B) ->
    <<I>> = B,
    I.

-spec emitter(integer()) -> binary().
emitter(I) ->
    B = <<I>>,
    B.
```

I think for `receiver`, it makes sense to require `non_neg_integer()` but with these changes, `emitter` errors out because we're building the binary with an `integer()` instead of a `non_neg_integer()`.

Considering the following:

```erlang
3> <<A/unsigned>> = <<-1>>.
<<"ÿ">>
4> A.
255
5> <<B/signed>> = <<-1>>.
<<"ÿ">>
6> B.
-1
7>
```

I think it should allow any sort of integer if it's on RHS, right?

src/typechecker.erl
```diff
@@ -3966,6 +3966,7 @@ type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}) ->
                           {string, _, _} -> true;
                           _              -> false
                       end,
+    IsSigned = length([signed || signed <- Specifiers]) > 0,
```
Zalastax: Use lists:element instead!

Zalastax:
From linked article:

> The signedness specification can be either signed or unsigned. Notice that signedness only matters for matching.

src/typechecker.erl
FrankBro: `lists:element` does not exist? 🤔

Zalastax: Sorry, I meant lists: member

FrankBro:
> From linked article:
> 
> > The signedness specification can be either signed or unsigned. Notice that signedness only matters for matching.

Indeed. Fixed with the new commit.

Zalastax:
Great! Add your examples as test too :smiley:

src/typechecker.erl

FrankBro:
Should be good now. Requests approvals and I think this is ready to merge.

src/typechecker.erl
```diff
@@ -4005,7 +4014,10 @@ type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}) ->
             type(string);
         [] ->
             %% <<X>>
-            type(integer);
+            case IsSigned of
+                true -> type(integer);
+                false -> type(non_neg_integer)
```
zuiderkwast: Combine nested case for less nesting:

```Erlang
    [] when IsSigned -> ...
    [] when not IsSigned -> ...
```

src/typechecker.erl
```diff
@@ -3957,15 +3957,21 @@ count_var_occurrences(Exprs) ->
                            Size       :: non_neg_integer() |
                                          default,
                            Specifiers :: [atom() | {unit, pos_integer()}] |
-                                         default}) -> type().
-type_of_bin_element({bin_element, Anno, Expr, Size, default}) ->
-    type_of_bin_element({bin_element, Anno, Expr, Size, []});
-type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}) ->
+                                         default},
+                           Source     :: pattern | expr) -> type().
```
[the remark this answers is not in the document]

src/typechecker.erl
```diff
@@ -3977,7 +3983,10 @@ type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}) ->
                                         %% <<"ab"/utf8>> == <<$a/utf8, $b/utf8>>.
                                         {true, type(string)};
                                     not IsStringLiteral ->
-                                        {true, type(integer)}
+                                        case IsSigned of
+                                            true -> {true, type(integer)};
+                                            false -> {true, type(non_neg_integer)}
+                                        end
```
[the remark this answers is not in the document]

src/typechecker.erl

FrankBro:
I agreed with all of them :)

src/typechecker.erl
```diff
@@ -4012,12 +4011,12 @@ type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}, Source) ->
         [] when IsStringLiteral ->
             %% <<"abc">>
             type(string);
-        [] ->
+        [] when IsSigned ->
             %% <<X>>
```
zuiderkwast: The comment should probably be `%% <<X/signed>>` in this case

zuiderkwast:
Awesome! Then I have just one more but just it's a comment about a comment...

src/typechecker.erl
FrankBro: It's also signed if it's an expression but I'll put two comments basically

FrankBro:
Should I merge?

zuiderkwast:
Or better you merge yourself, you can squash as you want. :+1:

PR #284, opened 2020-03-01
Constraint solver

josefs:
This is a technology preview of the constraint solver I've been working on. This is a very simple solver which doesn't handle union types properly, so there are programs which will still pass the typechecker despite not being welltyped. That being said, it can catch a fair amount of typeerrors where type variables are involved.

[Edit: removed comment about bug that I've now fixed.]


zuiderkwast:
Good progress!

nice!

But no tests?

I am curious to which polymorhic definitions and usages of polymorphic functions this solves. Also, is there any example of what should fail but doesn't because of the non-transitive compatible-subtyping issue?

I'll look more another day, but it's a bit hard to grasp without tests. Also, it would be nice if you can avoid adding tabs and avoid converting existing indentation from spaces to tabs.

erszcz:
...and fixed.

erszcz:
Superseded by #450.

Issue #285, opened 2020-03-01
Map showing multiple keys for documentation purpose.


FrankBro:
I came along this example while playing with cowboy: https://github.com/ninenines/cowboy/blob/master/src/cowboy_http2.erl#L25-L68

I also recreated a simpler version:

```erlang
-type one() :: #{
    one => atom(),
    two => atom()
}.

-type opts() :: one().

-spec test1(opts()) -> ok.
test1(Test) -> ok.

-spec test2() -> ok.
test2() ->
    test1(#{two => two}).
```

We get this:

```
The map on line 15 at column 11 is expected to have type opts() but it has type #{two := two}

-spec test2() -> ok.
test2() ->
    test1(#{two => two}).
          ^^^^^^^^^^^^^
```

I've boiled it down to this line in typechecker: https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L282

```erlang
%% For M1 <: M2, mandatory fields in M2 must be mandatory fields in M1
```

I'm down to take care of this with a PR but before doing in, I'm wondering if this is what is expected. I've omitted the fact that in the cowboy version, the map signature has a `_ => _` signature as well. I'm assuming all this does is allow any key to be in as well.

I would assume that writing a bunch of keys like so just means: here is all the possible key-value pairs you could put in the map. Should `compat_ty` here just try to find one matching pair?

Zalastax:
That is not expected:

> A map association has a key in AssociationList if it belongs to this type. AssociationList can contain both mandatory (:=) and optional (=>) association types. If an association type is mandatory, an association with that type needs to be present. In the case of an optional association type it is not required for the key type to be present.

FrankBro:
So it's a bug? It's treating it as mandatory? I get this when I thread `compat_ty`

```
DEBUG: Ty1 = {type,0,map,
                   [{type,0,map_field_exact,[{atom,0,two},{atom,0,two}]}]}
DEBUG: Ty2 = {type,0,map,
                   [{type,0,map_field_assoc,[{atom,0,one},{type,0,atom,[]}]},
                    {type,0,map_field_assoc,[{atom,0,two},{type,0,atom,[]}]}]}
DEBUG: Ty1 = {type,0,map_field_assoc,[{atom,0,one},{type,0,atom,[]}]}
DEBUG: Ty2 = {type,0,map_field_exact,[{atom,0,two},{atom,0,two}]}

```

zuiderkwast:
[...]

PR #286, opened 2020-03-02
Reverse map assocs check

FrankBro:
Check #285 

This is a WIP for now. It breaks a bunch of stuff.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -271,17 +271,17 @@ compat_ty({type, _, map, any}, {type, _, map, _Assocs}, A, _TEnv) ->
 compat_ty({type, _, map, _Assocs}, {type, _, map, any}, A, _TEnv) ->
     ret(A);
 compat_ty({type, _, map, Assocs1}, {type, _, map, Assocs2}, A, TEnv) ->
-    lists:foldl(fun (Assoc2, {As, Cs1}) ->
+    lists:foldl(fun (Assoc1, {As, Cs1}) ->
 			begin
-			    {Ax, Cs2} = any_type(Assoc2, Assocs1, As, TEnv),
+			    {Ax, Cs2} = any_type(Assoc1, Assocs2, As, TEnv),
 			    {Ax, constraints:combine(Cs1, Cs2)}
 			end
-		end, ret(A), Assocs2);
+		end, ret(A), Assocs1);
 compat_ty({type, _, AssocTag2, [Key2, Val2]},
           {type, _, AssocTag1, [Key1, Val1]}, A, TEnv)
         when AssocTag2 == map_field_assoc, AssocTag1 == map_field_assoc;
              AssocTag2 == map_field_exact, AssocTag1 == map_field_exact;
-             AssocTag2 == map_field_assoc, AssocTag1 == map_field_exact ->
+             AssocTag2 == map_field_exact, AssocTag1 == map_field_assoc ->
     %% For M1 <: M2, mandatory fields in M2 must be mandatory fields in M1
     {A1, Cs1} = compat(Key1, Key2, A, TEnv),
     {A2, Cs2} = compat(Val1, Val2, A1, TEnv),
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -271,14 +271,18 @@ compat_ty({type, _, map, any}, {type, _, map, _Assocs}, A, _TEnv) ->
 compat_ty({type, _, map, _Assocs}, {type, _, map, any}, A, _TEnv) ->
     ret(A);
 compat_ty({type, _, map, Assocs1}, {type, _, map, Assocs2}, A, TEnv) ->
-    lists:foldl(fun (Assoc1, {As, Cs1}) ->
+    lists:foldl(fun (Assoc2, {As, Cs1}) ->
 			begin
-			    {Ax, Cs2} = any_type(Assoc1, Assocs2, As, TEnv),
+                {Ax, Cs2} =
+                    case Assoc2 of
+                        {type, _, map_field_exact, _} -> any_type(Assoc2, Assocs1, As, TEnv);
+                        {type, _, map_field_assoc, _} -> safe_any_type(Assoc2, Assocs1, As, TEnv)
+                    end,
 			    {Ax, constraints:combine(Cs1, Cs2)}
 			end
-		end, ret(A), Assocs1);
-compat_ty({type, _, AssocTag1, [Key1, Val1]},
-          {type, _, AssocTag2, [Key2, Val2]}, A, TEnv)
+		end, ret(A), Assocs2);
+compat_ty({type, _, AssocTag2, [Key2, Val2]},
+          {type, _, AssocTag1, [Key1, Val1]}, A, TEnv)
         when AssocTag2 == map_field_assoc, AssocTag1 == map_field_assoc;
              AssocTag2 == map_field_exact, AssocTag1 == map_field_exact;
              AssocTag2 == map_field_assoc, AssocTag1 == map_field_exact ->
@@ -330,6 +334,17 @@ compat_record_fields(_, _, _, _) ->
 %% there were no type variables involved.
 ret(A) ->
     {A, constraints:empty()}.
+
+safe_any_type(_Ty, [], A, _TEnv) ->
+    ret(A);
+safe_any_type(Ty, [Ty1|Tys], A, TEnv) ->
+    try
+        compat(Ty, Ty1, A, TEnv)
+    catch
+        nomatch ->
+            safe_any_type(Ty, Tys, A, TEnv)
+    end.
+
 
 any_type(_Ty, [], _A, _TEnv) ->
     throw(nomatch);
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -553,11 +553,26 @@ glb_ty({type, _, record, _}, {type, _, record, _}, _A, _TEnv) ->
 
 %% Map types. These are a bit tricky and we can't reach this case in the
 %% current code. For now going with a very crude approximation.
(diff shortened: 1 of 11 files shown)


src/typechecker.erl
```diff
@@ -268,17 +268,17 @@ compat_ty({type, _, map, any}, {type, _, map, _Assocs}, A, _TEnv) ->
...
+    lists:foldl(fun (Assoc1, {As, Cs1}) ->
 			begin
-			    {Ax, Cs2} = any_type(Assoc2, Assocs1, As, TEnv),
+			    {Ax, Cs2} = any_type(Assoc1, Assocs2, As, TEnv),
 			    {Ax, constraints:combine(Cs1, Cs2)}
 			end
-		end, ret(A), Assocs2);
+		end, ret(A), Assocs1);
 compat_ty({type, _, AssocTag2, [Key2, Val2]},
           {type, _, AssocTag1, [Key1, Val1]}, A, TEnv)
```
zuiderkwast: Should these be swapped, instead/also?

FrankBro: Good catch, I believe it should. Failures went from 12 to 5 just from that change.

FrankBro:
It seems like some of them make sense. Considering usage is compared to expected.

```erlang
subtype ( ? t ( # { a => b } ) , ? t ( # { } ) )
```

Ok probably a mistake. It thinks it expects an empty map in this situation. Also, if the left side is what is produced, even if you write 

```erlang
#{ a => b}
```

It will be inferred as 

```erlang
#{ a := b }
```

Here's another one

```erlang
subtype ( ? t ( # { a => b , c => d } ) , ? t ( # { a => b } ) )
```

Well yea, you're adding a key that isn't expected? Is the test wrong? The compat stuff is called via subtype. The comment on top says:

```erlang
%% The first argument is a "compatible subtype" of the second.
```

Not sure `#{ a => b, c => d}` is a compatible subtype of `#{ a => b }`, is it?

josefs:
@FrankBro both of the examples you bring up are correct. A `msub` is a subtype of another map `msup` if it contains at least all the keys of `msup`. That certainly holds in all of the examples. In the first example, there are no keyes in the supertype so any record is a subtype of that. For the second example, the subtype contains all the keys and more.

A useful analog here might be to think about inheritance in an object oriented language. For instance in Java, you can defined a subclass (which corresponds to a subtype for us) by adding methods and members to an existing class. In a similar vein, we can add keys to a map to form a new subtype of that map.

FrankBro:
From that description alone, it would mean `#{b := b}` is indeed not a subtype of `#{a => a, b => b}`, which is the initial example I gave. Unless we start to introduce the `assoc`/`exact` difference. I guess I'm wondering what's the exact law of map subtyping in this situation. Maybe then it makes sense to go over the second map's keys instead of the first one. We should just skip if it's `=>` and not present in the first one then?

FrankBro:
The last change has only 1 failure then. One of them is a known problem being solved. Looking into it.

FrankBro:
Alrighty then, name of the branch is bad but everything makes sense. Fixed a known problem, removed a failing test that was supposed to pass and added a few tests. The solution was simpler than anticipated 😅

FrankBro:
I'll merge once approved

FrankBro:
Found a problem. `no_match` catches too many things. If the same key is used wrong, should be an error. Fixing...

```erlang
-spec optional_fail() -> #{a => a, b => b}.
optional_fail() -> #{a => b, b => b}.
```

zuiderkwast:
@josefs 
> A `msub` is a subtype of another map `msup` if it contains at least all the keys of `msup`.

That's not the full story. `#{}` is the empty map type, where no keys are allowed. If you add keys, it's not a subtype of the empty map. The supertype of all maps is `#{top() => top()}`.

All keys in the map subtype have to be a subtype of some key in the map supertype and the value must be a subtype of the corresponding value type. A map subtype can also turn optional keys from the supertype into mandatory keys.

@FrankBro 

> Unless we start to introduce the `assoc`/`exact` difference.

We already have some handling of optional/mandatory (assoc/exact) associations, although it seems it's not entirely correct.

I'd like to see unit tests in `typechecker_tests` for map subtyping, to isolate this subtyping logic from other typechecking logic.

src/typechecker.erl
```diff
@@ -317,6 +321,17 @@ compat_record_fields(_, _, _, _) ->
 ret(A) ->
     {A, constraints:empty()}.
 
+safe_any_type(_Ty, [], A, _TEnv) ->
```
zuiderkwast: Please document this. I can't see what it does from just the name. What's safe?

FrankBro:
@zuiderkwast so then this is wrong?

```erlang
subtype ( ? t ( # { a => b } ) , ? t ( # { } ) )
```

src/typechecker.erl
FrankBro: I figured out it's wrong anyway now. The point was to allow skipping keys that are not present, the side-effect right now is that it also allows a key to be there with the wrong type. I will make one for `assoc` and one for `exact` most likely and it'll make more sense.

zuiderkwast:
This is how I understand map types and a draft of an informal rule for map subtyping: https://github.com/josefs/Gradualizer/wiki/Map-types

josefs:
@zuiderkwast Thanks for writing this document! That's super helpful and a good basis for further discussion.

That being said, there are a few things that I disagree with. You say "Type #{}: The empty map type has only one value: The empty map #{}. No keys are allowed." But the syntax of Erlang disagrees. Consider the following program:
```
f(#{}) ->
    ok.
```
I'd argue that it should have the following spec: `f(#{}) -> ok.` That would mean that we can only call it with the empty map according to you semantics. But the pattern matching semantics in Erlang has subtyping built in. The call `f(#{a => b}).` doesn't crash, it returns `ok`. 

The same story goes for pattern matching against associations. Here's an example:
```
h(#{ a := B }) ->
    ok.
```
This function should have the type `h(#{ a := B })`.  But should still be fine to call this function with a map that contains more associations, like `h(#{a => b, c => d})` which returns `ok` if you try it in the repl.

If we're going to have correct type refinement for maps is it vital that we allow the kind of subtyping I outline above.

Zalastax:
I'd argue that the function actually has the spec `f(#{any() => any()) -> ok`. The documentation is quite specific in disallowing more associations. Unfortunately it doesn't talk about subtyping but I believe @zuiderkwast's documentation is correct.

josefs:
> I'd argue that the function actually has the spec f(#{any() => any()) -> ok.

I assume you're referring to the first function. I think it'd be very confusing and unfortunate to give the function that type. We have a piece of syntax, `#{}`, which occurs both on the type level, as a value, and as a pattern. We should make sure to keep their meaning together. It seems obvious that when patternmatching on an empty map `#{}`, that such a pattern would give rise to expecting the scrutinee to have the type `#{}`. Anything else would be confusing.

> The documentation is quite specific in disallowing more associations.

What documentation are you referring to? Are you referring to this sentence from the "Types and Function Specifications" section in the reference manual:

> The notation #{} specifies the singleton type for the empty map. Note that this notation is not a shorthand for the map() type.

Just because the type denotes a singleton value doesn't mean that it cannot have subtypes.

> Unfortunately it doesn't talk about subtyping but I believe @zuiderkwast's documentation is correct.

It's not correct for type refinement of maps. The pattern for the empty map matches all maps, not only the empty map.

zuiderkwast:
> We have a piece of syntax, `#{}`, which occurs both on the type level, as a value, and as a pattern. We should make sure to keep their meaning together. It seems obvious that when patternmatching on an empty map `#{}`, that such a pattern would give rise to expecting the scrutinee to have the type `#{}`. Anything else would be confusing.

The overloaded syntax with different meanings in patterns, expressions and types is confusing indeed. Unfortunately we cannot change the syntax of Erlang patterns and expressions. Regarding types, I think we should follow the reference manual.

A pattern can only have `:=` associations and extra keys don't prevent the pattern from matching. Map patterns are documented in the reference manual for expressions. That means, agreeing with @Zalastax, that the pattern `#{}` has the type `#{any() => any()}` (or maybe `#{top => top()}`). Perhaps we need to clarify somewhere what it means that a pattern has a type, since it's usually values that have types....

> > The notation #{} specifies the singleton type for the empty map. Note that this notation is not a shorthand for the map() type.
> 
> Just because the type denotes a singleton value doesn't mean that it cannot have subtypes.

Well..... it has `any()` and `none()` as a subtypes. It probably also has the unknown map (`map()`  or `#{any() => any()}` as a subtype (since `#{}` is equivallent to `#{none() => none()}` and we have `any() :: none()`.) Apart from that, it's similar to other singletons, such as `[]` and `{}` IMO.

> It's not correct for type refinement of maps. The pattern for the empty map matches all maps, not only the empty map.

That's a very good point❗️ When we do refinement based on patterns, we should be careful. For example, we shouldn't infer the type `#{}` from the pattern `#{}`.

I volunteer to add some section about map patterns on in the same wiki page. Probably we should cover it on the refinement page as well.

berbiche:
> This is how I understand map types and a draft of an informal rule for map subtyping: https://github.com/josefs/Gradualizer/wiki/Map-types

If I understand the conversation and the documentation above correctly, Gradualizer does not currently support subtyping `#{} :: #{_ -> _}`?

zuiderkwast:
> If I understand the conversation and the documentation above correctly, Gradualizer does not currently support subtyping `#{} :: #{_ -> _}`?

I don't remember anymore what's implemented and what's not. We'd need a test or a known problem to figure out. Or just try!

Most of the discussion above is about the confusing difference in the pattern `#{}`, the value or expression `#{}` and the type `#{}` which look the same but are very different. The pattern `#{}` doesn't have the type `#{}`.

erszcz:
This PR has been inactive for a long time and significant work has happened on maps in the meantime. I'm closing it. Please let me know if it needs revisiting.

PR #293, opened 2020-11-02
Warning and shadowing

FrankBro:
Fix two unused variables warnings as well as two frankly embarrassing shadowed variables which would remove record refinement.


zuiderkwast (approved):
Looks good to me. :+1:

PR #302, opened 2020-12-10
rebar3 gradualizer --use_beams

### LINKED ISSUE #300: Let rebar3 plugin check beam instead of erl files
I tried the rebar3 plugin on a project today. I ran into a few problems.

1. I get a lot of error like "Undefined type options/0 on line 0", which I assume is because the "include/" directory is not included when parsing the erl files (would be solved by checking beams instead, but it would also be possible to do better guessing of include dirs). (Maybe I could solve this by setting `gradualizer_opts` in rebar.config, but working out of the box would be nicer.)
2. The output doesn't contain the names of the files being checked, making it hard to see what each error is referring to.
3. I would like to make Gradualizer a dependency in my lib only when running checks, not when only compiling and when using my lib as a dependency in another project. Is this possible already?

ilya-klyuchnikov:
`rebar3 gradualizer --use_beams` uses `*.beam` files from `ebin` directories as input (beam files should be produced with debug info to be analyzable).

(closes #300)

Tested locally with the latest rebar. 

To make local testing easier use these options:
```erlang
{gradualizer_opts, [print_file]}.
```

In this case - you can check `XXX.erl` vs `XXX.beam` when gradualizer reports an error.

---

Discussion: making using beam files a default choice is out of scope of this PR, - it may be addressed separately.

### DIFF
--- a/examples/rebar3/README.md
+++ b/examples/rebar3/README.md
@@ -7,7 +7,20 @@ To run Gradualizer from rebar3, add it as a plugin in your `rebar.config`:
 ]}.
 ```
 
-# Options
+# Analyzing erl files or beam files
+
+Gradualizer can process both `*.erl` files and `*.beam` files.
+
+By default, `rebar3 gradualizer` processes `*.erl` files from source directories.
+In this default mode gradualizer tries its best to get include directories, header files
+and compiler options right. However, in case of complex setup gradualizer may miss something.
+
+`rebar3 gradualizer --use_beams` uses `*.beam` files from `ebin` directories as input
+(beam files should be produced with debug info to be analyzable).
+
+`--use_beams` may be a more robust choice if there are problems with analyzing `*.erl` files.
+
+# Options (`gradualizer_opts`)
 
 Gradualizer checks every source file in your app(s), unless configured not to.
 Configuration is read from `gradualizer_opts` in `rebar.config` which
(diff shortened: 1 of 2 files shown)


josefs:
This makes a lot of sense to me but I'm not the expert here. @Zalastax or @zuiderkwast what do you think?

Zalastax:
Looks great! I would go for beams by default. If we do that, what should the parameter be called?

ilya-klyuchnikov:
> I would go for beams by default. If we do that, what should the parameter be called?

Probably `--use_erls`? - To denote the fact that only `erl` files would be used. - In theory gradualizer should be applicable to elixir beams as well (since elixir compiles through abstract forms under the hood), - but gradualizer is not capable for now to read Elixir sources, - so `--use_erls` would create less confusion here.

ilya-klyuchnikov:
So, @Zalastax - let's land this and attack beams by default separately? (Maybe other people would suggest better naming for `--use_erls`).

zuiderkwast:
Nice! I think we can delete the erl way in the rebar plugin and always let it gradualize from beams. If rebar3 is used in a project, it's also used for compiling it, so the beams should be there already.

Alternatively, if erls are used, we should be able to figure out the include dirs used for compiling, since rebar should have them somewhere. (Not in the scope of this PR.)

Zalastax:
Feel free to keep the erl way or to just support via beams (preferred by me).

ilya-klyuchnikov:
It seems then that @Zalastax or @zuiderkwast may want to merge this PR or approve it (since I cannot merge it on my own).

Issue #291, opened 2020-10-21
Missing included path and function signature being present or not.


FrankBro:
```erlang
% include/test.hrl
-define(ZERO, 0).
```

```erlang
-module(test).

-include("test.hrl").

-spec test1(integer()) -> bool().
test1(I) -> test2(I).

-spec test2(integer()) -> bool().
test2(I) ->
    I < ?ZERO.
```

Depending on how you invoke gradualizer for this, you will get a success or not. The cases are like so:

| `-I include` | `-spec test2...` present | Result |
| - | - | - |
| True |  True | No error |
| True | False | No error |
| False |  True | No error |
| False | False | `Call to undefined function test2/1 on line 6 at column 13` |

Funny enough, if you check the output of `epp`, you see an error within the body of `test2` for the error case: makes sense, the macro `?ZERO` is undefined.

Does the error make sense? The function _IS_ defined in there but the body fails. There's no function signature anyway ... shouldn't we just expect `any()` anyway?

What's the best way to fix this?

zuiderkwast:
## How Gradualizer should handle the error

The reported error is very confusing, indeed. I think we should not typecheck at all if the include files can't be found. Epp returns an error tuple where the `-include` directive was if the file wasn't found. If the list of forms contains any error, we should abort IMO.

There is a simple way to catch this error and other errors such as function clauses in the wrong order and various other things which currently confuses Gradualizer. That is to call **`compile:forms(Forms, [basic_validation, return_errors, return_warnings])`**. It has been suggested before, but Josef thought that it might have a performance impact. I don't think so though. It's a simple pass and it solves a lot of problems.

## How to gradualize from source code when you don't have the right `-I` flags to use

You need the same `-I` flags as for compiling. Since it's very hard to know which `-I` flags are necessary to complile an erl file in the general case (e.g. a project might have complicated structures of make files and building up variables to compose the list of `-I` flags which are passed to `erlc`) the best way to run Gradualizer in this case is to make sure the modules are compiled with `debug_info` and then gradualize the beam files.

josefs:
First of all, sorry for going awol on this project and not responding to notifications. The whole covid thing has been rough for me. 

> That is to call compile:forms(Forms, [basic_validation, return_errors, return_warnings]). It has been suggested before, but Josef thought that it might have a performance impact. I don't think so though. It's a simple pass and it solves a lot of problems.

[...]

Issue #288, opened 2020-06-05
Should not pass on malformed expression.


FrankBro:
Simple example, returns no error, even if there is a syntax error.

```erlang
-record(r, {
    integer :: integer()
}).

-spec incr(integer()) -> integer().
incr(I) -> I + 1.

-spec test() -> integer().
test() ->
    Record = #record{integer = 0},
    incr(Record.integer). % Should be Record#r.integer
```

FrankBro:
Is it important if trying to compile it with Erlang itself throws an error?

```
syntax error before: '.'
```

Zalastax:
There's this idea that Gradualizer should in theory be possible to embed in the compiler so it should handle syntax errors and not rely on the compiler to detect syntax errors. I think an error should be produced here. Adding a special case for this abstract format should do the trick!

gomoripeti:
What happens in this case is that the AST contains some error entries
```
...
     {attribute,10,spec,
                {{test,0},
                 [{type,10,'fun',
                        [{type,10,product,[]},{type,10,integer,[]}]}]}},
     {error,{13,erl_parse,["syntax error before: ","'.'"]}},
     {eof,14}]}
```
Maybe gradualizer_file_utils could scan the AST forms to find such error entries and report it.
But there is also value in being able to check incomplete code, ie gradualizer can report type errors in other functions, and you already have a great tool to find syntax errors, the compiler.

zuiderkwast:
I think we should pass the AST to `compile:forms/2` with option `basic_validation`, which is "fast way to test whether a module will compile successfully".

Then we can also remove all the checks that the complier does, such as undefined functions, types, records, etc. (Note though that the compiler doesn't check for remote undefined functions and types.)

PR #303, opened 2020-12-10
Compile test data before gradualizing 

ilya-klyuchnikov:
1. Check that test files (in `test/known_problems`, `test/should_fail` and `test/should_pass`) are compilable. (Checking this revealed that there were some nits/flaws/typos in test files - and sometimes failure was not attributed correctly, - eg `intersection_using_constraits` - it had a parse error, which was considered as gradualizer error).
2. Fixing broken test files. - The most interesting cases are `intersection_using_constraints` - which actually worked, but were reported incorrectly because of parse errors.

- See CI status: https://github.com/ilya-klyuchnikov/Gradualizer/commits/upstream-test

Screenshot 2020-12-10 at 23 41 30

Not trying to compile `mix` - since it has `undefined_errors.erl` - many of them are caught by the compiler in the first place (undefined var, undefined local funs, etc).

-------------

Next possible step - to introduce "snapshot" testing of errors - https://github.com/ilya-klyuchnikov/Gradualizer/commit/8b4784f3f8221b06393fdd6ef598c97291c8f0ba - which makes reviewing of PRs easier and more comprehensible - having both input and precise expected output (with locations/highlighting) for a test case in the repo.

### DIFF
--- a/test/known_problems/should_fail/intersection_with_any_fail.erl
+++ b/test/known_problems/should_fail/intersection_with_any_fail.erl
@@ -1,20 +1,12 @@
 -module(intersection_with_any_fail).
 
 -export([any_refined_using_guard/1,
-         intersection_using_constraits/1,
          var_as_pattern/1,
          var_inside_pattern/1]).
 
 %% X :: any() & atom() by refinement
 -spec any_refined_using_guard(any()) -> 5.
 any_refined_using_guard(X) when is_atom(X) ->
-    X.
-
-%% X :: integer() & any() by the constraints
--spec intersection_using_constraits(X) when X :: integer(),
-                                            X :: any() ->
-                                                      atom().
-intersection_using_constraits(X) ->
     X.
 
 -spec var_as_pattern(atom()) -> integer().
--- a/test/should_pass/annotated_types.erl
+++ b/test/should_pass/annotated_types.erl
@@ -2,7 +2,7 @@
 
 -export([f/1, g/0, h/1, i/1]).
 
--include_lib("gradualizer/include/gradualizer.hrl").
+-include_lib("../../include/gradualizer.hrl").
 
 f(Expr) ->
     {call, _, _Name, Args} = Expr,
--- a/test/should_pass/intersection_with_any.erl
+++ b/test/should_pass/intersection_with_any.erl
@@ -7,9 +7,9 @@
 -spec any_refined_using_guard(any()) -> 5.
 any_refined_using_guard(X) when is_integer(X) -> X.
 
--spec intersection_using_constraints(X) when X :: integer(),
-                                             X :: any() ->
-                                                       5.
+-spec intersection_using_constraints(X) -> X when
+                                                X :: integer(),
+                                                X :: any().
 intersection_using_constraints(X) -> X.
 
 -spec guess_two_dice(integer(), integer()) -> 0..6.
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ test/covertool.hrl
 /gradualizer
 /_build
 /bin/
+/test_data/
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ clean:
 	rm -rf bin/gradualizer ebin cover test/*.beam
 
 .PHONY: tests eunit compile-tests cli-tests
-tests: eunit cli-tests
+tests: build_test_data eunit cli-tests
 
 test_erls=$(wildcard test/*.erl)
 test_beams=$(test_erls:test/%.erl=test/%.beam)
@@ -124,6 +124,12 @@ test/any.beam: test/should_pass/any.erl
 
 test/records.beam: test/should_pass/records.erl
 	erlc $(ERLC_OPTS) -o test $<
+
+.PHONY: build_test_data
+test_data_erls = $(wildcard test/known_problems/**/*.erl test/should_fail/*.erl test/should_pass/*.erl)
+build_test_data:
+	mkdir -p "test_data"
+	erlc -o test_data $(test_data_erls)
 
 EUNIT_OPTS =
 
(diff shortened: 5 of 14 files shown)


josefs (approved):
image

test/should_pass/annotated_types.erl
```diff
@@ -2,7 +2,7 @@
 
 -export([f/1, g/0, h/1, i/1]).
 
--include_lib("gradualizer/include/gradualizer.hrl").
+-include_lib("../../include/gradualizer.hrl").
```
zuiderkwast, line 5: Perhaps use `-include` instead of `-include_lib` when using a relative path?

test/should_pass/intersection_with_any.erl
```diff
@@ -7,9 +7,9 @@
 -spec any_refined_using_guard(any()) -> 5.
 any_refined_using_guard(X) when is_integer(X) -> X.
 
--spec intersection_using_constraints(X) when X :: integer(),
-                                             X :: any() ->
-                                                       5.
+-spec intersection_using_constraints(X) -> X when
+                                                X :: integer(),
+                                                X :: any().
```
zuiderkwast, line 12: Wow, amazing :)

zuiderkwast (commented):
Nice!

When gradualizing erl files, we have talked about running the basic validation pass of compiler (e.g. in #288, #291). That would also catch these errors.

test/should_pass/intersection_with_any.erl
zuiderkwast: But the return value should be `5`, i.e. a subset of `integer()`. It should fail if `X :: any()` is removed, since X :: integer() would mean any integer but `integer() & any()` would mean "some unknown subset of integer". At least I think that was the intention of this originally, but it's hard to tell since it was a syntax error.

ilya-klyuchnikov: Yep, - there are two tests now. - this one (which is testing the happy path) and that one:

https://github.com/josefs/Gradualizer/blob/d86b6b54b321255255afe26b80153fa2ea5dd81d/test/should_fail/intersection_with_any.erl#L5-L8

I believe that both intentions are there.

ilya-klyuchnikov: Note `should_pass` vs `should_fail` locations

test/known_problems/should_fail/intersection_with_any_fail.erl
```diff
@@ -10,13 +9,6 @@
 any_refined_using_guard(X) when is_atom(X) ->
     X.
 
-%% X :: integer() & any() by the constraints
--spec intersection_using_constraits(X) when X :: integer(),
-                                            X :: any() ->
-                                                      atom().
-intersection_using_constraits(X) ->
-    X.
```
zuiderkwast, line 18: If the syntax is corrected, i.e. `(X) -> atom() when X :: integer(), X :: any()`, (meaning X is some unknown subtype of integer) this should fail. Is this test case moved to another file? I don't see it.

test/should_pass/intersection_with_any.erl
zuiderkwast: Ah, thanks. Then, I think `(X) -> 5 when X :: integer(), X :: any()` should be under "known_problems/should_pass".

`X :: any()` makes it a dynamic ( = unknown at compile time) type which combined with `X :: integer()` makes it an unknown subtype of `integer()`, which may well be the singleton type `5`, so by gradual typing, we shouldn't give an error here.

We had some discussion about intersection types in #130. See also the paper *G. Castagna and V. Lanvin: Gradual Typing with Union and Intersection Types* [icfp17.pdf](https://www.irif.fr/~gc/papers/icfp17.pdf).

test/known_problems/should_fail/intersection_with_any_fail.erl
ilya-klyuchnikov: In a sense https://github.com/josefs/Gradualizer/pull/303/files#diff-101a1f8f433ecac6155ed3a0e816518e2ddf27a08f899bbe8785782cf91eca14 covers a similar case, but one more (#305 ) would not hurt

test/should_pass/intersection_with_any.erl
ilya-klyuchnikov: Got it,- I mixed/confused `any()` with `term()` here. Feel free to move it.

(Changing `any()` -> `term()` - makes it fine to me, and now I understand/capture this subtlety).

zuiderkwast: Actually `any()` and `term()` are both the dynamic type in Gradualizer. Originally `term()` meant the union of all static types, but we changed it because of how `term()` is used in practice in OTP and elsewhere.

test/known_problems/should_fail/intersection_with_any_fail.erl
zuiderkwast: They are different though. `5` is a subtype of `integer()` so we shouldn't report an error for this one (and if we do, we can call it a known problem). `atom()` isn't a subtype of `integer()` though, so that one is a type error whether `X :: any()` is present or not.

I'm sorry, it wasn't my intention to nag about these subtleties.

test/should_pass/intersection_with_any.erl
ilya-klyuchnikov: > so by gradual typing, we shouldn't give an error here

I see: https://github.com/josefs/Gradualizer/pull/305/commits/bf290481fb788af9ab3a4d8932173470bd5bc577

ilya-klyuchnikov: Justice restored: #305

test/known_problems/should_fail/intersection_with_any_fail.erl
ilya-klyuchnikov: Justice restored: #305

PR #304, opened 2020-12-11
error messages: always printing filename

ilya-klyuchnikov:
Per discussion in https://github.com/josefs/Gradualizer/issues/300#issuecomment-743126933

This PR:
- error printing is moved from typechecker into gradualizer_fmt
- Filename is always printed (no matter what). Fewer options, fewer code paths
- Tests are modified accordingly: tests checking how different options (about filename) are propagated down the line are just removed from `gradualizer_cli.erl`
- `cli-tests` in Makefile are modified accordingly. - Also testing that `*.beam` vs `*.erl` files are reported correctly.

----

The default/only format of printing - `filename: verbose message (XXX at line L column C ...)` vs `filename:L short message (XXX ...)` - is out of scope of this PR and can be discussed/PRed separately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,30 +149,15 @@ cli-tests: bin/gradualizer
 	# 1. When checking a dir; printing filename is the default
 	bin/gradualizer test/dir \
 	2>&1|perl -0777 -ne 'm%^test/dir/test_in_dir.erl:% or die "CLI 1 ($$_)"'
-	# 2. --no-print-file with directory
-	bin/gradualizer --no_print_file test/dir \
-	2>&1|perl -0777 -ne '/^The/ or die "CLI 2 ($$_)"'
-	# 3. --print-module with directory
-	bin/gradualizer --print_module test/dir \
-	2>&1|perl -0777 -ne '/^test_in_dir:/ or die "CLI 3 ($$_)"'
-	# 4. --print-basename with directory
-	bin/gradualizer --print_basename test/dir \
-	2>&1|perl -0777 -ne '/^test_in_dir.erl:/ or die "CLI 4 ($$_)"'
-	# 5. Checking a single file; not printing filename is the default
-	bin/gradualizer test/dir/test_in_dir.erl \
-	2>&1|perl -0777 -ne '/^The/ or die "CLI 5 ($$_)"'
 	# 6. Brief formatting
-	bin/gradualizer --fmt_location brief --print_basename test/dir \
-	2>&1|perl -0777 -ne '/^test_in_dir.erl:6:12: The variable/ or die "CLI 6 ($$_)"'
+	bin/gradualizer --fmt_location brief test/dir \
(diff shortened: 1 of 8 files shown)


ilya-klyuchnikov:
It's strange, that there is no CI signal here, but - it's available at original branch: https://github.com/ilya-klyuchnikov/Gradualizer/commits/filename

Makefile
```diff
@@ -144,36 +147,24 @@ eunit: compile-tests
...
-	# 9. Possible to exclude prelude (-0777 from https://stackoverflow.com/a/30594643/497116)
+	# 2. When checking a beam file; beam file name is printed
+	bin/gradualizer test/arg.beam \
+	2>&1|perl -0777 -ne 'm%^test/arg.beam:% or die "CLI 1 ($$_)"'
+	# 4. Brief formatting
+	bin/gradualizer --fmt_location brief test/dir \
+	2>&1|perl -0777 -ne '/^test\/dir\/test_in_dir.erl:6:12: The variable/ or die "CLI 6 ($$_)"'
+	# 5. Verbose formatting, without filename
+	bin/gradualizer --fmt_location verbose --no_fancy test/dir \
+	2>&1|perl -ne '/^test\/dir\/test_in_dir.erl: The variable N on line 6 at column 12/ or die "CLI 7 ($$_)"'
```
zuiderkwast, line 163: The comment "without filename" isn't right.

Makefile
```diff
@@ -144,36 +147,24 @@ eunit: compile-tests
...
-	bin/gradualizer --fmt_location verbose --no_print_file --no_fancy test/dir \
-	2>&1|perl -ne '/^The variable N on line 6 at column 12/ or die "CLI 7 ($$_)"'
-	# 8. No location, no filename
-	bin/gradualizer --fmt_location none --no_print_file --no_fancy test/dir/test_in_dir.erl \
-	2>&1|perl -ne '/^The variable N is expected/ or die "CLI 8 ($$_)"'
-	# 9. Possible to exclude prelude (-0777 from https://stackoverflow.com/a/30594643/497116)
+	# 2. When checking a beam file; beam file name is printed
+	bin/gradualizer test/arg.beam \
+	2>&1|perl -0777 -ne 'm%^test/arg.beam:% or die "CLI 1 ($$_)"'
+	# 4. Brief formatting
```
zuiderkwast, line 158: There is no `#3` (nitpick)

zuiderkwast (approved):
Very nice!

I'll wait a few hours to let others have a say before merging.

Does everyone (@josefs, @gomoripeti, @Zalastax) agree on the future steps that have been mentioned? That is

* Always print "filename:N" instead of "on line N" (with column when available)
* Print source filename from AST instead of beam filename

gomoripeti:
thanks @ilya-klyuchnikov for bringing all these changes upstream

(I did not realize there were so many file options, amazing)

PR #305, opened 2020-12-11
should_fail/intersection_with_any - one more example

ilya-klyuchnikov:
as per discussion in https://github.com/josefs/Gradualizer/pull/303#discussion_r540945875


zuiderkwast (approved):
Exactly! Good comments also.

PR #306, opened 2020-12-12
Fix expecting tuple type against any() 

gomoripeti:
In real world came up as
```
test.erl: The tuple is expected to have type any() | pid() but it has type {a, 1}

f() ->
    supervisor:terminate_child(?MODULE, {a, 1}).
```

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -959,6 +959,8 @@ infer_literal_string(Str) ->
                                {char, erl_anno:new(0), lists:last(SortedChars)}])])
     end.
 
+expect_tuple_type({type, _, any, []}, _N) ->
+    any;
 expect_tuple_type({type, _, tuple, any}, _N) ->
     any;
 expect_tuple_type({type, _, tuple, Tys}, N) when length(Tys) == N ->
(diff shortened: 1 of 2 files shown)


ilya-klyuchnikov:
Good find!

I was wondering, - whether it was the case only for tuples - started looking at other cases, - first 5 were OK, but then:

```erlang
test(X) ->
    f(<<Y || Y <- X >>).

-spec f(any() | atom()) -> ok.
f(X) -> ok.
```

```
test/dev/expects_any.erl: The binary comprehension on line 9 at column 7 is expected to have type any() | atom() but it has type <<_:_*1>>

test(X) ->
    f(<<Y || Y <- X >>).
      ^^^^^^^^^^^^^^^^
```

So, at least binary comprehensions have the same problem. - Probably some other expressions also need the same fix 

(may be a different PR).

ilya-klyuchnikov (approved):
Looks good to me.

Thanks

zuiderkwast (approved):
Good!

zuiderkwast:
@ilya-klyuchnikov 

> So, at least binary comprehensions have the same problem. - Probably some other expressions also need the same fix

Good. It's possible to add it to known problems...

I don't know if there is an `expect_binary` function at all and whether we should have it.

gomoripeti:
> binary comprehensions

Looks like there is no `expect_binary` and there should be
https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L2790

Even the below fails (good find @ilya-klyuchnikov )
```
test(X) ->
    f(<<Y || Y <- X >>).

-spec f(binary() | atom()) -> ok.
f(X) -> ok.
```

As it is not related to tuples nor any(), and it's a bit more than a one liner fix, I would keep it out of this PR

PR #307, opened 2020-12-13
Issue: arith_op_arg_types/2 for '-' and non_neg_integer() is too restricitve

gomoripeti:
Historically this function was probably used to encode:
IF A :: pos_integer() AND B :: 1 => (A - B) :: non_neg_integer()

However the logic is now flipped:
IF (A - B) :: non_neg_integer() => A :: pos_integer() AND B :: 1

I only hit this particular [clause](https://github.com/josefs/Gradualizer/blob/d973b3ee/src/typechecker.erl#L2531), but there could be further issues given the flip


josefs (approved):
Good catch!

PR #308, opened 2020-12-17
typechecker: toggle exhaustiveness checking for records

berbiche:
This toggles exhaustiveness checking for records.

While testing, I discovered that type guards weren't handled in the refinement for record fields.
This will be fixed in an upcoming PR.

``` console
$ bin/gradualizer test/should_fail/record_exhaustive.erl
test/should_fail/record_exhaustive.erl: Nonexhaustive patterns on line 19 at column 1
Example values which are not covered: ["#record_one_field{inner_rec :: undefined}"]
test/should_fail/record_exhaustive.erl: Nonexhaustive patterns on line 25 at column 1
Example values which are not covered: ["#record_two_fields{a ::\n                                 #record_one_field{inner_rec ::\n                                                       undefined},\n                             b :: integer()}"]
test/should_fail/record_exhaustive.erl: Nonexhaustive patterns on line 47 at column 1
Example values which are not covered: ["#union_rec{field :: c}"]
```

The formatting would need some improvement.

### DIFF
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -47,11 +47,9 @@ format_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
     format_expr_type_error(Expression, ActualType, ExpectedType, Opts);
 format_type_error({nonexhaustive, Anno, Example}, Opts) ->
     FormattedExample =
-        case Example of
-            X when not is_list(X) -> io_lib:format("~p", [X]);
-            %% X is a list, only records are formatted as io_lists,
-            %% Add a newline for readability and formatting
-            X -> ["\n", X]
+        case io_lib:deep_char_list(Example) of
+            true -> ["\n", Example];
+            false -> io_lib:format("~p", [Example])
         end,
     io_lib:format(
       "~sNonexhaustive patterns~s~n"
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -47,13 +47,16 @@ format_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
     format_expr_type_error(Expression, ActualType, ExpectedType, Opts);
 format_type_error({nonexhaustive, Anno, Example}, Opts) ->
     FormattedExample =
-        case io_lib:deep_char_list(Example) of
-            true -> ["\n", Example];
-            false -> io_lib:format("~p", [Example])
+        case Example of
+            [X | Xs] ->
+                lists:foldl(fun(A, Acc) ->
+                    [erl_pp:expr(A), $\n | Acc]
+                end, [erl_pp:expr(X)], Xs);
+            X -> erl_pp:expr(X)
         end,
     io_lib:format(
       "~sNonexhaustive patterns~s~n"
-      "Example values which are not covered: ~s~n",
+      "Example values which are not covered:~n\t~s~n",
       [format_location(Anno, brief, Opts),
        format_location(Anno, verbose, Opts),
        FormattedExample]);
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -46,12 +46,19 @@ format_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
   when is_tuple(Expression) ->
     format_expr_type_error(Expression, ActualType, ExpectedType, Opts);
 format_type_error({nonexhaustive, Anno, Example}, Opts) ->
+    FormattedExample =
+        case Example of
+            X when not is_list(X) -> io_lib:format("~p", [X]);
+            %% X is a list, only records are formatted as io_lists,
+            %% Add a newline for readability and formatting
+            X -> ["\n", X]
+        end,
     io_lib:format(
       "~sNonexhaustive patterns~s~n"
-      "Example values which are not covered: ~p~n",
+      "Example values which are not covered: ~s~n",
       [format_location(Anno, brief, Opts),
        format_location(Anno, verbose, Opts),
-       Example]);
+       FormattedExample]);
 format_type_error({call_undef, Anno, Func, Arity}, Opts) ->
     io_lib:format(
       "~sCall to undefined function ~p/~p~s~n",
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -49,9 +49,11 @@ format_type_error({nonexhaustive, Anno, Example}, Opts) ->
     FormattedExample =
         case Example of
             [X | Xs] ->
-                lists:foldl(fun(A, Acc) ->
-                    [erl_pp:expr(A), $\n | Acc]
-                end, [erl_pp:expr(X)], Xs);
+                lists:reverse(
+                    lists:foldl(fun(A, Acc) ->
+                        [erl_pp:expr(A), $\n | Acc]
+                    end, [erl_pp:expr(X)], Xs)
+                );
             X -> erl_pp:expr(X)
         end,
     io_lib:format(
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -51,7 +51,7 @@ format_type_error({nonexhaustive, Anno, Example}, Opts) ->
             [X | Xs] ->
                 lists:reverse(
                     lists:foldl(fun(A, Acc) ->
-                        [erl_pp:expr(A), $\n | Acc]
+                        [erl_pp:expr(A), "\n\t" | Acc]
                     end, [erl_pp:expr(X)], Xs)
                 );
             X -> erl_pp:expr(X)
--- a/src/gradualizer_lib.erl
+++ b/src/gradualizer_lib.erl
@@ -9,7 +9,7 @@
 -define(type(T), {type, _, T, []}).
 -define(type(T, A), {type, _, T, A}).
 
-%% Prettyprinting Records nesting spaces (or tabs)
+%% Number of space used when prettyprinting records for nesting
 -define(PP_RECORD_NESTING_OFFSET, 2).
 
 %% merge_with for maps. Similar to merge_with for dicts.
@@ -144,6 +144,7 @@ prettyprint(Record) ->
         end,
     prettypr:format(Doc).
 
+-spec pp_record(gradualizer_type:abstract_type()) -> prettyprint:document() | [prettyprint:document()].
 pp_record(?type(record, [{atom, _, RecName} | Fields])) ->
     PPFields = pp_record_fields(Fields),
     %% We return a list instead of a document to achieve "C-style" nesting of tuples/records.
@@ -151,19 +152,20 @@ pp_record(?type(record, [{atom, _, RecName} | Fields])) ->
     %% to align the field nesting with the name of the field;
     %% otherwise, the nesting would be aligned with the name of the record.
     [
-        prettypr:text(["#", atom_to_list(RecName), "{"]),
+        prettypr:text(lists:concat(["#", atom_to_list(RecName), "{"])),
         prettypr:nest(?PP_RECORD_NESTING_OFFSET, PPFields),
         prettypr:text("}")
     ];
 pp_record(?type(field_type, [{atom, _, FieldName}, FieldValue])) ->
     Val = lists:flatten([pp_record(FieldValue)]),
     prettypr:par([
-        prettypr:text([atom_to_list(FieldName), " ="])
+        prettypr:text(atom_to_list(FieldName) ++ " =")
         | Val
     ], 0);
 pp_record(Value) ->
     prettypr:text(io_lib:format("~p", [pick_value(Value)])).
 
+-spec pp_record_fields([gradualizer_type:abstract_type()]) -> prettyprint:document().
 pp_record_fields([X | Tail]) ->
     %% Format all fields with a separator between
     %% The fold of prettypr:beside is not done in one go because
@@ -175,6 +177,7 @@ pp_record_fields([X | Tail]) ->
 pp_record_fields(_) ->
     prettypr:empty().
 
+% -spec pp_record_fields([prettyprint:document()], [prettyprint:document()]) -> prettyprint:document().
 pp_record_fields([], Acc) ->
     Acc;
 pp_record_fields([X, sep | Tail], Acc) ->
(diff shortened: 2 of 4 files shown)


test/should_fail/record_exhaustive.erl
```diff
@@ -0,0 +1,53 @@
...
+two_fields(undefined) -> 0.
+
+%% -spec test5(#record_two_fields{} | undefined) -> integer().
+%% test5(#record_two_fields{a = #record_one_field{field = Field = #inner{}}}) -> Field#inner.field;
+%% %% unhandled
+%% %%test5(#record_two_fields{a = #record_one_field{field = undefined}}) -> 2;
+%% test5(#record_two_fields{b = B}) when is_integer(B) -> B;
+%% %% unhandled
+%% %%test5(#record_two_fields{b = undefined}) -> 1;
+%% test5(undefined) -> 0.
```
josefs, line 40: Let's remove this bit.

josefs (commented):
Thanks for the diff, nice catch.

src/gradualizer_lib.erl
```diff
@@ -109,6 +109,8 @@ pick_value(?type(tuple, any)) ->
     {};
 pick_value(?type(tuple, Tys)) ->
     list_to_tuple([pick_value(Ty) || Ty <- Tys]);
+pick_value(Rec = ?type(record, _)) ->
+    typelib:pp_type(Rec);
```
zuiderkwast, line 113: Pretty-printing the type doesn't seem like the right thing to do here. It should return an example value of the type.

test/should_fail/record_exhaustive.erl
```diff
@@ -0,0 +1,53 @@
...
+    field :: a | b | c
+}).
+-spec union_rec(#union_rec{}) -> integer().
+union_rec(#union_rec{field = a}) ->
+    0;
+union_rec(#union_rec{field = b}) ->
+    1.
+%% unhandled
+%% union_rec(#union_rec{field = c}) ->
+%%     2.
```
zuiderkwast, line 53: Preferably write a comment about what the error message would be instead of the commented-out code.

zuiderkwast (commented):
Nice!

src/gradualizer_lib.erl
berbiche: I went for pp_type to have "pretty" record value printed out (instead of the usual tuple)

`pick_value` is only used in one place:
- https://github.com/josefs/Gradualizer/blob/3e55441b2fbc32f84636ec05b43f487203cddd4f/src/typechecker.erl#L3211
- https://github.com/josefs/Gradualizer/blob/3e55441b2fbc32f84636ec05b43f487203cddd4f/src/gradualizer_fmt.erl#L48-L54

berbiche: Would it make sense to replace the call to `io_lib:format` with some custom logic with `io_lib_pretty:print`?
That way we would have a way to preserve both machine and human consumption of the suggested value.

test/should_fail/record_exhaustive.erl
berbiche: Done

berbiche:
I gave a shot at prettyprinting a plausible record value using the `prettyprint` module:

``` console
$ bin/gradualizer test/should_fail/record_exhaustive.erl
test/should_fail/record_exhaustive.erl: Nonexhaustive patterns on line 24 at column 1
#record_one_field{
  inner_rec = undefined
}

test/should_fail/record_exhaustive.erl: Nonexhaustive patterns on line 38 at column 1
#record_two_fields{
  a = #record_one_field{
    inner_rec = undefined
  },
  b = 0
}

test/should_fail/record_exhaustive.erl: Nonexhaustive patterns on line 78 at column 1
#union_rec{
  field = c
}
```

src/gradualizer_fmt.erl
```diff
@@ -46,12 +46,17 @@ format_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
   when is_tuple(Expression) ->
     format_expr_type_error(Expression, ActualType, ExpectedType, Opts);
 format_type_error({nonexhaustive, Anno, Example}, Opts) ->
+    FormattedExample =
+        case io_lib:deep_char_list(Example) of
+            true -> ["\n", Example];
+            false -> io_lib:format("~p", [Example])
+        end,
```
berbiche, line 53: The `["\n", Example]` is to align the printed record to the leftmost column.

It makes more sense to have this logic here (instead of pick_value), but the downside is that an exhaustive check for a list could return something like `[40]` which would be showed on a newline.

src/gradualizer_lib.erl
```diff
@@ -120,6 +125,67 @@ pick_value(?type(range, [{_TagLo, _, neg_inf}, {_TagHi, _, Hi}])) ->
...
+    Doc =
+        case pp_record(Record) of
+            X when is_list(X) -> lists:foldr(fun prettypr:above/2, prettypr:empty(), X);
+            X -> X
+        end,
+    prettypr:format(Doc).
+
+-spec pp_record(gradualizer_type:abstract_type()) -> prettyprint:document() | [prettyprint:document()].
+pp_record(?type(record, [{atom, _, RecName} | Fields])) ->
+    PPFields = pp_record_fields(Fields),
```
zuiderkwast, line 149: I don't quite see why you would have to implement your own pretty-printing for records. It's already implemented in `erl_pp` (abstract form) and in `io_lib:format("~p", [Rec])`

src/gradualizer_lib.erl
```diff
@@ -109,6 +112,8 @@ pick_value(?type(tuple, any)) ->
     {};
 pick_value(?type(tuple, Tys)) ->
     list_to_tuple([pick_value(Ty) || Ty <- Tys]);
+pick_value(Rec = ?type(record, _)) ->
+    prettyprint(Rec);
```
zuiderkwast, line 116: If this function is not going to pick a value of the given type anymore, you better change its name and change all the other clauses too.

I wouldn't mind changing the error message into something like `"The patterns are non-exhaustive. Values of type #{foo :: integer()} are not covered by any clause"` instead of giving an example *value* which is not covered.

I still don't see why you can't actually pick a record value of the given record type, as is done for all the other types. To do that, you would create a record value like `{record, erl_anno:new(0), RecName, Fields}` where Fields is created using pick_value for each field type in the record type.

src/gradualizer_lib.erl
```diff
@@ -120,6 +125,67 @@ pick_value(?type(range, [{_TagLo, _, neg_inf}, {_TagHi, _, Hi}])) ->
...
+%% Prettyprints a ?type(record)
+%%
+%% What to expect:
+%% ```
+%%   #record_name{
+%%     field_one = #nested_record{
+%%       nested_field = 1
+%%     },
+%%     field_two = some_atom
+%%   }
```
zuiderkwast, line 137: A `?type(record, [...])` is the abstract form of a *record type* but under "What to expect" is a *record value*. So, this function is prettyprinting a record type as if it were a record value?

In a record type, the fields have types but no values, e.g. `#record_name{field_one :: #nested_record{nested_field :: 1 | 5..7 | atom() }}`.

src/typechecker.erl
```diff
@@ -3463,6 +3463,8 @@ refinable(?type(union,Tys)) when is_list(Tys) ->
     lists:all(fun refinable/1, Tys);
 refinable(?type(tuple, Tys)) when is_list(Tys) ->
     lists:all(fun refinable/1, Tys);
+refinable(?type(record, [_ | Fields])) ->
+    lists:all(fun refinable/1, [X || ?type(field_type, X) <- Fields]);
```
zuiderkwast, line 3467: Do the Fields here contain all the fields of the record type or only the refined ones? (@FrankBro do you know?) If now, we would need to pass Env here to be able to get the non-refined ones and check if they are refinable as well.

src/gradualizer_fmt.erl
zuiderkwast: Printing the example on a separate line should rather depend on if it's multiline or not. It could be multiline also if it's large map, tuple, etc. and a record could fit on a single line if it's small.

Besides, I don't like that this can be either already pretty-printed stuff or any Erlang value for two reasons: ambiguity and complexity.

zuiderkwast (changes requested):
I think the overall result is good for the cases you have, e.g. `Example values which are not covered: #union_rec{field = c}`, but I have some concerns about the implementation.

src/gradualizer_lib.erl
berbiche: Thanks, I didn't know about `erl_pp`.

src/gradualizer_lib.erl
berbiche: Given that the values returned by other function clause are not using Erlang's abstract format (hopefully I'm using the right word), it makes little sense for the `record` type clause to be the only one returning a value compatible with the abstract format.

Which solution do we want to take?

1. Return all values as Erlang's abstract format (`pick_value(abstract_type()) -> [erl_parse:abstract_expr()]`) like the proposed example.

2. Return a formatted string (`pick_value(abstract_type()) -> [iolist()]`)

3. ?

src/gradualizer_lib.erl
berbiche: The values that are mapped are obtained by calling `pick_value` for all the record fields.

src/typechecker.erl
berbiche: From the testing I have done, The list of fields has always been empty.

src/gradualizer_lib.erl
zuiderkwast: Ahh, I understand the problem now! Because pick_value returns a simple value, it is not possible to tell a record apart from a tuple, so if we return `#foo{bar = c}` here, it'd be printed as e.g. `{foo, c}`. At first, I didn't even realize it's a problem at all. :facepalm: 

Well then, spontaneously I'm leaning towards (1) since we're working with abstract syntax everywhere (abstract format, parse tree representation, etc. all these terms mean the same).

I don't prefer (2) because it means that we would have to do formatting in every clause. Doing it in one place seems better. It makes it uniform and easier to extend. It's already possible to provide a formatting fun to produce e.g. Elixir code in the error messages, but it's not applied in this particular error. I guess it should in the future (not in scope of this PR).

Another idea (3) could be to just include the type in the error message. Then we don't need pick_value. I gave an example in another comment the other day. We'd have to think about if the error messages will be easy to understand for each of the types, but I think they would. @josefs any opinion?

    test/should_fail/record_exhaustive.erl: Nonexhaustive patterns on line 47 at column 1
    Example values which are not covered: #union_rec{field = c}

vs

    test/should_fail/record_exhaustive.erl: Nonexhaustive patterns on line 47 at column 1
    Type of values which are not covered: #union_rec{field :: c | d | e}

src/typechecker.erl
zuiderkwast: Ah. Then, only the refined fields are included in Fields here.

Given a record definition

```Erlang
-record(rec, {foo :: a | b | c, bar :: float()}).
```

... the type `#rec{}` means the full record (Fields empty) and `#rec{foo :: c}` is a refined record type (Fields contains foo :: c). `bar` is not included because it's the same as in the record definition and is not printed when you print the record type. But because `float()` is not refinable, neither `#rec{}` nor `#rec{foo :: c}` are refinable.

Perhaps it would be good to include a test case with a non-refinable field in a record. If we're not handling it correcty, just place it under known_problems.

src/gradualizer_lib.erl
josefs: Thanks for laying out the options so well, @zuiderkwast .

The way I see it is that printing the type gives maximal information, but might be confusing because the types can look very contrived after refinement. Also, these types don't actually occurr anywhere in the program which adds to the confusion. Providing a concrete value is, at least in my mind, much simpler and easier to understand. It has the problem that we're not providing the full story. 

Using example values in the error messages optimizes for understandability. 

Other languages also tent to print values when reporting exhaustiveness errors, although they might try to print multiple values e.g. [ocaml](https://stackoverflow.com/questions/50801956/fatal-error-exception-match-failuremain-ml-8-15), [haskell](https://stackoverflow.com/questions/54726077/avoid-inappropriate-non-exhaustive-pattern-match-warning-in-ghci). 

Another option (4) would be to print both the refined type and an example value of that type. That would provide maximal information while at the same time provide an example to help understanding the type. A potential downside is information overload in the error message.

My preference is for (1) and (4), I'd like to keep the example values in the error. Of the two, I think I slightly prefer (4) since it offers all the information available. I agree that (2) should be avoided.

zuiderkwast: OK, good points @josefs. So the natural incremental steps would be to first do (1) i.e. make pick_value return values in AST form and then later (4) also print the type. So for this PR, maybe only do (1)?

josefs: Yes, that sounds reasonable.

berbiche: Sorry for the silence, I worked on other project during the holidays.

`pick_value` now returns a value in abstract format that is prettyprinted by `erl_pp` in `format_type_error` (solution 1)

src/gradualizer_fmt.erl
```diff
@@ -46,12 +46,22 @@ format_type_error({type_error, Expression, ActualType, ExpectedType}, Opts)
...
+                    lists:foldl(fun(A, Acc) ->
+                        [erl_pp:expr(A), "\n\t" | Acc]
+                    end, [erl_pp:expr(X)], Xs)
+                );
+            X -> erl_pp:expr(X)
+        end,
     io_lib:format(
       "~sNonexhaustive patterns~s~n"
-      "Example values which are not covered: ~p~n",
+      "Example values which are not covered:~n\t~s~n",
```
berbiche, line 61: Values are now written on a newline with a leading tab.

zuiderkwast (approved):
LGTM!

test/should_fail/record_exhaustive.erl
```diff
@@ -0,0 +1,69 @@
+-module(record_exhaustive).
+
+-compile(export_all).
```
josefs, line 3:
```suggestion
-compile([export_all, nowarn_export_all]).
```

josefs: I've been trying to remove as many warnings from the test suite as possible lately.

josefs (changes requested):
Looks good overall! I just have one small wibble that I'd like you to fix.

test/should_fail/record_exhaustive.erl
berbiche: done

josefs (approved):
Thanks for your contribution!

PR #309, opened 2021-01-02
Remove most warnings when compiling the test suite

josefs:
Warnings can be useful even when compiling the test suite so I've
taken some time to remove as many warnings as I can.

I've replaced some `export` pragmas with
`-compile([export_all,nowarn_export_all])`. I don't think it makes
sense to maintain an export list in the test suite, as there isn't
any abstraction defined in those modules that is in need of protection
by not exporting all functions.

There are a few types of warnings that I haven't been able to
remove, and if someone knows how, I'd be interested in learning:

* "this clause cannot match because a previous clause at line 6 always
  matches"
  In some tests we explicitly want to test overlapping matches, e.g.
  type refinement, so this warning is not helpful.
* "this expression will fail with a 'badarg' exception"
  We want to test guards where the types are not correct to make sure
  that they fail in Gradualizer.
* "this expression will fail with a 'badarith' exception"
  Similar to above, we want to test that typechecking fails when
  arithmetic expressions are given the wrong type of arguments.
* "type bool() is now called boolean(); please use the new name
  instead"
  We want to support the old bool() type so we want to have tests for
  it.

### DIFF
--- a/test/should_pass/scope.erl
+++ b/test/should_pass/scope.erl
@@ -1,6 +1,6 @@
 -module(scope).
 
--compile([export_all]).
+-compile([export_all, nowarn_export_all, nowarn_unused_vars]).
 
 f(X) ->
     case g(X) of
@@ -20,7 +20,7 @@ g(_X) ->
 %     (X = 5) + X.
 
 t() ->
-    { X = 5, Y = 4 },
+    _ = { X = 5, Y = 4 },
 % Fails
 %    { X = 5, Y = 4, X },
     X.
--- a/test/should_pass/try.erl
+++ b/test/should_pass/try.erl
@@ -1,11 +1,11 @@
 -module('try').
 
--compile([export_all]).
+-compile([export_all, nowarn_export_all, nowarn_unused_vars]).
 
 -spec t() -> 2..4.
 t()->
     try A = 1 of
-        ok ->
+        1 ->
             %% a:b(A), unsafe
             B = 2
     catch _:_ ->
@@ -25,7 +25,7 @@ t()->
 
 r()->
     try A = 1 of
-        ok ->
+        1 ->
             %% a:b(A), unsafe
             B = 2
     catch _:_ ->
--- a/test/known_problems/should_fail/guard_fail.erl
+++ b/test/known_problems/should_fail/guard_fail.erl
@@ -1,6 +1,6 @@
 -module(guard_fail).
 
--compile(export_all).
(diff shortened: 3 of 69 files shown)


test/should_pass/try.erl
```diff
@@ -1,11 +1,11 @@
...
 
--compile([export_all]).
+-compile([export_all, nowarn_export_all, nowarn_unused_vars]).
 
 -spec t() -> 2..4.
 t()->
     try A = 1 of
-        ok ->
+        1 ->
             %% a:b(A), unsafe
```
gomoripeti, line 9: not related to this PR, just this part reminded me that `A` here will be safe from OTP 24 (https://github.com/erlang/otp/pull/2662) so will have to update try/catch var env handling (admittedly quite an edge case)

gomoripeti (approved):
quite an effort 👍

gomoripeti:
regarding remaining warnings:
- `bool()` : can be tested with code snippet instead of full compiled modules eg in https://github.com/josefs/Gradualizer/blob/master/test/typechecker_tests.erl#L253
- "this expression will fail with ...": does the compiler complain because we use literals? maybe if we use variables and type info only comes from spec, that's out of the visibility of the compiler but Gradualizer can still see type mismatch eg
```
-spec f(integer(), atom()) -> number().
f(A, B) -> A + B.
```

test/should_pass/try.erl
josefs: Ah, I didn't know that. I've created  #310 to track the issue.

josefs:
Thanks @gomoripeti . Based on your recommendations, I managed to remove the `bool()` warning by putting some tests in `typechecker_tests.erl`. I also managed to get rid of most "this expression will fail with ..." warnings. One was actually due to a broken test, in `tests/should_fail/guard_fail.erl`. But in `test/should_fail/infer_enabled.erl` we are using variables and the compiler still figures out that the addition is going to fail. It's probably possible to make the test more complicated to trick the compiler but I'm going to leave it for now.

test/should_pass/scope.erl
```diff
@@ -20,7 +20,8 @@ g(_X) ->
 %     (X = 5) + X.
 
 t() ->
-    { X = 5, Y = 4 },
+    % Z is just used to avoid a warning that the tuple is not used.
```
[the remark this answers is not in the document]
josefs: Thanks! Fixed.

Issue #311, opened 2021-02-02
Exhaustiveness checking when matching against the same variable in a case of


berbiche:
The code base I'm testing Gradualizer against has the following pattern in some places:

``` erlang
-spec neg_integer(integer()) -> ok.
neg_integer(Number) ->
    case Number of
        0 -> ok;
        Number -> ok
    end.
```
i.e. binding/checking against the same variable name.

Which results in

```
/gradualizer/test/known_problems/should_pass/exhaustiveness_variable_rebinding_in_case_of.erl: Nonexhaustive patterns on line 9 at column 9
Example values which are not covered:
	-1
```

berbiche:
Sorry, duplicate

PR #312, opened 2021-02-02
Add exhaustiveness check of variable match in case of to known_problems/should_pass

berbiche:
See #311


berbiche:
Oups, didn't see #295

PR #314, opened 2021-02-09
Record update missing fields fix

FrankBro:
The record refinement update introduced this bug where it would try to typecheck missing fields even for record updates instead of simply type checking only the fields that are used. Confirmed test was broken before and pass now.

This is also specific to cases where the spec would be missing and the initial record would be compared to `any()`.


josefs (approved):
Awesome!

PR #315, opened 2021-02-09
Any type fix within complex patterns

FrankBro:
We uncovered a really weird bug: 

This gives an error:

`test.erl: The pattern error on line 6 at column 19 doesn't have the type ok`

```erlang
-spec process_response({ok, binary()} | {error, term()}) -> integer().
process_response({error, key_not_found} = _Response) ->
    1;
process_response({error, _} = _Response) ->
    2;
process_response({ok, _} = _Response) ->
    3.
```

While this does not:

```erlang
-spec process_response({ok, binary()} | {error, term()}) -> integer().
process_response({error, key_not_found}) ->
    1;
process_response({error, _}) ->
    2;
process_response({ok, _}) ->
    3.
```

The simple fact of binding the thing to a variable, even if not used, would cause refinement to go wrong for complex types (in this case, tuples and records).

Introduced a simple fix and everything passes.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -45,6 +45,7 @@
 -define(top(), {remote_type, _, [{atom,_,gradualizer}
 				,{atom,_,top},[]]}).
 -define(record_field(Name), {record_field, _, {atom, _, Name}, _}).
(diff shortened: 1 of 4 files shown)


josefs (approved):
Nice catch!

image

PR #319, opened 2021-03-09
Fix var bind in function heads for tuple and records when not spec

berbiche:
If a function is not `spec` and the function head pattern matches against a tuple or a record, the special `none()` type is inferred for the record/tuple expression, but the `any()` type is inferred for the var bind.
These types are then passed down to `glb()` and `glb_ty`, which returns `any()` instead of `none()`, leading to eager "unreachable clause" because the resulting varbind expression `:: any()`.

This bug was discovered thanks to #315.
To test, simply move the `glb()` case for `none()` after the `any()` case and run the `varbind_in_function_head.erl` test.
```
test/should_pass/varbind_in_function_head.erl: The clause on line 8 at column 1 cannot be reached
test/should_pass/varbind_in_function_head.erl: The clause on line 16 at column 1 cannot be reached
```

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -430,6 +430,12 @@ glb(T1, T2, A, TEnv) ->
             end
     end.
 
+%% none() is the bottom of the hierarchy
+glb_ty({type, _, none, []} = Ty1, _Ty2, _A, _TEnv) ->
+    ret(Ty1);
+glb_ty(_Ty1, {type, _, none, []} = Ty2, _A, _TEnv) ->
+    ret(Ty2);
+
 %% We don't know anything if either type is any()
 glb_ty({type, _, any, []} = Ty1, _Ty2, _A, _TEnv) ->
     ret(Ty1);
@@ -441,12 +447,6 @@ glb_ty(?top(), Ty2, _A, _TEnv) ->
     ret(Ty2);
 glb_ty(Ty1, ?top(), _A, _TEnv) ->
     ret(Ty1);
-
-%% none() is the bottom of the hierarchy
-glb_ty({type, _, none, []} = Ty1, _Ty2, _A, _TEnv) ->
-    ret(Ty1);
-glb_ty(_Ty1, {type, _, none, []} = Ty2, _A, _TEnv) ->
-    ret(Ty2);
 
 %% glb is idempotent
 glb_ty(Ty, Ty, _A, _TEnv) ->
--- a/test/should_pass/varbind_in_function_head.erl
+++ b/test/should_pass/varbind_in_function_head.erl
@@ -1,0 +1,17 @@
+-module(varbind_in_function_head).
+
+-compile([export_all, nowarn_export_all]).
+-compile([nowarn_shadow_vars, nowarn_unused_vars]).
+
+a({Key, Value, undefined} = _Unused) ->
+    {Key, Value, 0};
+a({Key, Value, Timestamp}) ->
+    {Key, Value, 2}.
+
+-record(b, {
+    field1 :: atom()
+}).
+b(#b{field1 = undefined} = _Unused) ->
+    1;
+b(#b{field1 = _}) ->
+    2.
--- a/test/typechecker_tests.erl
+++ b/test/typechecker_tests.erl
@@ -174,9 +174,9 @@ glb_test_() ->
 
     [
      %% any()
-     [ ?glb( ?t(any()),  T, ?t(any()) )   || T <- Ts ],
+     [ ?glb( ?t(any()),  T, ?t(any()) )   || T <- Ts, T /= ?t(none()) ],
      [ ?glb( ?t(gradualizer:top()), T, T) || T <- Ts ],
-     [ ?glb( ?t(none()), T, ?t(none()) )  || T <- Ts, T /= ?t(any()) ],
+     [ ?glb( ?t(none()), T, ?t(none()) )  || T <- Ts ],
 
      %% Integer types
      ?glb( ?t(-5..10), ?t(2..3 | 5..15), ?t(2..3 | 5..10) ),


zuiderkwast:
Are you saying that none() is inferred for the pattern {X, Y} in

```Erlang
    f({X, Y}) ->
        ...
```

Shouldn't that be fixed instead...? It should be {_, _} if the `infer` option is used and `any()` otherwise.

Note that `any()` means "not statically typed". It doesn't mean that all values are possible, but rather that it is unknown to the type system at compile time.

berbiche:
> Are you saying that none() is inferred for the pattern {X, Y} in
> 
> ```erlang
>     f({X, Y}) ->
>         ...
> ```
> 
> Shouldn't that be fixed instead...? It should be {_, _} if the `infer` option is used and `any()` otherwise.
> 
> Note that `any()` means "not statically typed". It doesn't mean that all values are possible, but rather that it is unknown to the type system at compile time.

My bad, I used the wrong language because I'm still not entirely familiar with the project, I did not mean inferring nor the `infer` option 🙂 

To reiterate, the fixes in #315 fixed the case where a variable binding occurred for `spec` functions then it would fail to typecheck.
This PR is a continuation of #315 that fixes the case where a function is not `spec`.

zuiderkwast:
Don't worry about terminology. I just want to understand why the fix works. :)

It makes sense that GLB(none, any) = none, and I think that's a good change in itself, but I'm not sure I understand why it fixes the particular issue.

In `a({Key, Value, undefined} = _Unused) ->`, is `{Key, Value, undefined} :: none()` and `_Unused :: any()` and these are unified to `any()` for the whole pattern? And then, `any()` is used as if it meant `top()` so the type checker thinks that the clause matches everything and thus the next clause can never be reached?

It seems strange that a match-all pattern (the unbound variable `_Unused`) is given the type `any()` (which means "some unknown type somewhere between none() and top()"). Shouldn't it be `top()`? (This could possibly be the difference between with/without a spec.)

It seems strange that `{Key, Value, undefined} :: none()`. (I have some vague memory that none() is used as some special value during refinement to say that refinement isn't possible.. But "clause cannot be reached" is exhaustiveness-checking, which IIRC is separate from refinement...) Hm.. I don't fully understand what's happening.

josefs:
I agree with @zuiderkwast here that I think the current patch actually masks the real problem. A key to understanding this problem is that it goes away if we remove the match pattern in the first clause of the examples. So I suspect the problem is there. I've spent a bit of time trying to find the problem but haven't made much progress yet.

berbiche:
I've added prints at different places and I think the issue lies in that all fields of a tuple are `:: any()` and so are elements of a list and fields in records.

```
Checking module varbind_in_function_head
Checking function a/1
6:1: Checking clause :: any()
7:5: Checking that {Key, Value, 0} :: any()
7:6: Propagated type of Key :: any()
7:11: Propagated type of Value :: any()
7:18: Propagated type of 0 :: any()
7:5: Propagated type of {Key, Value, 0} :: any()
refine: OrigTy: {type,0,tuple,[{type,0,any,[]}]} - Ty: {type,0,tuple,
                                                        [{type,0,any,[]}]}
refine: normalize: {type,0,tuple,[{type,0,any,[]}]}
refine: refine_ty_normalize: {type,0,none,[]} - Ty: {type,0,tuple,[{type,0,any,[]}]} - OrigTy: {type,0,tuple,[{type,0,any,[]}]} - NormTy: {type,0,tuple,[{type,0,any,[]}]}
Checking function b/1
16:1: Checking clause :: any()
17:5: Checking that 1 :: any()
17:5: Propagated type of 1 :: any()
refine: OrigTy: {type,0,tuple,[{type,0,any,[]}]} - Ty: {type,0,tuple,
                                                        [{type,0,any,[]}]}
refine: normalize: {type,0,tuple,[{type,0,any,[]}]}
refine: refine_ty_normalize: {type,0,none,[]} - Ty: {type,0,tuple,[{type,0,any,[]}]} - OrigTy: {type,0,tuple,[{type,0,any,[]}]} - NormTy: {type,0,tuple,[{type,0,any,[]}]}
Checking function c/1
21:1: Checking clause :: any()
22:5: Checking that 1 :: any()
22:5: Propagated type of 1 :: any()
refine: OrigTy: {type,0,tuple,[{type,0,any,[]}]} - Ty: {type,0,tuple,
                                                        [{type,0,any,[]}]}
refine: normalize: {type,0,tuple,[{type,0,any,[]}]}
refine: refine_ty_normalize: {type,0,none,[]} - Ty: {type,0,tuple,[{type,0,any,[]}]} - OrigTy: {type,0,tuple,[{type,0,any,[]}]} - NormTy: {type,0,tuple,[{type,0,any,[]}]}
Checking function e/1
31:1: Checking clause :: any()
32:5: Checking that 1 :: any()
32:5: Propagated type of 1 :: any()
refine: OrigTy: {type,0,tuple,[{type,0,any,[]}]} - Ty: {type,0,tuple,
                                                        [{integer,0,1}]}
refine: normalize: {type,0,tuple,[{type,0,any,[]}]}
refine_ty: Tys1: [{type,0,any,[]}] - Tys2: [{integer,0,1}]
refine: OrigTy: {type,0,any,[]} - Ty: {integer,0,1}
refine: normalize: {type,0,any,[]}
33:1: Checking clause :: any()
34:5: Checking that 2 :: any()
34:5: Propagated type of 2 :: any()
refine: OrigTy: {type,0,tuple,[{type,0,any,[]}]} - Ty: {type,0,tuple,
                                                        [{type,0,any,[]}]}
refine: normalize: {type,0,tuple,[{type,0,any,[]}]}
refine: refine_ty_normalize: {type,0,none,[]} - Ty: {type,0,tuple,[{type,0,any,[]}]} - OrigTy: {type,0,tuple,[{type,0,any,[]}]} - NormTy: {type,0,tuple,[{type,0,any,[]}]}
test/should_pass/varbind_in_function_head.erl: The clause on line 8 at column 1 cannot be reached
test/should_pass/varbind_in_function_head.erl: The clause on line 18 at column 1 cannot be reached
test/should_pass/varbind_in_function_head.erl: The clause on line 23 at column 1 cannot be reached
```

When the expression is type checked in `do_type_check_expr`, all fields of the tuple/list/records are `any` so it returns `any()` instead.

The function `e` in the prints above is defined as:

``` erlang
e(1 = _Unused) ->
    1;
e(_) ->
    2.
```

I believe the issue lies in the fact that fields of tuples and records, and element of cons are refined? to `any`.

berbiche:
> It makes sense that GLB(none, any) = none, and I think that's a good change in itself, but I'm not sure I understand why it fixes the particular issue.
> 
> In `a({Key, Value, undefined} = _Unused) ->`, is `{Key, Value, undefined} :: none()` and `_Unused :: any()` and these are unified to `any()` for the whole pattern? 

This is correct.

> And then, `any()` is used as if it meant `top()` so the type checker thinks that the clause matches everything and thus the next clause can never be reached?

No, this is wrong. What happens is the following (**without the glb change**):

1. `check_clause` on the first clause of the function

2. `add_types_pats`

3. `add_type_pat (match(tuple(), var()))`

4. `add_type_pat_var (Pattern: none(), PatternVariable: any())` -> `add_type_pat`

5.
    a) `glb_ty(none, any) -> any()`
        The pattern retains its `none()` type (the pattern's type cannot be changed in `add_type_pat`) whereas the pattern variable gets `any()`.
    
    b) `glb_ty(none, any) -> none()` with the `glb_ty` changes in 23f22a0.
        With the `glb_ty` changes, the pattern variable has type :: none(). This matters later on.

6. ... checking of return type

7. Still in `check_clause`: `ArgsTy :: [any()]` and `PatTys :: [any()]`, with the glb change, `PatTys :: [none()]`

8. `refine_clause_arg_tys(Tys: [any()], MatchedTys: [any()])` or `MatchedTys: [none()]` with the changes

9. `type_diff(Ty: tuple([any()]), MatchedTy: tuple([any()])` calls `refine` and in turn calls `refine_ty`

10. 
    a) `refine_ty` is invoked with the same types (`any()`, `any()`) so it returns `none()`
    `refine` returns the result to `type_diff` and `type_diff` to `refine_clause_arg_tys`.
    `refine_clause_arg_tys` matches on the returned type, it forwards the `none()` type it receives. 
    
    Refinement of the first clause: `none()`

    b) 
    This is where the change to glb matters.

    With the glb change, `refine_ty` is invoked with `any()` and `none()`.
    The types differ and `none` is present, so it throws `no_refinement`.
    After throwing `no_refinement`, `type_diff` returns the first type, `tuple(any())`.
    
    Refinement of the first clause: `any()`

11. `check_clause` continues and checks the next clause of the function
12.
    a) Without the change, it sees that the type of the last clause is `none()` and throws an `unreachable_clause` error.

    b) With the change, the type of the last clause is `any()`, so it continues and does the same procedure starting from 1. 

> It seems strange that a match-all pattern (the unbound variable `_Unused`) is given the type `any()` (which means "some unknown type somewhere between none() and top()"). Shouldn't it be `top()`? (This could possibly be the difference between with/without a spec.)

I've tried different fixes:
- Returning `{type(none), type(none), VEnv, Constraints)` in `add_type_pat_var` for tuples, cons/list and records when the type is `any()`. Some unit tests broke because some `match()`es where non-exhaustive.
- Handling `tuple(any()), tuple(any())` in `refine_ty`. This completely broke the unreachable clause check.

Ultimately, I'm not sure what the correct fix is and where the problem really lies:
- Should the pattern variable have type `none()` like the pattern besides it?
- Should the pattern variable have type `top()`?
- ?

berbiche:
Hi @zuiderkwast @josefs,

Any idea as to what would be the most appropriate fix for https://github.com/josefs/Gradualizer/pull/319#issuecomment-808370252 😃 ?

josefs:
@berbiche thanks for the thorough investigation. I've now spent some time looking in to this as well and tried to understand the situation and the tl;dr is that I think your original PR is the correct one. 
Initially I wasn't convinced that the original PR was actually correct so I went back to the paper ["Abstracting Gradual Typing"](https://pleiad.cl/papers/2016/garciaAl-popl2016.pdf) to figure out what glb should return. And the answer was clearly stated on top of page 6. glb(any(), none()) = none().
As for the rest of the refinement machinery, it seems to be working correctly. I've spent a fair bit of time trying to see if there was a problem there but I haven't been able to identify any errors.

Sorry for making you jump through all these hoops when your original PR was good, @berbiche .

Can you please remove the diff with all the printing. Then I'll happily accept it.

berbiche:
@josefs done! 😄

josefs (approved):

PR #321, opened 2021-03-17
easy: dead code

ilya-klyuchnikov:
1. When being invoked from CL the branch with type_check_dir is never taken - since type_check_file_or_dir is called first
2. This branch is just logically incorrect - since type_check_dir doesn't return {ok, Forms} which is expected by the outer case


zuiderkwast (approved):
LGTM.

Are the rest of the functions in this module correct/alive/useful?

Issue #322, opened 2021-03-25
Gradualizer may fall into infinite loop for recursive types.


ilya-klyuchnikov:
A repro:

```erlang
-module(recursive_types).

-compile([export_all, nowarn_export_all]).

-type rec1(A) :: A | {rec, rec1(A)}.

-spec unwrap(rec1(rec1(atom()))) -> rec1(atom()).
unwrap({rec, Z}) -> Z.
```

This goes into loop:
```
./bin/gradualizer recursive_types.erl
```

gomoripeti:
very nice minimal example
This reminds me of https://github.com/josefs/Gradualizer/pull/135, apparently the cache did not solve all cases :(

mheiber:
Not sure if a similar example has been covered already, but I'm seeing an infinite loop with the following code:

```erl
-module(sample).
-compile([export_all]).

-type t() :: t().

-spec main() -> t().
main() -> ok.
```

tested with `208f5816b0157f282212fc036ba7560f0822f9fc` (master)

zuiderkwast:
@mheiber thanks for this example. I think it's a different issue. Types can refer to itself inside a tuple etc. but it's allowed to be recursive on the top level, so we should detect it and give an error for this.

mheiber:
Thanks for confirming, @zuiderkwast . Would it be better for me to open a separate issue, or would you prefer to keep these recursive types issues together?

zuiderkwast:
Separate please (unless you think it will be solved by the same fix??)

mheiber:
> Separate please (unless you think it will be solved by the same fix??)

I think both issues have to do with checking subtype or type equivalence for recursive types–I suspect they will have a common solution.

PR #323, opened 2021-03-27
do not normalize types recursively (#fixes 322)

ilya-klyuchnikov:
For the `recursive_types.erl` example, - the problem has been tracked down to recursive normalization of types in `typechecker:normalize/2` when normalizing type aliases (user defined types).

Stop recursive normalization is a possible fix against going into infinite loop.

Restricting normalization this way, however, makes `type_pattern:f` example fail (because the "inner alias" is not expanded/normalized anymore). - Moving it into `known_problems` then.

```
known_problems/should_pass/type_pattern.erl: 
The pattern "foo" on line 14 at column 3 doesn't have the type [mychar()]
```

While `recursive_types/unwrap/1` from #322 stops falling into the infinite loop of normalization, it's not treated as well typed by Gradualizer - adding it to `known_problems` too.

```
recursive_types.erl: The variable on line 8 at column 24 is expected to have type 
rec(atom()) 
but it has type 
rec(atom()) | {rec, rec(rec(atom()))}

-spec unwrap(rec(rec(atom()))) -> rec(atom()).
unwrap({rec, Elem}) -> Elem.
                       ^^^^
```

----

A bit more context about `recursive_types` example: a similar example is accepted by Typed Racket as well-typed:

```racket
#lang typed/racket

(define-type
  (MyRec a)
  (U a (Pair 'rec (MyRec a))))

(: unwrap (-> (MyRec (MyRec Symbol)) (MyRec Symbol)))
(define (unwrap my_rec)
  (match my_rec [(cons 'rec elem) elem]))
```

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -638,8 +638,7 @@ normalize({user_type, P, Name, Args} = Type, TEnv) ->
             case TEnv#tenv.types of
                 #{TypeId := {Vars, Type0}} ->
                     VarMap = maps:from_list(lists:zip(Vars, Args)),
-                    Type1 = typelib:substitute_type_vars(Type0, VarMap),
-                    normalize(Type1, TEnv);
+                    typelib:substitute_type_vars(Type0, VarMap);
                 _NotFound ->
                     throw({undef, user_type, P, {Name, length(Args)}})
             end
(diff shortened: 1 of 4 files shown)


josefs (changes requested):
Thanks for the effort @ilya-klyuchnikov . I'm not a big fan making other program fail as a result of this change so I made #324 which supercedes this PR. That PR doesn't fail on currently working programs and it fixed the infinite loop, but it still doesn't accept the program you gave.

PR #324, opened 2021-03-29
Make sure not to recurse forever in typechecker:normalize()

### LINKED ISSUE #322: Gradualizer may fall into infinite loop for recursive types.

josefs:
Given a recursive type, typechecker:normalize() can recurse forever if
we don't take care to stop the recursion. This patch adds a set to
keep track of the types that `normalize` has unfolded so far and stops
unfolding if asked to normalize a type that has already been unfolded.

@ilya-klyuchnikov  identified the problem with recursives types in #322 and offered
a nice example program to demonstrate the problem.

Fixes #322 although the program is still not accepted as it should be
so I'm adding it to the known problems.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -608,25 +608,36 @@ glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
 %% * Replace built-in type synonyms
 %% * Flatten unions and merge overlapping types (e.g. ranges) in unions
 -spec normalize(type(), TEnv :: #tenv{}) -> type().
-normalize({type, _, record, [{atom, _, Name}|Fields]}, TEnv) when length(Fields) > 0 ->
-    NormFields = [type_field_type(FieldName, normalize(Type, TEnv)) 
+normalize(Ty, TEnv) ->
+    normalize_rec(Ty, TEnv, sets:new()).
+
+%% The third argument is a set of user types that we've already unfolded.
+%% It's important that we don't keep unfolding such types because it will
+%% lead to infinite recursion.
+normalize_rec({type, _, record, [{atom, _, Name}|Fields]}, TEnv, Unfolded)
+  when length(Fields) > 0 ->
+    NormFields = [type_field_type(FieldName, normalize_rec(Type, TEnv, Unfolded)) 
         || ?type_field_type(FieldName, Type) <- Fields],
     type_record(Name, NormFields);
-normalize({type, _, union, Tys}, TEnv) ->
-    Types = flatten_unions(Tys, TEnv),
+normalize_rec({type, _, union, Tys}, TEnv, Unfolded) ->
+    Types = flatten_unions(Tys, TEnv, Unfolded),
     case merge_union_types(Types, TEnv) of
         []  -> type(none);
         [T] -> T;
         Ts when length(Ts) > ?union_size_limit -> type(any); % performance hack
         Ts  -> type(union, Ts)
     end;
-normalize({user_type, P, Name, Args} = Type, TEnv) ->
-    case typelib:get_module_from_annotation(P) of
+normalize_rec({user_type, P, Name, Args} = Type, TEnv, Unfolded) ->
+    case sets:is_element({P,Name}, Unfolded) of
+      true -> Type;
+      false -> 
+      UnfoldedNew = sets:add_element({P,Name}, Unfolded),
+      case typelib:get_module_from_annotation(P) of
         {ok, Module} ->
             %% Local type in another module, from an expanded remote type
             case gradualizer_db:get_type(Module, Name, Args) of
                 {ok, T} ->
-                    normalize(typelib:remove_pos(T), TEnv);
+                    normalize_rec(typelib:remove_pos(T), TEnv, UnfoldedNew);
                 opaque ->
                     Type;
                 not_found ->
@@ -639,18 +650,19 @@ normalize({user_type, P, Name, Args} = Type, TEnv) ->
                 #{TypeId := {Vars, Type0}} ->
                     VarMap = maps:from_list(lists:zip(Vars, Args)),
                     Type1 = typelib:substitute_type_vars(Type0, VarMap),
-                    normalize(Type1, TEnv);
+                    normalize_rec(Type1, TEnv, UnfoldedNew);
                 _NotFound ->
                     throw({undef, user_type, P, {Name, length(Args)}})
             end
+      end
     end;
-normalize(T = ?top(), _TEnv) ->
+normalize_rec(T = ?top(), _TEnv, _Unfolded) ->
     %% Don't normalize gradualizer:top().
     T;
-normalize({remote_type, P, [{atom, _, M}, {atom, _, N}, Args]}, TEnv) ->
+normalize_rec({remote_type, P, [{atom, _, M}, {atom, _, N}, Args]}, TEnv, Unfolded) ->
     case gradualizer_db:get_exported_type(M, N, Args) of
         {ok, T} ->
-            normalize(typelib:remove_pos(T), TEnv);
+            normalize_rec(typelib:remove_pos(T), TEnv, Unfolded);
         opaque ->
             typelib:annotate_user_types(M, {user_type, P, N, Args});
         not_exported ->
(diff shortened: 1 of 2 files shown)


ilya-klyuchnikov:
Just FYI - #323 - a naive fix to not get into the loop. (I have not read this PR yet fully)

src/typechecker.erl
```diff
@@ -608,25 +608,36 @@ glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
...
-normalize({type, _, record, [{atom, _, Name}|Fields]}, TEnv) when length(Fields) > 0 ->
-    NormFields = [type_field_type(FieldName, normalize(Type, TEnv)) 
+normalize(Ty, TEnv) ->
+    normalize_rec(Ty, TEnv, sets:new()).
+
+%% The third argument is a set of user types that we've already unfolded.
+%% It's important that we don't keep unfolding such types because it will
+%% lead to infinite recursion.
+normalize_rec({type, _, record, [{atom, _, Name}|Fields]}, TEnv, Unfolded)
+  when length(Fields) > 0 ->
```
zuiderkwast, line 618: Don't we need a base case to actually stop recursion? (not as a guard expression but you know what I mean)

```Erlang
normalize_rec(Type, _TEnv, Unfolded) when sets:is_element(Type, Unfolded) ->
    Type;
```

zuiderkwast (commented):
Nice! I think this is the way. It's the same way we prevent infinite recursion in subtype checking.

Perhaps just it needs some small tweak to solve the known problem..?

src/typechecker.erl
```diff
@@ -608,25 +608,36 @@ glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
...
     case merge_union_types(Types, TEnv) of
         []  -> type(none);
         [T] -> T;
         Ts when length(Ts) > ?union_size_limit -> type(any); % performance hack
         Ts  -> type(union, Ts)
     end;
-normalize({user_type, P, Name, Args} = Type, TEnv) ->
-    case typelib:get_module_from_annotation(P) of
+normalize_rec({user_type, P, Name, Args} = Type, TEnv, Unfolded) ->
+    case sets:is_element({P,Name}, Unfolded) of
```
ilya-klyuchnikov, line 631: `{P,Name}` may seem a bit strange identity for a type being unfolded. `P` is a (modified) annotation, - which may depend of how Gradualizer uses/stores/handles annotations.

Eg. - running under debugger reveals the this `P` is always `0`, - meaning that all types with the same name will be put into the same "basket".

Probably, not only the name, but also the arity of the type should be considered in this identity - since Erlang allows to have types with the same name by different arities.

An example:

With this diff, this snippet is type-checked:

```erlang
-module(arity_overloading).

-type pair(A, B) :: {A, B}.
-type pair1(A) :: pair(A, A).

-spec foo(pair1(pair(atom(), atom()))) -> {atom(), atom()}.
foo({X, _}) -> X.
```

But changing `pair1` -> `pair`, makes it ill-typed. (Gradualizer doesn't distinguish between `pair/1` and `pair/2`).

```erlang
-module(arity_overloading).

-type pair(A, B) :: {A, B}.
-type pair(A) :: pair(A, A).

-spec foo(pair(pair(atom(), atom()))) -> {atom(), atom()}.
foo({X, _}) -> X.
```

```
-module(arity_overloading).

-type pair(A, B) :: {A, B}.
-type pair(A) :: pair(A, A).

-spec foo(pair(pair(atom(), atom()))) -> {atom(), atom()}.
foo({X, _}) -> X.
```

```
./bin/gradualizer arity_overloading.erl

arity_overloading.erl: The pattern {X, _} on line 9 at column 5 doesn't have the type pair(pair(atom(), atom()))
```

ilya-klyuchnikov (approved):
~~Yes, this should definitely prevent infinite unfolding in `normalize`.~~

Update: I was wrong

ilya-klyuchnikov (changes requested):
Found another counter-example - now with records.

ilya-klyuchnikov:
Just thinking about @zuiderkwast's remarks, this example makes gradualizer fall into the loop.

```erlang
-module(recursive_record).

-type rec(A) :: A | #rec{rec :: A | rec(A)}.
-record(rec, {rec :: any()}).

-spec unwrap(rec(#rec{rec :: A})) -> rec(A).
unwrap(#rec{rec = Rec}) -> Rec.
```

```
./bin/gradualizer recursive_record.erl
```

src/typechecker.erl
zuiderkwast: Ah, sorry, this is about recursive user-defined types only.

src/typechecker.erl
zuiderkwast: Right, shouldn't we store `{Name, Args}` in the set instead?

zuiderkwast:
@ilya-klyuchnikov are you saying we also need the set mechanism for recursive `-type`s?

ilya-klyuchnikov:
> are you saying we also need the set mechanism for recursive records?

Yes, - in this example with records - the loop happens at line 619. - Where a (parameterized) record is normalized.

josefs:
> Perhaps just it needs some small tweak to solve the known problem..?

Yes, I think there should be a way to fix what I have to make the counter example of @ilya-klyuchnikov typecheck. Not sure exactly what that tweak is though.

src/typechecker.erl
josefs: Yes, you're both right that there needs to be some arity information in the set. I'll update the PR. However, adding `Args` to the set as you suggest @zuiderkwast wouldn't be correct, I think. Consider the type below:

```
-type apa(A) :: A | apa({A}).
```

This type keeps changing the type of its argument so that we get an increasing level of nesting that we have never seen before and the normalization would go on forever.

zuiderkwast: True. Unless A is unexpanded... I'm not sure. :-)

josefs:
@ilya-klyuchnikov I'm too stupid to figure out why your recursive record example loops. To me it seems like if it's going to loop forever it needs to unfold the recursive type forever, and that should be taken care of by the mechanism that I've implemented in this PR. However, that's clearly not the case. But I need help figuring out where exactly the problem is.

ilya-klyuchnikov:
It's also possible to loop without type parameters:

```erlang
-module(even_more_recursive_types).

-type rec1() :: rec1 | {rec1, rec1()} | {rec1, {rec1, rec1()}}.

-spec unwrap(rec1()) -> rec1().
unwrap({_, Z}) -> Z.
```

erszcz:
Having tracked down one of the infinite loops to a problem in `glb` - https://github.com/josefs/Gradualizer/pull/356 - I tried this PR rebased on top of the `glb` one - https://github.com/erszcz/Gradualizer/tree/wip-recursive-types. I also made sure the set carried through `normalize_rec` stores type arity in case of user types. Alas, I think there's more than one problem here. I added all the code snippets from this PR to the repo (sorry for the mess, but it's just a quick hack now).

Here's a summary of the tests:
- `norm_test.erl` is the one from the glb PR - this works fine
- `norm_test2.erl` is [@josefs's `-type apa(A) :: A | apa({A}).`](https://github.com/josefs/Gradualizer/pull/324#discussion_r603636005) - this also works fine
- `loop_test3.erl` is [@ilya-klyuchnikov's recursive records example](https://github.com/josefs/Gradualizer/pull/324#issuecomment-809596764) - this seems to loop somewhere in `typechecker:flatten_unions`
- `loop_test4.erl` is [@ilya-klyuchnikov's looping without records](https://github.com/josefs/Gradualizer/pull/324#issuecomment-810393618) - this loops in `typelib:remove_pos`
- `loop_test5.erl` is [@ilya-klyuchnikov's `even_more_recursive_types`] - this seems to be a bigger cycle with `normalize` and `glb`

I think the test cases listed in this thread uncover more than one place where we hit an infinite loop. It also seems to me that the `sets:is_element` check in just one of `normalize_rec` clauses is not enough - I think we have to make that check in each clause which calls `normalize_rec` recursively - similarly to how `refinable` does it with `stop_refinable_recursion`.

All in all, there's more work to be done here, but there's some light at the end of the tunnel. The `dbg` invocation I use is available on the `wip-recursive-types` I link to in the first paragraph.

erszcz:
I don't have more updates than my former comment here, but I'd like to take a closer look at this when time permits. So far it seems that the loop in `normalize` is crucial, but that there might be more looping functions throughout the code. I'll post fixes as I come up with any :+1:

erszcz:
I'm closing this because of #458 already being merged.

PR #325, opened 2021-03-29
Use epp:open/1 for OTP-24+

lpgauth:
`epp:open/5` was removed in OTP-24 (https://github.com/erlang/otp/commit/5281a8c7f77d45a3c36fca9c1a2e4d3812f6fc3d#diff-580a349c49b1d9b5415166e18f5279728d934efe0cebc4ee5a87823055ec3413). This fix checks if `epp:open/5` is exported otherwise it uses `epp:open/1` instead.

### DIFF
--- a/src/gradualizer_file_utils.erl
+++ b/src/gradualizer_file_utils.erl
@@ -37,7 +37,7 @@ epp_parse_file(File, Includes) ->
         {ok, Fd} ->
             try
                 StartLocation = {1, 1},
-                case epp:open(File, Fd, StartLocation, Includes, []) of
+                case epp_open(File, Fd, StartLocation, Includes) of
                     {ok, Epp} ->
                         %% The undocumented `epp:parse_file/1' just
                         %% takes an internal state, and calls the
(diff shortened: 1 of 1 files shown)


lpgauth:
A bit puzzled by the `cli-tests` failing in Travis...

zuiderkwast:
Looks good!

> A bit puzzled by the `cli-tests` failing in Travis...

Yes, I'm puzzled too. Where it fails, it doesn't print the column number. Can it be that `erlang:function_exported(epp, open, 5)` has returned `false`? (shouldn't be...)

src/gradualizer_file_utils.erl
```diff
@@ -56,6 +56,20 @@ epp_parse_file(File, Includes) ->
             Error1
     end.
 
+% epp:open/5 was removed in OTP-24
+% https://github.com/erlang/otp/commit/5281a8c7f77d45a3c36fca9c1a2e4d3812f6fc3d#diff-580a349c49b1d9b5415166e18f5279728d934efe0cebc4ee5a87823055ec3413
+epp_open(File, Fd, StartLocation, Includes) ->
+    code:atomic_load([epp]),
```
zuiderkwast, line 62: Nice! This should do. Perhaps slightly cleaner:

```suggestion
    code:ensure_loaded(epp),
```

zuiderkwast (approved):
LGTM

src/gradualizer_file_utils.erl
lpgauth: @zuiderkwast yep, that's cleaner.

josefs (approved):

PR #327, opened 2021-04-15
Fix record refinement attempt when not all field types are refined

erszcz:
When playing with an application of the typestate pattern in Erlang for a blog post, I encountered this bug. I'm positively surprised that with the fix applied, everything works fine - the pattern works with Gradualizer as the typechecker!

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -609,7 +609,7 @@ glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
 %% * Flatten unions and merge overlapping types (e.g. ranges) in unions
 -spec normalize(type(), TEnv :: #tenv{}) -> type().
 normalize({type, _, record, [{atom, _, Name}|Fields]}, TEnv) when length(Fields) > 0 ->
-    NormFields = [type_field_type(FieldName, normalize(Type, TEnv)) 
+    NormFields = [type_field_type(FieldName, normalize(Type, TEnv))
         || ?type_field_type(FieldName, Type) <- Fields],
     type_record(Name, NormFields);
 normalize({type, _, union, Tys}, TEnv) ->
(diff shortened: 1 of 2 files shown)


test/should_pass/record_refinement.erl
```diff
@@ -78,3 +78,9 @@ two_level2(#two_level1{two_level2 = #two_level2{value = undefined}}) ->
     0;
 two_level2(R1) ->
     R1#two_level1.two_level2#two_level2.value.
+
+-record(not_all_fields_refined, {a, b}).
+-type not_all_fields_refined() :: #not_all_fields_refined{b :: b | c}.
+
+-spec not_all_fields_refined() -> not_all_fields_refined().
```
zuiderkwast, line 85: I think it's a tiny bit confusing to use the same name for a function and a type.

src/typechecker.erl
```diff
@@ -1145,8 +1145,10 @@ expect_record_type({type, _, record, [{atom, _, Name}|RefinedTypes]}, Record, #t
                 case RefinedTypes of
                     [] -> Fields;
                     _ ->
-                        [ {typed_record_field, RecordField, RefinedType}
-                        || {{typed_record_field, RecordField, _}, ?type_field_type(_, RefinedType)} <- lists:zip(Fields, RefinedTypes)]
+                        %% `RefinedTypes' and `Fields' might be of different sizes
+                        RefinedMap = maps:from_list([ {N, Ty} || ?type_field_type(N, Ty) <- RefinedTypes ]),
```
zuiderkwast, line 1149: The name `N` made me think it's a number, but it's a name, right? Please use a different name (such as `FieldName`).

zuiderkwast (commented):
Very nice!

I have a slight memory that there was some remaining TODO in this area. @FrankBro, you implemented this refinement, right? Mind having a look?

FrankBro:
Would love to see a test that used to fail before and passes now to visualize what was fixed.

erszcz:
@zuiderkwast Good points, fixed up!

zuiderkwast (approved):

erszcz:
BTW, I've just the blog post I mentioned:
https://erszcz.medium.com/make-illegal-states-unrepresentable-but-how-the-typestate-pattern-in-erlang-16b37b090d9d

PR #328, opened 2021-05-05
Fix gradualizer_db resolving local types in the context of remote types

### LINKED ISSUE #326: Local types are resolved as remote types when using generics
I couldn't find any issue detailing this bug but given the following files:

`user_types.erl`
``` erlang
-module(user_types).

-type my_atom() :: atom().
-type my_shadowed_local() :: user_types.
-type my_generic(T) :: my_atom() | {ok, T}.

-export_type([my_generic/1]).
```

and 

`remote_types.erl`
``` erlang
-module(remote_types).

-type my_shadowed_local() :: remote_types.

-spec my_function(atom()) -> user_types:my_generic(my_shadowed_local()).
my_function(undefined) -> error.
my_function(_) -> {ok, remote_types}.
```

This fails to typecheck because `my_shadowed_local` is resolved in the "context" of the `user_types` module.

**My understanding**

```
62:1: Checking clause :: user_types:my_generic(my_shadowed_type())
62:47: Checking that {ok, shadowed} :: user_types:my_generic(my_shadowed_type())
[type_check_expr_in:2063] {remote_type,0,
                              [{atom,{61,53},user_types},
                               {atom,{61,64},my_generic},
                               [{user_type,0,my_shadowed_type,[]}]]}
```

Starting from `type_check_expr_in(Env, ResTy, Expr)`, the `ResTy` does not have any information about the provenance of the `user_type`.

`normalize` is then called before going into `do_type_check_expr_in`.
`[normalize:655] get_exported_type{remote_type, M :: user_types, N :: my_generic, Args :: [{user_type, 0, my_shadowed_type, []}]}`.

We make it to `gradualizer_db` to get the exported type:
`[handle_call:174] (get_exported_type) M :: user_types, T :: my_generic, Args :: [{user_type,0,my_shadowed_type,[]}`,
which handles the call and ends up in `handle_get_type`.

`handle_get_type` finds the exported type `my_generic` and calls `typelib:substitute_type_vars`:

``` erlang
{type,16,union,
             [{user_type,16,my_atom,[]},
              {type,16,tuple,
                    [{atom,16,ok},{user_type,0,my_shadowed_type,[]}]}]}
```

It then calls `typelib:annotate_user_types(ModuleOfTheExportedType, Type)`:

``` erlang
{type,16,union,
             [{user_type,[{file,"user_types.erl"},{location,16}],my_atom,[]},
              {type,16,tuple,
                    [{atom,16,ok},
                     {user_type,[{file,"user_types.erl"},{location,0}],
                                my_shadowed_type,[]}]}]}
```

Where `annotate_user_types` simply adds the filename from the module to all the types:

``` erlang
annotate_user_types(Module, Type) when is_atom(Module) ->
    annotate_user_types(atom_to_list(Module) ++ ".erl", Type);
annotate_user_types(Filename, {user_type, Anno, Name, Params}) ->
    %% Annotate local user-defined type
    {user_type, erl_anno:set_file(Filename, Anno), Name,
     [annotate_user_types(Filename, Param) || Param <- Params]};
...
```

edit: I think this behavior exists to be able to find `my_atom()` type in the example above.

My takeaway is that when `gradualizer_db` is interrogated, it doesn't have enough information about the origin of a `user_type`.

I'm currently experimenting with the following solution:
1. Annotate all types with a filename when parsing forms.
2. Pass TEnv to `gradualizer_db:get_exported_type` to retrieve the types and their annotation.

``` erlang
{tenv,remote_types,
    #{},
      {local_list,0} =>
          {[],
           {type,
               [{file,"remote_types.erl"},{location,0}],
               list,
               [{type,0,atom,[]}]}},
      {local_record_with_remote_type,0} =>
      {my_shadowed_type,0} => {[],{atom,0,shadowed}}}
}
```

berbiche:
Fixes #326

I found a fix that was much much simpler than described in #326.
Essentially, reordering the operations in `get_type` to expand generics after resolving annotated types
made it more lazy.
The function that annotates types does not recurse in remote types so nested generics worked but not
if they were using a type with a similar name.
By expanding generics later, it does not recurse nor incorrectly annotate user types from the local module.

Three test case were added as well as a dummy module for tests:
1. The first test for nesting generics from different modules.
2. The second test for generics with an opaque type.
3. The third test for nesting generics within the same module.

Should I add a test case for records?

### DIFF
--- a/src/gradualizer_db.erl
+++ b/src/gradualizer_db.erl
@@ -315,8 +315,8 @@ handle_get_type(M, T, Args, RequireExported, ExpandOpaque, State) ->
                  #typeinfo{params = Vars,
                            body = Type0} ->
                      VarMap = maps:from_list(lists:zip(Vars, Args)),
-                     Type1 = typelib:substitute_type_vars(Type0, VarMap),
-                     Type2 = typelib:annotate_user_types(M, Type1),
+                     Type1 = typelib:annotate_user_types(M, Type0),
+                     Type2 = typelib:substitute_type_vars(Type1, VarMap),
                      {reply, {ok, Type2}, State}
              end;
         _NoMatch ->
--- a/test/should_pass/other_module.erl
+++ b/test/should_pass/other_module.erl
@@ -1,0 +1,14 @@
+-module(other_module).
+
+-compile([export_all, nowarn_export_all]).
+
+%% these exported types are used in remote_types module
+-export_type([my_generic/1]).
+
+% -type other_module_type() :: other_module.
+
+%% v-- Either a local type (my_tuple) to this module or a remote user type
+-type my_generic(RetType) :: my_shadowed_type() | {ok, RetType}.
+%% A type to be shadowed by a type with the same name in a local module.
+%% To be used in `user_types:my_generic(my_shadowed_type())`
+-type my_shadowed_type() :: other_module.
--- a/test/should_pass/other_module.erl
+++ b/test/should_pass/other_module.erl
@@ -3,7 +3,7 @@
 -compile([export_all, nowarn_export_all]).
 
 %% these exported types are used in remote_types module
--export_type([my_generic/1]).
+-export_type([my_generic/1, nested_generic/1]).
 
 % -type other_module_type() :: other_module.
 
@@ -12,3 +12,5 @@
 %% A type to be shadowed by a type with the same name in a local module.
 %% To be used in `user_types:my_generic(my_shadowed_type())`
 -type my_shadowed_type() :: other_module.
+
+-type nested_generic(RetType) :: nested_generic | my_generic(RetType).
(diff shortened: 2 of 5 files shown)


zuiderkwast (approved):
Awesome!

Add a record test case if you want, but I'm OK without it too.

berbiche:
@zuiderkwast I'm not sure what kind of record test I should make. Should I test remote types in record fields or should I test generics with records or something else?

zuiderkwast:
> @zuiderkwast I'm not sure what kind of record test I should make. Should I test remote types in record fields or should I test generics with records or something else?

I haven't thought about it very much. Maybe there is no point? Records can't be parametrized with type variables anyway. AFAICT you didn't touch any record code anyway and didn't break any legacy test so I guess it's fine as it is.

ilya-klyuchnikov:
> Records can't be parametrized with type variables anyway.

They can through a level of indirection of type aliasing.

```
-record(rec, {field :: term()}).

-type param_record(A) :: #rec{ field :: A}.
```

zuiderkwast:
The fix is very simple, so I merged it now.

It would be nice to know if @ilya-klyuchnikov's example with records refined with type variable also works, but it can be done later.

PR #330, opened 2021-05-23
Non-trivial sum type exhaustiveness checking

erszcz:
I tried exercising sum type exhaustiveness checking on some practical examples and realised that these checks only worked on simple sum types whose variants were builtin types - e.g. an alternative of atoms:

```erlang
-type allergen() :: eggs
                  | chocolate
                  | pollen
                  | cats.
```

I hoped I could rely on these checks when defining functions over a type with many variants, each carrying extra data. For example:

```erlang
-type term_() :: {true, info()}
               | {false, info()}
               | {'if', info(), term_(), term_(), term_()}
               | {zero, info()}
               | {succ, info(), term_()}
               | {pred, info(), term_()}
               | {is_zero, info(), term_()}.
```

The first approach was to enable `typechecker:refinable()` and `gradualizer_lib:pick_value()` to handle `{user_type, _, _, _}` types. However, this first implementation was prone to falling into an infinite loop on recursive variants. This lead me to add the `Trace` parameter to `refinable` - in other words, I optimistically assume that a type T is refinable as soon as recursion on T's variants reaches T again. This assumes that the other non-recursive variants of T will ultimately determine its refinability. Thanks to storing a trace of already seen types instead of a binary `seen | not-seen` flag mutually recursive types are also handled gracefully.

Then it turned out that if the sum type is defined in another module, the check doesn't work. Obviously, data types like this might be handled in multiple modules of a project, so this was not acceptable. The fix I applied is too look up user types in `gradualizer_db` first and only if not found there try the available `TEnv`. [I tried the reverse order](https://github.com/erszcz/Gradualizer/tree/exhaustive-user-type-normalize-remote-type) (`TEnv` first, then `gradualizer_db`), but it seems to lead to more convoluted code since `#tenv{}` is not available in a header, therefore not shareable with `gradualizer_lib` - nothing impossible to handle, but the former fix was already good enough.

Ultimately, this allows to rely on exhaustiveness warnings when working with (mutually recursive) non-trivial variant types from external modules, which IMO is a significant ergonomics improvement. For a larger example, please see https://github.com/erszcz/tapl-erlang/commit/300d1ed100ab1c820dc2fd54d290345f3449a231 and try:

```
$ rebar3 compile
$ gradualizer -pa _build/default/lib/arith/ebin -- src/arith_core.erl
src/arith_core.erl: Nonexhaustive patterns on line 34 at column 9
	{false, 0}
```

Gradualizer correctly detects the missing clauses:

```erlang
-type term_() :: arith_syntax:term_().

-spec eval1(term_()) -> term_().
eval1(T) ->
    case T of
        %{false, _} ->
        %    erlang:throw(constant);
        %{true, _} ->
        %    erlang:throw(constant);
        %{zero, _} ->
        %    erlang:throw(constant);
        %{'if', _, {true, _}, T2, _} ->
        %    T2;
        %{'if', _, {false, _}, _, T3} ->
        %    T3;
        %{'if', Info, T1, T2, T3} ->
        %    T1_ = eval1(T1),
        %    {'if', Info, T1_, T2, T3};
        {succ, Info, T1} ->
            T1_ = eval1(T1),
            {succ, Info, T1_};
        {pred, _, {zero, _}} ->
            {zero, 0};
        {pred, _, {succ, _, T1}} ->
            T1;
        {pred, Info, T1} ->
            T1_ = eval1(T1),
            {pred, Info, T1_};
        {is_zero, _, {zero, _}} ->
            {true, 0};
        {is_zero, _, {succ, _, _}} ->
            {false, 0};
        {is_zero, Info, T1} ->
            T1_ = eval1(T1),
            {is_zero, Info, T1_}
    end.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ test/arg.beam: test/should_fail/arg.erl
 test_data_erls = $(wildcard test/known_problems/**/*.erl test/should_fail/*.erl test/should_pass/*.erl)
 build_test_data:
 	mkdir -p "test_data"
-	erlc -o test_data $(test_data_erls)
+	erlc $(ERLC_OPTS) -o test_data $(test_data_erls)
 
 EUNIT_OPTS =
 
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ erls = $(wildcard src/*.erl)
 beams = $(erls:src/%.erl=ebin/%.beam)
 
 ERLC_OPTS = -I include -pa ebin +debug_info
+TEST_ERLC_OPTS = +debug_info
 
 app: $(beams) ebin/gradualizer.app
 
@@ -132,7 +133,7 @@ test/arg.beam: test/should_fail/arg.erl
 test_data_erls = $(wildcard test/known_problems/**/*.erl test/should_fail/*.erl test/should_pass/*.erl)
 build_test_data:
 	mkdir -p "test_data"
-	erlc $(ERLC_OPTS) -o test_data $(test_data_erls)
+	erlc $(TEST_ERLC_OPTS) -o test_data $(test_data_erls)
 
 EUNIT_OPTS =
 
--- a/src/gradualizer_lib.erl
+++ b/src/gradualizer_lib.erl
@@ -2,7 +2,7 @@
 
 -module(gradualizer_lib).
 
--export([merge_with/3, top_sort/1, get_user_type_definition/4,
+-export([merge_with/3, top_sort/1, get_user_type_definition/2,
          pick_value/2, fold_ast/3, get_ast_children/1,
          empty_tenv/0, create_tenv/3]).
 -export_type([graph/1, tenv/0]).
@@ -91,7 +91,16 @@ reverse_graph(G) ->
     from_edges(maps:keys(G), [ {J, I} || {I, Js} <- maps:to_list(G), J <- Js ]).
 
 
-get_user_type_definition(Types, Anno, Name, Args) ->
+%% Look up `UserTy' definition:
+%% first in `gradualizer_db', then, if not found, in provided `Types' map.
+%% `UserTy' is actually an unexported `gradualizer_type:af_user_defined_type()'.
+
+-spec get_user_type_definition(UserTy, Types) -> Ty when
+      UserTy :: gradualizer_type:abstract_type(),
+      Types :: #{{Name :: atom(), arity()} => {Params :: [atom()],
+                                               Body :: gradualizer_type:abstract_type()}},
+      Ty :: gradualizer_type:abstract_type().
+get_user_type_definition({user_type, Anno, Name, Args}, Types) ->
     %% Let's check if the type is a known remote type.
     case typelib:get_module_from_annotation(Anno) of
         {ok, Module} ->
--- a/src/gradualizer_lib.erl
+++ b/src/gradualizer_lib.erl
@@ -95,23 +95,25 @@ reverse_graph(G) ->
 %% first in `gradualizer_db', then, if not found, in provided `Types' map.
 %% `UserTy' is actually an unexported `gradualizer_type:af_user_defined_type()'.
 
--spec get_type_definition(UserTy, Types) -> {ok, Ty} | opaque | not_found when
+-spec get_type_definition(UserTy, TEnv) -> {ok, Ty} | opaque | not_found when
       UserTy :: gradualizer_type:abstract_type(),
-      Types :: #{{Name :: atom(), arity()} => {Params :: [atom()],
-                                               Body :: gradualizer_type:abstract_type()}},
+      TEnv :: tenv(),
       Ty :: gradualizer_type:abstract_type().
-get_type_definition({remote_type, _Anno, [{atom, _, Module}, {atom, _, Name}, Args]}, _Types) ->
+get_type_definition({remote_type, _Anno, [{atom, _, Module}, {atom, _, Name}, Args]}, _TEnv) ->
     gradualizer_db:get_type(Module, Name, Args);
-get_type_definition({user_type, Anno, Name, Args}, Types) ->
+get_type_definition({user_type, Anno, Name, Args}, TEnv) ->
     %% Let's check if the type is a known remote type.
     case typelib:get_module_from_annotation(Anno) of
         {ok, Module} ->
             gradualizer_db:get_type(Module, Name, Args);
         none ->
             %% Let's check if the type is defined in the context of this module.
-            case maps:get({Name, length(Args)}, Types, not_found) of
-                {_Params, Ty} ->
-                    {ok, Ty};
+            case maps:get({Name, length(Args)}, maps:get(types, TEnv), not_found) of
+                {Params, Type0} ->
+                    VarMap = maps:from_list(lists:zip(Params, Args)),
+                    Type1 = typelib:annotate_user_types(maps:get(module, TEnv), Type0),
+                    Type2 = typelib:substitute_type_vars(Type1, VarMap),
+                    {ok, Type2};
                 not_found ->
                     not_found
             end
@@ -122,61 +124,60 @@ get_type_definition({user_type, Anno, Name, Args}, Types) ->
 %% Used in exhaustiveness checking to show an example value
 %% which is not covered by the cases.
 
--spec pick_value(Ty, Types) -> AbstractVal when
+-spec pick_value(Ty, TEnv) -> AbstractVal when
       Ty :: gradualizer_type:abstract_type(),
-      Types :: #{{Name :: atom(), arity()} => {Params :: [atom()],
-                                               Body :: gradualizer_type:abstract_type()}},
+      TEnv :: tenv(),
       AbstractVal :: gradualizer_type:abstract_expr().
-pick_value(List, Types) when is_list(List) ->
-    [pick_value(Ty, Types) || Ty <- List ];
-pick_value(?type(integer), _Types) ->
+pick_value(List, TEnv) when is_list(List) ->
+    [pick_value(Ty, TEnv) || Ty <- List ];
+pick_value(?type(integer), _TEnv) ->
     {integer, erl_anno:new(0), 0};
-pick_value(?type(char), _Types) ->
+pick_value(?type(char), _TEnv) ->
     {char, erl_anno:new(0), $a};
-pick_value(?type(non_neg_integer), _Types) ->
+pick_value(?type(non_neg_integer), _TEnv) ->
     {integer, erl_anno:new(0), 0};
-pick_value(?type(pos_integer), _Types) ->
+pick_value(?type(pos_integer), _TEnv) ->
     {integer, erl_anno:new(0), 0};
-pick_value(?type(neg_integer), _Types) ->
+pick_value(?type(neg_integer), _TEnv) ->
     {integer, erl_anno:new(0), -1};
-pick_value(?type(float), _Types) ->
+pick_value(?type(float), _TEnv) ->
     {float, erl_anno:new(0), -1};
-pick_value(?type(atom), _Types) ->
+pick_value(?type(atom), _TEnv) ->
     {atom, erl_anno:new(0), a};
-pick_value({atom, _, A}, _Types) ->
+pick_value({atom, _, A}, _TEnv) ->
     {atom, erl_anno:new(0), A};
-pick_value({ann_type, _, [_, Ty]}, Types) ->
-    pick_value(Ty, Types);
-pick_value(?type(union, [Ty|_]), Types) ->
-    pick_value(Ty, Types);
-pick_value(?type(tuple, any), _Types) ->
+pick_value({ann_type, _, [_, Ty]}, TEnv) ->
+    pick_value(Ty, TEnv);
+pick_value(?type(union, [Ty|_]), TEnv) ->
+    pick_value(Ty, TEnv);
+pick_value(?type(tuple, any), _TEnv) ->
     {tuple, erl_anno:new(0), []};
-pick_value(?type(tuple, Tys), Types) ->
-    {tuple, erl_anno:new(0), [pick_value(Ty, Types) || Ty <- Tys]};
-pick_value(?type(record, [{atom, _, RecordName}]), _Types) ->
+pick_value(?type(tuple, Tys), TEnv) ->
+    {tuple, erl_anno:new(0), [pick_value(Ty, TEnv) || Ty <- Tys]};
+pick_value(?type(record, [{atom, _, RecordName}]), _TEnv) ->
     {record, erl_anno:new(0), RecordName, []};
-pick_value(?type(record, [{atom, _, RecordName} | Tys]), Types) ->
+pick_value(?type(record, [{atom, _, RecordName} | Tys]), TEnv) ->
     MFields = [
-        {record_field, erl_anno:new(0), {atom, erl_anno:new(0), FieldName}, pick_value(Ty, Types)}
+        {record_field, erl_anno:new(0), {atom, erl_anno:new(0), FieldName}, pick_value(Ty, TEnv)}
         || ?type(field_type, [{atom, _, FieldName}, Ty]) <- Tys
     ],
     {record, erl_anno:new(0), RecordName, MFields};
-pick_value(?type(list), _Types) ->
+pick_value(?type(list), _TEnv) ->
     {nil, erl_anno:new(0)};
-pick_value(?type(list,_), _Types) ->
+pick_value(?type(list,_), _TEnv) ->
     {nil, erl_anno:new(0)};
-pick_value(?type(nil), _Types) ->
+pick_value(?type(nil), _TEnv) ->
     {nil, erl_anno:new(0)};
 %% The ?type(range) is a different case because the type range
 %% ..information is not encoded as an abstract_type()
 %% i.e. {type, Anno, range, [{integer, Anno2, Low}, {integer, Anno3, High}]}
-pick_value(?type(range, [{_TagLo, _, neg_inf}, Hi = {_TagHi, _, _Hi}]), _Types) ->
-    %% pick_value(Hi, Types);
+pick_value(?type(range, [{_TagLo, _, neg_inf}, Hi = {_TagHi, _, _Hi}]), _TEnv) ->
+    %% pick_value(Hi, TEnv);
     Hi;
-pick_value(?type(range, [Lo = {_TagLo, _, _Lo}, {_TagHi, _, _Hi}]), _Types) ->
-    %% pick_value(Lo, Types).
+pick_value(?type(range, [Lo = {_TagLo, _, _Lo}, {_TagHi, _, _Hi}]), _TEnv) ->
+    %% pick_value(Lo, TEnv).
     Lo;
-pick_value(Type, Types)
+pick_value(Type, TEnv)
   when element(1, Type) =:= remote_type; element(1, Type) =:= user_type ->
     {Kind, Anno, Name, Args} = case Type of
                                    {remote_type, Anno, [_, {atom, _, Name}, Args]} ->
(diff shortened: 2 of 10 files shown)


src/typechecker.erl
```diff
@@ -3453,27 +3454,44 @@ pick_one_refinement_each([Ty|Tys], [RefTy|RefTys]) ->
     RefHeadCombinations ++ RefTailCombinations.
 
 %% Is a type refinable to the point that we do exhaustiveness checking on it?
-refinable(?type(integer)) ->
+refinable(TEnv, Ty) ->
+    refinable(TEnv, Ty, sets:new()).
+
+refinable(_TEnv, ?type(integer), _Trace) ->
```
zuiderkwast, line 3460: Perhaps a matter of taste, and I know it's not even consistent within this module, but I think `refinable(Ty, TEnv)` is a better order of the parameters.

src/typechecker.erl
```diff
@@ -3453,27 +3454,44 @@ pick_one_refinement_each([Ty|Tys], [RefTy|RefTys]) ->
...
-refinable(?type(record, [_ | Fields])) ->
-    lists:all(fun refinable/1, [X || ?type(field_type, X) <- Fields]);
-refinable(_) ->
+refinable(TEnv, ?type(union,Tys), Trace) when is_list(Tys) ->
+    lists:all(fun (Ty) -> refinable(TEnv, Ty, Trace) end, Tys);
+refinable(TEnv, ?type(tuple, Tys), Trace) when is_list(Tys) ->
+    lists:all(fun (Ty) -> refinable(TEnv, Ty, Trace) end, Tys);
+refinable(TEnv, ?type(record, [_ | Fields]), Trace) ->
+    lists:all(fun (Ty) -> refinable(TEnv, Ty, Trace) end, [X || ?type(field_type, X) <- Fields]);
+refinable(TEnv = #tenv{}, RefinableTy = {user_type, Anno, Name, Args}, Trace) ->
```
zuiderkwast, line 3480: No need to check that TEnv is a `#tenv{}` here, is it?

src/gradualizer_lib.erl
```diff
@@ -83,58 +84,86 @@ reverse_graph(G) ->
...
+    end.
+
+
 % Given a type, pick a value of that type.
 % Used in exhaustiveness checking to show an example value
 % which is not covered by the cases.
-pick_value(List) when is_list(List) ->
-    [pick_value(Ty) || Ty <- List ];
-pick_value(?type(integer)) ->
+pick_value(Types, List) when is_list(List) ->
```
zuiderkwast, line 111: I'd prefer to reverse the order of parameters, i.e. put Types param after the Type we're picking a value from. Also add some short comment about what the params are (or a spec) please.

src/gradualizer_lib.erl
```diff
@@ -83,58 +84,86 @@ reverse_graph(G) ->
     from_edges(maps:keys(G), [ {J, I} || {I, Js} <- maps:to_list(G), J <- Js ]).
 
 
+get_user_type_definition(Types, Anno, Name, Args) ->
```
zuiderkwast, line 87: To get fewer and easier-to-remember parameters, why not keep the `user_type` tuple here? as in

```suggestion
get_user_type_definition({user_type, Anno, Name, Args}, Types) ->
```

A short comment about the function and the parameters is always nice to have. Types is a map from name to type, right? (Or a spec.)

zuiderkwast (changes requested):
Very nice feature!

I have only some suggestions on style.

I have the feeling we now have multiple ways to lookup user types. Is it worth refactoring the other accesses to get less ways of doing it?

src/gradualizer_lib.erl
```diff
@@ -83,58 +84,86 @@ reverse_graph(G) ->
...
+    %% Let's check if the type is a known remote type.
+    case typelib:get_module_from_annotation(Anno) of
+        {ok, Module} ->
+            case gradualizer_db:get_opaque_type(Module, Name, Args) of
```
zuiderkwast, line 91: I just saw you're using `get_opaque_type` here. I don't think want to do exhaustiveness-checking on opaque remote types, since they are opaque...

src/gradualizer_lib.erl
```diff
@@ -83,58 +84,86 @@ reverse_graph(G) ->
...
+                {ok, Ty} ->
+                    {ok, Ty};
+                not_found ->
+                    not_found
+            end;
+        none ->
+            %% Let's check if the type is defined in the context of this module.
+            case maps:get({Name, length(Args)}, Types, not_found) of
+                {_Params, Ty} ->
+                    {ok, Ty};
```
zuiderkwast, line 101: Don't we need to substitute Args for Params in Ty here?

src/gradualizer_lib.erl
erszcz: Indeed, this was just an easy to way to avoid the `opaque` return value from `gradualizer_db:get_type`, which can be handled easily in `refinable` (`opaque` is not refinable), but [will require an error to be thrown if it ever shows up when returned in `pick_value`](https://github.com/erszcz/Gradualizer/blob/exhaustive-user-type/src/gradualizer_lib.erl#L181-L182). I wanted to avoid introducing a new kind of error as I don't see anything relevant to `opaque` in `gradualizer_fmt`.

src/gradualizer_lib.erl
erszcz: Hmmm, I'm not sure yet, I'd have to devise a test to work it out on some examples...

erszcz:
I've pushed some changes - all the style / parameter order / spec points are addressed (unless you have more comments on these). The open issues I still have to think of are the `throw(opaque, ...)` from `pick_value` and the `Args <-> Params` substitution - any more feedback is appreciated :)

src/gradualizer_lib.erl
zuiderkwast: Good point. If we want to pick_value of type `{tag, opaque_type()}`, we'd need a value of the opaque type. But OTOH opaque values shouldn't be shown.

Since pick_value is only used for producing error messages, perhaps we can put an underscore or three dots in there instead? `{tag, _}` or `{tag, ...}`, whatever is best. We could also change the exhaustiveness error messages slightly so that it doesn't explicitly say that it's a value of the type. Just some idea. WDYT?

src/gradualizer_lib.erl
```diff
@@ -83,58 +84,105 @@ reverse_graph(G) ->
...
-    [pick_value(Ty) || Ty <- List ];
-pick_value(?type(integer)) ->
+%% Look up `UserTy' definition:
+%% first in `gradualizer_db', then, if not found, in provided `Types' map.
+%% `UserTy' is actually an unexported `gradualizer_type:af_user_defined_type()'.
+
+-spec get_user_type_definition(UserTy, Types) -> {ok, Ty} | opaque | not_found when
+      UserTy :: gradualizer_type:abstract_type(),
+      Types :: #{{Name :: atom(), arity()} => {Params :: [atom()],
+                                               Body :: gradualizer_type:abstract_type()}},
```
zuiderkwast, line 94: Nice. Since this Types map is used in many places, we *could* consider creating a type for this, like `gradualizer:type_map()` or so... But I won't demand it.

src/gradualizer_lib.erl
```diff
@@ -83,58 +84,105 @@ reverse_graph(G) ->
...
+
+%% Given a type `Ty', pick a value of that type.
+%% Used in exhaustiveness checking to show an example value
+%% which is not covered by the cases.
+
+-spec pick_value(Ty, Types) -> AbstractVal when
+      Ty :: gradualizer_type:abstract_type(),
+      Types :: #{{Name :: atom(), arity()} => {Params :: [atom()],
+                                               Body :: gradualizer_type:abstract_type()}},
+      AbstractVal :: gradualizer_type:abstract_expr().
```
zuiderkwast, line 127: Types might be called UserTypes, UserTypeMap or something (only in the spec) to avoid confusion Ty/Types.

src/gradualizer_lib.erl
```diff
@@ -83,58 +84,105 @@ reverse_graph(G) ->
...
-    Lo.
+pick_value(?type(range, [Lo = {_TagLo, _, _Lo}, {_TagHi, _, _Hi}]), _Types) ->
+    %% pick_value(Lo, Types).
+    Lo;
+pick_value({user_type, Anno, Name, Args} = UserTy, Types) ->
+    case get_user_type_definition(UserTy, Types) of
+        {ok, Ty} ->
+            pick_value(Ty, Types);
+        opaque ->
+            throw({opaque, user_type, Anno, {Name, length(Args)}});
```
zuiderkwast, line 182: Idea: return `{var, Anno, '_'}` here, for now. Or generate a unique OpaqueValueNNN variable name using unique_integer or something.

src/gradualizer_lib.erl
zuiderkwast: Yes, tests are good to have.

Just thinking that `gradualizer_db:get_type/3` does substitute type params and that local and remote types probably should be handled in the same way.

zuiderkwast: Btw, this clause in `typechecker` is very similar and can maybe be refactored to use `get_user_type_definition`:

```Erlang
normalize({user_type, P, Name, Args} = Type, TEnv) ->
    case typelib:get_module_from_annotation(P) of
        {ok, Module} ->
            %% Local type in another module, from an expanded remote type
            case gradualizer_db:get_type(Module, Name, Args) of
                {ok, T} ->
                    normalize(typelib:remove_pos(T), TEnv);
                opaque ->
                    Type;
                not_found ->
                    throw({undef, user_type, P, {Module, Name, length(Args)}})
            end;
        none ->
            %% Local user-defined type
            TypeId = {Name, length(Args)},
            case TEnv#tenv.types of
                #{TypeId := {Vars, Type0}} ->
                    VarMap = maps:from_list(lists:zip(Vars, Args)),
                    Type1 = typelib:substitute_type_vars(Type0, VarMap),
                    normalize(Type1, TEnv);
                _NotFound ->
                    throw({undef, user_type, P, {Name, length(Args)}})
            end
    end;
```

test/should_fail/exhaustive_remote_user_type.erl
```diff
@@ -24,3 +26,17 @@ remote(T) ->
     case T of
         {true, _} -> ok
     end.
+
+-type g(T) :: ok | {generic, T}.
+
+-spec generic_with_remote_opaque(g(exhaustive_user_type:opaque_t())) -> ok.
```
zuiderkwast, line 32: Really nice that you managed to test this stuff!

src/typechecker.erl
```diff
@@ -3486,15 +3491,18 @@ refinable(RefinableTy = {user_type, _Anno, _Name, _Args}, TEnv, Trace) ->
...
-                opaque -> false;
+                opaque -> true;
                 not_found -> false
             end
     end;
 refinable(_, _, _) ->
     false.
 
+tenv_to_map(#tenv{module = M, types = Tys, records = Rs}) ->
+    #{module => M, types => Tys, records => Rs}.
```
zuiderkwast, line 3504: So now we have two representations of TEnv... Not ideal but I can live with it if there is a plan to unify them in the future.  I'd prefer either of

1. Change TEnv to a map everywhere.
2. Use the record everywhere and move it to an include file.

test/test.erl
```diff
@@ -33,15 +33,17 @@ gen_should_fail() ->
         fun() ->
             %% user_types.erl is referenced by opaque_fail.erl.
             %% It is not in the sourcemap of the DB so let's import it manually
-            gradualizer_db:import_erl_files(["test/should_pass/user_types.erl"])
+            gradualizer_db:import_erl_files(["test/should_pass/user_types.erl"]),
+            %% exhaustive_user_type.erl is referenced by exhaustive_remote_user_type.erl
+            gradualizer_db:import_erl_files(["test/should_fail/exhaustive_user_type.erl"])
```
zuiderkwast, line 38: You could use `user_types:my_opaque()` in the new test cases. Then you don't need to import another module here.

src/typechecker.erl
erszcz: I'd prefer changing it to a map everywhere, unless you think it's better to keep it as a record. I can do this in this PR or a separate one.

erszcz: BTW, this map for the time being is described by https://github.com/erszcz/Gradualizer/blob/exhaustive-user-type/src/gradualizer_lib.erl#L87-L90 - where do you think is the right place to put this type?

test/test.erl
erszcz: Good idea, I'll try reworking the tests this way :+1:

src/typechecker.erl
zuiderkwast: It's internal so any other module than `gradualizer`. I suppose `gradualizer_lib` is fine.

The modules are a bit messy. Some are prefixed by `gradualizer_` and some aren't. The `typechecker` module has `export_all`. We should clean up this mess some day...

zuiderkwast: > I'd prefer changing it to a map everywhere, unless you think it's better to keep it as a record. I can do this in this PR or a separate one.

I guess a separate PR would be better, if it's not too much to ask. It can be done either before or after this PR is merged, as long as it's not forgotten.

src/gradualizer_lib.erl
erszcz: Ok, this one is refactored to use `get_type_definition`, too.

src/typechecker.erl
erszcz: Done with #333.

test/test.erl
erszcz: Thinking about it again, I think it makes sense to leave it the way it is.

Tests in `exhaustive_remote_user_type` mirror those from `exhaustive_user_type`. If an `exhaustive_user_type` test generates warnings, its counterpart in `exhaustive_remote_user_type` should generate the same or a similar warning. This wouldn't be easy to replicate if `exhaustive_remote_user_type` referred to tests in `user_types`, since the latter is a `should_pass` type test, whereas `exhaustive_user_type` is a `should_fail` type test. From yet another perspective, it makes sense to make `exhaustive_remote_user_type` and `exhaustive_user_type` work on the same types, but access them respectively as remote or local - it's just easier to reason about.

erszcz:
I think this one is ready to merge, unless you have further comments.

Maybe the last commit introducing behaviour described by `test/known_problems/should_fail/exhaustive_map_variants.erl` is debatable, since AFAICT map typing is still not completely sound, and I'm not sure if you think map variants are a reasonable goal (I personally think they are). WDYT?

zuiderkwast:
Yes, I will look through it again. I might have a few minor comment.

I think we should aim for handling map variants/unions.

Makefile
```diff
@@ -132,7 +132,7 @@ test/arg.beam: test/should_fail/arg.erl
 test_data_erls = $(wildcard test/known_problems/**/*.erl test/should_fail/*.erl test/should_pass/*.erl)
 build_test_data:
 	mkdir -p "test_data"
-	erlc -o test_data $(test_data_erls)
+	erlc $(ERLC_OPTS) -o test_data $(test_data_erls)
```
zuiderkwast, line 135: The idea of not using the same `-I include -pa ebin` for the test programs as for the Gradualizer src itself was to make them independent. I think that was a good thing.

Is it the include or the ebin you need in the tests here?

src/gradualizer_lib.erl
```diff
@@ -90,58 +91,114 @@ reverse_graph(G) ->
...
+                                   {remote_type, Anno, [_, {atom, _, Name}, Args]} ->
+                                       {remote_type, Anno, Name, Args};
+                                   {user_type, Anno, Name, Args} ->
+                                       {user_type, Anno, Name, Args}
+                               end,
+    case get_type_definition(Type, TEnv, [annotate_user_types]) of
+        {ok, Ty} ->
+            pick_value(Ty, TEnv);
+        opaque ->
+            {var, Anno, '_Opaque'};
```
zuiderkwast, line 198: If there are multiple opaque types in a picked value, using the same variable name would imply all opaque values are equal, as in `{picked_tuple, _Opaque, _Opaque}`. That's why I suggested adding a unique integer to the name or use just a `_`. Not a big deal though since this only for an error message. There is a point in not flooding the atom table as well.

zuiderkwast (commented):
I may accept it as is, if you reply to the comments. :-)

Makefile
erszcz: Actually, the `debug_info`, as I think it's necessary to refer to remote types from `test_module` as in:

```
gradualizer -pa ./test_data -- test/should_fail/<test_module>.erl
```

test/known_problems/should_fail/exhaustive_map_variants.erl
```diff
@@ -0,0 +1,16 @@
...
+-export_type([map_sum_t/0]).
+
+-type map_sum_t() :: #{field_one := _}
+                   | #{field_two := _}.
+
+-spec map_variants(map_sum_t()) -> ok.
+map_variants(T) ->
+    case T of
+        #{field_one := _} -> ok
+    end.
```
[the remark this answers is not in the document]
erszcz: I haven't really looked for the reason of this not being detected - I'm aware that typing maps in general is not 100% correct yet, so it might be because of this PR or because of the existing code. I just wanted to add map variants for completeness sake, since ordinary tuples and records are already taken care of.

test/known_problems/should_fail/exhaustive_remote_map_variants.erl
```diff
@@ -0,0 +1,15 @@
...
+%% See also test/should_fail/exhaustive_user_type.erl
+%% and test/should_fail/exhaustive_remote_user_type.erl
+
+-export([remote_map_variants/1]).
+
+-spec remote_map_variants(exhaustive_user_type:map_sum_t()) -> ok.
+remote_map_variants(T) ->
+    case T of
+        #{field_one := _} -> ok
+    end.
```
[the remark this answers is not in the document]
erszcz: I wanted to make it uniform with other examples from `exhaustive_remote_user_type` and `exhaustive_user_type` - i.e. that each local type tested for exhaustiveness is also accessed as a remote type. In the "worst" case (which is not that bad ;)) it'll just start passing as soon as the non-remote test is fixed.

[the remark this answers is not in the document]
zuiderkwast: OK, fair enough.

Makefile
zuiderkwast: Ah, so can we have a `TEST_ERLC_OPTS` instead or simply add `debug_info` on this command line?

src/gradualizer_lib.erl
erszcz: Indeed, this might be slightly misleading if more than one opaque type is used... I wanted to avoid passing extra mappings from types to names, but it shouldn't be hard to add - I'll do it in this PR.

Makefile
erszcz: Yep. Ok, I'll take care of that 👍

src/gradualizer_lib.erl
zuiderkwast: Is there a need for passing around extra mappings? I was thinking just

```Erlang
list_to_atom("Opaque" ++ integer_to_list(erlang:unique_integer([positive])))
```

Numbers might look a bit random in this way but at least they would be unique.

zuiderkwast: > passing extra mappings from types to names

The same type used twice doesn't mean it's the same value, so I think it would actually be wrong to use the same variable twice in this case.

zuiderkwast (approved):
I'm happy now.

Do you wan to summarize a commit message for a squash merge?

erszcz:
Maybe:

> Add non-trivial sum type exhaustiveness checking 
> 
> Enable exhaustiveness checking on sum types whose variants are tuples and records. Local and remote types, as well as (mutually) recursive types are supported. Map variant tests are added to known problems, but support is not present yet.

?

zuiderkwast:
Awesome!

PR #333, opened 2021-05-28
Extract type env from typechecker.erl and convert it to a map

erszcz:
As sketched out in https://github.com/josefs/Gradualizer/pull/330#discussion_r639679615 I've refactored `#tenv{}` to a map. I also moved it to a separate module, to slim down `typechecker.erl` a tiny bit.

I guess it's arguable if we want to define helpers like `gradualizer_type_env:{module,types,records}` in every case or just rely on `maps:get(Field, Map)`. The former offers more type safety at the cost of verbosity. The latter leads to less code, but is flimsier. What do you think?


src/gradualizer_type_env.erl
```diff
@@ -0,0 +1,58 @@
...
+      records => RecordMap}.
+
+-spec module(t()) -> module().
+module(#{module := Module}) -> Module.
+
+-spec types(t()) -> type_map().
+types(#{types := Types}) -> Types.
+
+-spec records(t()) -> record_map().
+records(#{records := Records}) -> Records.
```
zuiderkwast, line 58: Personally I don't like this kind of boilerplate.

If the loss of type information due to `maps:get/2` is the problem, it's possible to use pattern matching instead:

```Erlang
#{module := Module} = TEnv.
```

zuiderkwast (changes requested):
I don't think a separate module for each type is necessary. It gets rather bloated IMO. Just define the type in e.g. `gradualizer_lib`. If `gradualizer_lib:tenv()` is too long to write in every spec in typechecker, then we can just define a local alias using `-type tenv() :: gradualizer_lib:tenv().`.

It's my personal optinion anyway.

erszcz:
How about now?

src/typechecker.erl
```diff
@@ -416,13 +411,13 @@ glb(T1, T2, A, TEnv) ->
         %% actual use case.
         true -> {type(none), constraints:empty()};
         false ->
-            case gradualizer_cache:get_glb(TEnv#tenv.module, T1, T2) of
+            case gradualizer_cache:get_glb(maps:get(module, TEnv), T1, T2) of
```
zuiderkwast, line 414: If we want to avoid maps:get/2 here, we could match it out as `TEnv = #{module := Module}` in the function head... (but I'm not saying maps:get/2 is a problem).

src/gradualizer_lib.erl
```diff
@@ -170,3 +177,32 @@ get_ast_children({clauses, Clauses}) ->
...
+    TypeMap =
+        maps:from_list([begin
+                            Id       = {Name, length(Vars)},
+                            Params   = [VarName || {var, _, VarName} <- Vars],
+                            {Id, {Params, typelib:remove_pos(Body)}}
+                        end || {Name, Body, Vars} <- TypeDefs]),
+    RecordMap =
+        maps:from_list([{Name, [{typed_record_field, Field, typelib:remove_pos(Type)}
+                                || {typed_record_field, Field, Type}
+                                       <- lists:map(fun absform:normalize_record_field/1,
```
zuiderkwast, line 203: We could think about what the responsibility is of the different modules `typelib`, `absform` and `gradualizer_lib`... Maybe there is none...?

src/gradualizer_lib.erl
```diff
@@ -170,3 +177,32 @@ get_ast_children({clauses, Clauses}) ->
...
     Children.
+
+
+%% ------------------------------------------------
+%% Type environment
+%% ------------------------------------------------
+
+-spec empty_tenv() -> tenv().
+empty_tenv() ->
+    #{module => undefined,
```
zuiderkwast, line 188: The empty tenv is sort of a hack. It is (or should) probably only be used when the the module is not needed, but `undefined` is an atom that could actually be a module name... Not sure we can do anything about it though. Maybe omit module from the map, is that better?

src/gradualizer_lib.erl
```diff
@@ -170,3 +177,32 @@ get_ast_children({clauses, Clauses}) ->
...
+      types => #{},
+      records => #{}}.
+
+-spec create_tenv(_, _, _) -> tenv().
+create_tenv(Module, TypeDefs, RecordDefs) ->
```
zuiderkwast, line 193: Specs like this one can always be improved or be accompanied by some comment. I know it's just moved code so I won't request it as part of this PR.

zuiderkwast (commented):
Better!

src/typechecker.erl
erszcz: Yes, I'm aware of it. I think it's fine the way it is.

src/gradualizer_lib.erl
erszcz: Hmm, this is a direct translation from using a record, but indeed, it seems the original use of a record was buggy in this place - https://github.com/josefs/Gradualizer/blob/212db7019c91d66e38be3ae5bc8cf35f52a1b951/src/typechecker.erl#L419. In other words, if we remove this field from the empty tenv, tests don't pass with:

```
test:27: gen_should_pass (intersection_with_any.erl)...*failed*
in function erlang:map_get/2
  called as map_get(module,#{records => #{},types => #{}})
in call from typechecker:glb/4 (src/typechecker.erl, line 414)
in call from typechecker:add_any_types_pat/2 (src/typechecker.erl, line 4058)
in call from typechecker:'-add_type_pat/4-lc$^1/1-0-'/2 (src/typechecker.erl, line 3769)
in call from typechecker:add_type_pat/4 (src/typechecker.erl, line 3769)
in call from typechecker:add_types_pats/7 (src/typechecker.erl, line 3677)
in call from typechecker:check_clause/5 (src/typechecker.erl, line 3235)
in call from typechecker:'-check_clauses/5-fun-0-'/6 (src/typechecker.erl, line 3197)
**error:{badkey,module}
  output:<<"">>
```

Which means that when the `#tenv{}` record was in use, `undefined` was the module name passed to `gradualizer_cache:get_glb()`.

zuiderkwast: Right, let's leave it as is then. Perhaps leave a TODO comment if you want..

src/typechecker.erl
```diff
@@ -416,13 +411,14 @@ glb(T1, T2, A, TEnv) ->
         %% actual use case.
         true -> {type(none), constraints:empty()};
         false ->
-            case gradualizer_cache:get_glb(TEnv#tenv.module, T1, T2) of
+            Module = maps:get(module, TEnv, '$__improbable__module__name__$'),
```
zuiderkwast, line 414: Haha! Okay, it works. :-)

zuiderkwast (approved):
OK, we could merge it now if you're happy with it. Are you?

erszcz:
I am if you are :) Let's merge and I'll rebase https://github.com/josefs/Gradualizer/pull/330#discussion_r639679615 and continue with it.

src/gradualizer_lib.erl
zuiderkwast: Seems like you'd prefer a tenv module... Encapsulating an abstract datatype is a natural thing to do, but in this case there isn't much to encapsulate since the type is not opaque and we access it using pattern matching. Everything in such module would need to be exported anyway.

zuiderkwast:
I did a squash merge.

PR #334, opened 2021-05-31
Clean up hack in `glb` and `add_any_types_pat`

josefs:
In `add_any_types_pat` we called `glb` with one of the arguments
being `any()`. But that always returns `any()`! So we can simplify
that whole call away.
Furthermore, the call to `glb` used an `empty_tenv` which is going
to crash since `glb` needs the module name from the `tenv` but
`empty_tenv` doesn't set the module name.
So now, instead of hacking around the fact that `glb` might not get
a proper module name, we will simply always fail. This will help
us catch bugs like these faster in the future.

This issue was highlighted by @erszcz in #333. Thanks!

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -411,7 +411,7 @@ glb(T1, T2, A, TEnv) ->
         %% actual use case.
         true -> {type(none), constraints:empty()};
         false ->
-            Module = maps:get(module, TEnv, '$__improbable__module__name__$'),
+            Module = maps:get(module, TEnv),
             case gradualizer_cache:get_glb(Module, T1, T2) of
                 false ->
                     Ty1 = typelib:remove_pos(normalize(T1, TEnv)),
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
Great! Was this the only occurence of empty_tenv()? If yes, then the whole empty_tenv() can be removed.

@erszcz Please have a look.

erszcz:
@zuiderkwast Certainly not the only one:

```
$ rg empty_tenv
src/gradualizer_lib.erl
7:         empty_tenv/0, create_tenv/3]).
243:-spec empty_tenv() -> tenv().
244:empty_tenv() ->

src/typechecker.erl
791:    normalize({type, P, union, [negate_num_type(Ty)||Ty <- Tys]}, gradualizer_lib:empty_tenv());
937:            type(nonempty_list, [normalize(type(union, CharTypes), gradualizer_lib:empty_tenv())]);
1905:    case subtype(Ty, type(number), gradualizer_lib:empty_tenv()) of
1912:    case subtype(Ty, type(number), gradualizer_lib:empty_tenv()) of
1920:    case {subtype(Ty1, TInteger, gradualizer_lib:empty_tenv())
1921:         ,subtype(Ty2, TInteger, gradualizer_lib:empty_tenv())} of
1926:            case {subtype(Ty1, TFloat, gradualizer_lib:empty_tenv())
1927:                 ,subtype(Ty2, TFloat, gradualizer_lib:empty_tenv())} of
1932:                    case {subtype(Ty1, TNumber, gradualizer_lib:empty_tenv())
1933:                         ,subtype(Ty2, TNumber, gradualizer_lib:empty_tenv())} of
4068:            {RefinedTy, _Cs} = glb(VarTy, type(any), gradualizer_lib:empty_tenv()),
```

I've not checked how relevant this fix is for its other uses, though.

josefs:
> Great! Was this the only occurence of empty_tenv()? If yes, then the whole empty_tenv() can be removed.
> 
No, there are many occurrences of `empty_tenv()`. They all look benign to me. But it's unfortunate that the code is so fragile and that we can easily introduce a big by using `empty_tenv()` in the wrong place.

One option would be to require that `empty_tenv()` always took a module name so that the field would never be undefined. I briefly looked at doing that, but it would require some rather unpleasant plumbing of the module name in some places so I decided against it, at least for now.

I'm open to other suggestions for how to improve the code base in this regard.

josefs:
> I've not checked how relevant this fix is for its other uses, though.

The other ones are not as easy to remove as this one. This was just a special case.

PR #338, opened 2021-07-16
Fix refinable record fields

berbiche:
Previously, every record would be refined because the record
that was passed to `refinable/3` always had no fields.

Gradualizer would then crash on `pick_value` for types it cannot
pick an example value.

Now, the record is looked up in the REnv and every typed field of
the record is checked for "refinability".

Do I also need to lookup the records in gradualizer_db at that point (e.g. when checking for the refinement of a remote type)?

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -5,7 +5,7 @@
 
 %% Export all for easier testing and debugging while the project is
 %% still in an early stage
--compile([export_all]).
+-compile([export_all, nowarn_export_all]).
 
 -include("typelib.hrl").
 
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3456,8 +3456,21 @@ refinable(?type(union,Tys), TEnv, Trace) when is_list(Tys) ->
     lists:all(fun (Ty) -> refinable(Ty, TEnv, Trace) end, Tys);
 refinable(?type(tuple, Tys), TEnv, Trace) when is_list(Tys) ->
     lists:all(fun (Ty) -> refinable(Ty, TEnv, Trace) end, Tys);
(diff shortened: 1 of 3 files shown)


berbiche:
Found an issue with where `pick_value` is called with `{type, _, any, []}`.

I will fix this in a following commit.

src/typechecker.erl
```diff
@@ -3456,8 +3456,21 @@ refinable(?type(union,Tys), TEnv, Trace) when is_list(Tys) ->
     lists:all(fun (Ty) -> refinable(Ty, TEnv, Trace) end, Tys);
 refinable(?type(tuple, Tys), TEnv, Trace) when is_list(Tys) ->
     lists:all(fun (Ty) -> refinable(Ty, TEnv, Trace) end, Tys);
-refinable(?type(record, [_ | Fields]), TEnv, Trace) ->
-    lists:all(fun (Ty) -> refinable(Ty, TEnv, Trace) end, [X || ?type(field_type, X) <- Fields]);
+refinable(?type(record, [{atom, _, RecordName}]), TEnv = #{records := REnv}, Trace) ->
```
zuiderkwast, line 3459: Why not allow any refined Fields here, e.g. types like `#rec{field :: a}`?

berbiche: From my testing (with Gradualizer's unit tests and a significant code base), `refinable` is never called with the full type for a record i.e. the type does not have any parameters beyond the record name.
`begin ?type(record, [_ | X]), length(X) =:= 0 end` is always true.

zuiderkwast: That seems weird. I would guess it should call `refinable` with any type, but I haven't looked at the details.

PR #340, opened 2021-08-09
Some map typing fixes

erszcz:
This carries a number of fixes to the way Gradualizer typechecks maps, especially map subtyping, and therefore moves a number of tests from `known_problems` to their respective places. Trivial map type inference is also put in place. The subtyping implementation changes follow [the wiki](https://github.com/josefs/Gradualizer/wiki/Map-types#subtyping) as far as I could make them.

The code could use some cleanup, so please don't merge yet.

One thing I'm specifically unsure what to do with is the third clause below (also marked in the source code):

```erlang
glb_ty(Ty1 = {type, _, map, Assocs1}, Ty2 = {type, _, map, Assocs2}, _A, _TEnv) ->
    case {Assocs1, Assocs2} of
        {any, _} -> ret(Ty2);
        {_, any} -> ret(Ty1);
        _ ->
            ret(type(map, Assocs1 ++ Assocs2))
    end;
```

On one hand, I think that for as long as everywhere else where Gradualizer uses these assocs it processes them order-independently (which seems to be the case), we're fine with sorting them whatever way we like or leaving as above.

On the other hand, the order of params (`Ty1` and `Ty2`) should not impact the value of the result. Alas, with this simplistic treatment it does, and it shows in the following test:

```erlang
      ?glb( ?t(#{ a := integer() }), ?t(#{ b := float() }),
            ?t(#{ a := integer(), b := float() }) ),
```

This is taken care of by a hack in `deep_normalize` which sorts the assocs for the sake of comparisons in tests, but I'm afraid it's not safe to just sort that way in `typechecker.erl` in the general case.

Anyway, this PR fixes some `known_problems` and reduces the number of reported false positive and false negative diagnoses, so I'd say it's a move in the right direction.


erszcz:
BTW, I'm aware of https://github.com/josefs/Gradualizer/pull/286 and I hope I managed to take into account it's major conclusions (by following @zuiderkwast's wiki doc on maps).

zuiderkwast:
Awesome!

I haven't looked at the code yet, but the 3rd clause in the code from the PR description

```Erlang
glb_ty(Ty1 = {type, _, map, Assocs1}, Ty2 = {type, _, map, Assocs2}, _A, _TEnv) ->
    case {Assocs1, Assocs2} of
        {any, _} -> ret(Ty2);
        {_, any} -> ret(Ty1);
        _ ->
            ret(type(map, Assocs1 ++ Assocs2))
    end;
```

is just wrong. It's not according to how subtyping is described in the wiki anyway. For example `GLB(#{a := integer()}, #{b := float()})` should be `none()`, since the two types have no values in common.

berbiche:
Cool!

Does this also fix the issue with map patterns in functions with no type specs?
i.e.
```
my_func(_D = #{
    a := _A,
    b := _B,
    c := C
}) ->
  C.

%% The pattern #{a := _A, b := _B, c := C} on line X at column Y doesn't have the type any()
```

erszcz:
@zuiderkwast Good points, thanks! I was like "let's do this and see how far we can get with it" when changing the `glb` code 😅

@berbiche I think I've not seen these examples before - are they in the repo under `test/known_problems` or are they new ones? To be honest, I'm not sure when and how pattern types are inferred 🤔 I guess it's the same code as for regular map expressions, but I'm not 100% sure without checking.

zuiderkwast:
> I'm not sure when and how pattern types are inferred thinking I guess it's the same code as for regular map expressions, but I'm not 100% sure without checking.

Patterns are handled separately from expressions. They're checked by the `type_check_pat` and `type_check_pat_ty` functions if I remember correctly. Perhaps we should try to deal with patterns in a separate PR to limit the scope of each PR?

It's a good idea to add know problems for your examples @berbiche.

erszcz:
Ok, with the latest changes I think we're getting somewhere. I'm pretty sure `glb_ty` is still not correct, but at least somewhat more reasonable. For example, it doesn't try to be smart at all with regard to keys - it just uses them as they are.

I'd be really grateful for a review, especially on the test side, and feedback if this is any good or should rather be dropped.

src/typechecker.erl
```diff
@@ -567,20 +568,37 @@ glb_ty({type, _, record, _}, {type, _, record, _}, _A, _TEnv) ->
...
+            MAssocs2 = maps:from_list([ {Key, As} || ?type(_, [Key, _]) = As <- Assocs2 ]),
+            %% TODO: repeated, extract to external fun
+            IsMandatory = fun
+                              (?type(map_field_exact, _)) -> true;
+                              (_) -> false
+                          end,
+            MandatoryKeys0 = [ Key || ?type(_, [Key, _]) <- lists:filter(IsMandatory, Assocs1 ++ Assocs2) ],
+            MandatoryKeys = lists:usort(MandatoryKeys0),
+            {Assocs, Css} = lists:unzip(lists:map(fun (Key) -> glb_ty_map_step(Key, MAssocs1, MAssocs2, A, TEnv) end,
+                                                  MandatoryKeys)),
```
zuiderkwast, line 588: Please add some comments to explain what's going on here. I'm sure you know exactly what you're doing right now, but look at it again next year and you will be as lost as I am right now. :-)

src/typechecker.erl
```diff
@@ -639,6 +657,35 @@ glb_ty({type, _, Name, Args1}, {type, _, Name, Args2}, A, TEnv)
 %% Incompatible
 glb_ty(_Ty1, _Ty2, _A, _TEnv) -> {type(none), constraints:empty()}.
 
+glb_ty_map_step(Key, MAssocs1, MAssocs2, A, TEnv) ->
```
zuiderkwast, line 657: Add some comment about what this function is doing too, please.

test/typechecker_tests.erl
```diff
@@ -211,11 +211,11 @@ glb_test_() ->
...
      %% Maps
      ?glb( ?t(map()), ?t(#{a := integer()}), ?t(#{a := integer()}) ),
-     %?debugVal( deep_normalize(element(1, glb(?t(#{ a := integer() }), ?t(#{ b := float() })))) , 1000),
-     %?debugVal( deep_normalize(element(1, glb(?t(#{ b := float() }), ?t(#{ a := integer() })))) , 1000),
-     %?debugVal( deep_normalize( ?t(#{ a := integer(), b := float() }) ) , 1000),
-     ?glb( ?t(#{ a := integer() }), ?t(#{ b := float() }),
+     ?glb( ?t(#{ a := integer() }), ?t(#{ b := float() }), ?t(none()) ),
+     ?glb( ?t(#{ a := 1 }), ?t(#{ a := integer() }), ?t(#{ a := 1 }) ),
+     ?glb( ?t(#{ a := integer(), _ => _ }), ?t(#{ b := float(), _ => _ }),
            ?t(#{ a := integer(), b := float() }) ),
```
zuiderkwast: Should the result contain `_=>_` here, i.e. should it be `#{ a := integer(), b := float(), _ => _ }`?

erszcz: Indeed, it seems so when I think about it again. Marking as TODO.

src/typechecker.erl
```diff
@@ -570,25 +570,38 @@ glb_ty({type, _, record, _}, {type, _, record, _}, _A, _TEnv) ->
...
             MAssocs2 = maps:from_list([ {Key, As} || ?type(_, [Key, _]) = As <- Assocs2 ]),
+
             %% TODO: repeated, extract to external fun
             IsMandatory = fun
                               (?type(map_field_exact, _)) -> true;
                               (_) -> false
                           end,
+
+            %% Mandatory associations (`map_field_exact') are key here because they define what is
+            %% required, but also what is forbidden in a map.
```
zuiderkwast, line 591: Mandatory means required yes, but forbidden is every key which is not covered by any assoc (mandatory or not).

Good to have comments btw. :-)

erszcz: Yeah, what I wanted to highlight is that if a map has no `_ => _`, then mandatory assocs do not only define what has to be present in the map, but also what cannot go there. Maybe the wording is not 100% accurate.

From another perspective, if we go over optional assocs, it's not a big deal if we don't find a corresponding assoc in the other map. If we don't find one for mandatory assocs, we should return `type(none)` as the GLB. Please note, though, that we go over the list of mandatory assocs, but we then check if they're present in either mandatory or optional assocs in the other map.

zuiderkwast: Let's say `#{integer() => integer()}`. It has no mandatory assoc, yet non-integers are forbidden.

    GLB(#{integer() => integer()}, #{1..5 => 1..5, foo => bar}) = #{1..5 => 1..5}

erszcz: Good example!

With the latest changes it's solved properly.
We're still prone to the order of associations in map types, so I brought back the sorting hack for comparisons. This problem manifested in

```erlang
         ?glb( ?t(#{ a := integer(), _ => _ }), ?t(#{ b := float(), _ => _ }),
               ?t(#{ a := integer(), b := float(), _ => _ }) )
```

where the generated map, depending on attribute order, might be `#{ a := integer(), b := float(), _ => _ }` or `#{ b := float(), a := integer(), _ => _ }`.

I'm thinking that to solve it properly maps' associations on generation or normalisation ought to be sorted according to the order of the type hierarchy: first on keys, then values, then if still ambiguous, using the default Erlang total ordering. I think this would make sure map types are always represented uniformly. WDYT?

src/gradualizer_lib.erl
```diff
@@ -236,6 +237,34 @@ get_ast_children(Node) ->
...
+%% but `Ty2' is not `Ty1''s subtype.
+%% To ensure total ordering, if both types are subtypes of each other or they're not subtypes at all
+%% the total order over Erlang terms is used for comparison.
+-spec subtype_less_or_eq(typechecker:type(), typechecker:type(), tenv()) -> boolean().
+subtype_less_or_eq(Ty1, Ty2, TEnv) ->
+    case {typechecker:subtype(Ty1, Ty2, TEnv), typechecker:subtype(Ty2, Ty1, TEnv)} of
+        {{true, _}, {true, _}} ->
+            %% Cannot determine order in the type hierarchy.
+            %% Compare using Erlang term total order.
+            Ty1 =< Ty2;
```
zuiderkwast, line 258: I'm worried about reordering the keys in a map in this way.

If we ignore the unknown (or dynamic) type, it's probably fine to do this,
because if two types are subtypes of each other, it means they're equal.
That's not the case with gradual types though. We have `1..10 :: any()` but also
`any() :: 1..10`.

We don't know at compile time what this particular instance of `any()`
represents. It may be `1..5` or `integer()` or anything else; the programmer
might know but the type checker doesn't.

So, for example, A = `#{1..10 => atom(), any() => pid()` is not equivalent to B
= `#{any() => pid(), 1..10 => atom()`. The expression `#{5 => self()}` is of
type B but it's not of type A.

If we want to reorder the keys, we first have to make them disjoint.

I've seen in a paper, don't remember where but perhaps in
https://www.irif.fr/~gc/papers/icfp17.pdf (Gradual Typing with Union and
Intersection Types, Castagna & Lanvin), a method to do this. It involves
intersection and negation types. By example, it goes like this:

1. `#{1..10 => atom(), any() => pid()` -- Start here and let's reorder the keys
2. `#{1..10 => atom(), any() \ 1..10 => pid()` -- This is equivalent to the
above type, since the leftmost key takes precedence. The type `any() \ 1..10`
means "unknown, but at least we know it can't be anything in the range 1..10`.
3. #{any() \ 1..10 => pid(), 1..10 => atom()` -- Now the keys are disjoint, so they can be safely reordered.

The "set minus" is short-hand for "and not", so `any() & ~1..10` with another
notation.

To be able to do things properly like GLB for map types and to handle function
specs with multiple clauses (to make them disjoint in the same way), I think we
need union, intersection and negation types, but so far we have only
intersection types.

erszcz: Thanks for the examples and pointers. I've started reading the referenced paper, but following further ones it refers to seems like a rabbit hole - it's going to take a while before I grok them fully.

> I think we need union, intersection and negation types, but so far we have only intersection types.

I think we only have union types, don't we? Having started reading the above paper it seems to me that intersection and negation types would be those enabling us to express `any() \ 1..10`.

That being said, I'd like to try this approach, but given I can only make small steps in that direction, I'm afraid it's going to take a while :|

---

However, to salvage as much as possible from this PR:
- I've completely removed sorting map field associations when normalising map type or computing the GLB
- I've short-circuited `glb()` on maps to return `none()` if keys of any of the map params overlap - in other words, we return a GLB in more situations than before, but still decline to compute it if negation types are required to do it properly

This means the meaningful changes this PR carries are:
- map type inference of map creation expressions
- `subtype` (or actually `compat_ty`) logic changes which follow the wiki
- handling of map unions in `update_map_type`
- map normalisation, so they're easier to handle

To sum up, https://github.com/erszcz/Gradualizer/tree/map-tests is a fork of the Gradualizer master branch with tests from this PR added on top - it fails on 15 test cases. This PR passes all of these 15 cases.

If you think it's reasonable to introduce such an incremental improvement, instead of a complete solution with intersection types, I'll clean up this PR and prepare it for review. In such a case I think it's best to start the review from the tests side - https://github.com/erszcz/Gradualizer/tree/map-tests - that is by checking if changes I made there are valid.

Otherwise, I think we can close this PR while I (or someone else) work on a proper implementation with intersection types.

zuiderkwast: Good summary @erszcz. I think this is great work and want to see it merged!

And yes of course I meant union types. Sorry for that, if it confused anyone.

@josefs has stated before that he thinks the type representation in that paper is unnecessarily complex and that our representation of types using the erl_parse's abstract format is good enough.

I've considered rewriting the whole thing to using that representation from the paper, but that would be a lot of work, as you say. That paper doesn't solve all our problems anyway, for example polymorphism with type variables isn't solved AFAIK.

Another rewrite we could do first is to merge `type_check_expr` with `type_check_expr_in` into one that both takes a type to check against and returns a type, instead of one function returning a type and the other checking against a given type and not returning a type. Similarly for `type_check_pat(_in)` and others. That may simplify things a bit and get rid of code duplication.

test/typechecker_tests.erl
```diff
@@ -96,10 +96,16 @@ subtype_test_() ->
...
-     ?_assert(subtype(?t( #{a => b}         ), ?t( #{}              ))),
+     ?_assert(subtype(?t( #{}               ), ?t( #{a => b}        ))),
      ?_assert(subtype(?t( #{a := b}         ), ?t( #{a => b}        ))),
-     ?_assert(subtype(?t( #{a => b, c => d} ), ?t( #{a => b}        ))),
+     ?_assert(subtype(?t( #{a => b}         ), ?t( #{a => b, c => d}))),
+     %% TODO: should the below actually hold?
+     %% #{a := b} "forbids" any key but 'a',
+     %% whereas #{a := b, c := d} allows another key 'c'.
+     %% The latter does not really seem like a subtype of the former.
+     %?_assert(subtype(?t( #{a := b, c := d} ), ?t( #{a := b}        ))),
```
erszcz, line 106: @zuiderkwast I don't think this should hold, but could you also have a look? It's very similar to the original test which was here on line 101, but I think it actually was buggy.

src/gradualizer_lib.erl
erszcz: Whew, it's taken a while, but I've finally cleaned the code up. The commit history is still quite messy, but I guess it's not a problem if we consider a squash merge - please let me know if we want to keep it, though, I'll clean it up in such a case.

test/typechecker_tests.erl
zuiderkwast: I agree with the comments. It's not a subtype.

zuiderkwast: If you want to fix the test, add _ => _ to the supertype. Then it holds. If this is already covered, just delete it.

src/gradualizer_lib.erl
zuiderkwast: Nice! I think hou can squash the commits and update the commit message to a summary of what was done finally.

I will look at the code tomorrow (or the next days).

erszcz:
Closing and reopening to update CI status.

zuiderkwast:
@erszcz Is this ready to merge?

erszcz:
It is :+1:

PR #341, opened 2021-09-18
Improve self gradualization

erszcz:
This fixes the low-hanging fruit in self-gradualization. If applied together with #340 all map related errors are gone, too. This is done on OTP 24, I've not tested on other versions.

The remaining errors present even with this PR applied are due to:
- missing Rebar .beam files (typechecking the Rebar plugin)
- map typing issues (fixed if this one is used together with #340)
- actual limitations in Gradualizer, as far as I can tell mostly type refinement in guards

I think that with improved refinement some of the type assertions I added could also go away, but they're necessary as of now.

### DIFF
--- a/src/gradualizer_bin.erl
+++ b/src/gradualizer_bin.erl
@@ -118,5 +118,5 @@ gcd1(A, 0) -> A;
 gcd1(A, B) ->
   case A rem B of
     0 -> B;
-    X -> gcd1(B, X)
+    X when X > 0 -> gcd1(B, X)
   end.
--- a/src/gradualizer_cli.erl
+++ b/src/gradualizer_cli.erl
@@ -92,7 +92,7 @@ print_usage() ->
     io:format("       --fancy                   - Use fancy error messages when possible (on by default)~n"),
     io:format("       --no_fancy                - Don't use fancy error messages.~n"),
     io:format("       --union_size_limit        - Performance hack: Unions larger than this value~n"),
-    io:format("                                   are replaced by any() in normalization.").
+    io:format("                                   are replaced by any() in normalization (default: 30)~n").
 
 -spec parse_opts([string()], gradualizer:options()) -> {[string()], gradualizer:options()}.
 parse_opts([], Opts) ->
--- a/src/gradualizer_cli.erl
+++ b/src/gradualizer_cli.erl
@@ -90,7 +90,9 @@ print_usage() ->
     io:format("                                   detection of a TTY doesn't work when running as an escript.~n"),
     io:format("       --no_color                - Alias for `--color never'~n"),
     io:format("       --fancy                   - Use fancy error messages when possible (on by default)~n"),
-    io:format("       --no_fancy                - Don't use fancy error messages.~n").
+    io:format("       --no_fancy                - Don't use fancy error messages.~n"),
+    io:format("       --union_size_limit        - Performance hack: Unions larger than this value~n"),
+    io:format("                                   are replaced by any() in normalization.").
 
 -spec parse_opts([string()], gradualizer:options()) -> {[string()], gradualizer:options()}.
 parse_opts([], Opts) ->
@@ -117,6 +119,7 @@ parse_opts([A | Args], Opts) ->
         "--no_color"               -> parse_opts(Args, [{color, never} | Opts]);
         "--fancy"                  -> parse_opts(Args, [fancy | Opts]);
         "--no_fancy"               -> parse_opts(Args, [{fancy, false} | Opts]);
+        "--union_size_limit"       -> handle_union_size_limit(A, Args, Opts);
         "--"                       -> {Args, Opts};
         "-" ++ _                   -> erlang:error(string:join(["Unknown parameter:", A], " "));
         _                          -> {[A | Args], Opts}
@@ -163,6 +166,12 @@ handle_color(["always"|Args], Opts) -> parse_opts(Args, [{color, always} | Opts]
 handle_color(["never" |Args], Opts) -> parse_opts(Args, [{color, never}  | Opts]);
 handle_color(["auto"  |Args], Opts) -> parse_opts(Args, [{color, auto}   | Opts]);
 handle_color(Args,            Opts) -> parse_opts(Args, [{color, always} | Opts]).
+
+handle_union_size_limit(_, [LimitS | Args], Opts) ->
+    Limit = list_to_integer(LimitS),
+    parse_opts(Args, [{union_size_limit, Limit}]);
+handle_union_size_limit(A, [], _Opts) ->
+    erlang:error(string:join(["Missing argument for", A], " ")).
 
 no_start_dash("-" ++ _) ->
     false;
--- a/src/gradualizer_highlight.erl
+++ b/src/gradualizer_highlight.erl
@@ -177,7 +177,7 @@ highlight_line(Line, N, {L1, _C1}, {L2, C2}, Color) when L1 < N, N == L2 ->
     color_and_mark_line(Line, step_spaces(Line, 1), C2, Color);
 highlight_line(Line, N, {L1, _}, {L2, __}, Color) when L1 < N, N < L2 ->
     %% Internal line
-    color_and_mark_line(Line, step_spaces(Line, 1), length(Line) + 1, Color).
+    color_and_mark_line(Line, step_spaces(Line, 1), ?assert_type(length(Line) + 1, pos_integer()), Color).
 
 -spec step_spaces(string(), erl_anno:column()) -> erl_anno:column().
 step_spaces([$\t | Str], Col) -> step_spaces(Str, Col + 1);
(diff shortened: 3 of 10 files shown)


erszcz:
Funny, because previously this worked fine on my machine, but now after a rebase on top of the current master, I'm hitting some infinite loop in normalize 🤔  Well, apparently this needs more work :)

zuiderkwast:
Great that you're working on it. If you find it useful to have tracing functionality, we can add it and enable it by some CLI option.

erszcz:
Nah, there's no need for that, it's temporary. I'm toggling this on and off as I need, it wouldn't be feasible to expose all of `dbg` through a CLI interface. I just left it on the branch to show that we're spinning, and spinning, ... on `make gradualize`.

zuiderkwast:
You changed union size limit from 50 to 20 in some place. Can it be that?

erszcz:
> You changed union size limit from 50 to 20 in some place

I changed the other way round, since exhaustiveness checking gets disabled for unions above the limit and I ran into types with more than 20 variants in real life. Maybe it's related, I'll check.

erszcz:
@zuiderkwast That was a good lead! I timed self-gradualization with different values of `union_size_limit` and here are the results (1 run for each value, no averaging):

```
20  real	0m1.295s
30  real	0m1.629s
35  real	1m2.686s -- I terminated the program after this time, approx. 1 minute
32  real	0m1.559s
34  real	0m1.919s
```

It's especially interesting that the change from 34 to 35 is not a gradual one, but a sudden jump. I don't know the reason, but it seems suspicious. I checked if there's any type with >34 variants defined in Gradualizer, but no, the biggest one seems to be `gradualizer_type:abstract_expr()` with 29 variants 🤔 

In any case, I think it's a good reason to make this limit a configurable option instead of a hardcoded value - otherwise, it might not work well on different machines. When it's configurable, the user might just tweak it if need be.

zuiderkwast:
You can make it configurable, but remember that it's a hack that we'd prefer we wouldn't need, i.e. we'd prefer to solve it without compromising exactness.

erszcz:
> it's a hack that we'd prefer we wouldn't need

Of course :+1: If we can get rid of it, the option can go away, too. For now, I just made it possible to override the threshold without requiring a rebuild.

src/gradualizer_cli.erl
```diff
@@ -90,7 +90,9 @@ print_usage() ->
     io:format("                                   detection of a TTY doesn't work when running as an escript.~n"),
     io:format("       --no_color                - Alias for `--color never'~n"),
     io:format("       --fancy                   - Use fancy error messages when possible (on by default)~n"),
-    io:format("       --no_fancy                - Don't use fancy error messages.~n").
+    io:format("       --no_fancy                - Don't use fancy error messages.~n"),
+    io:format("       --union_size_limit        - Performance hack: Unions larger than this value~n"),
+    io:format("                                   are replaced by any() in normalization.").
```
zuiderkwast, line 95: You could mention what the default value is, like for the other options.

src/typechecker.erl
```diff
@@ -709,11 +709,12 @@ normalize({type, _, record, [{atom, _, Name}|Fields]}, TEnv) when length(Fields)
         || ?type_field_type(FieldName, Type) <- Fields],
     type_record(Name, NormFields);
 normalize({type, _, union, Tys}, TEnv) ->
+    UnionSizeLimit = ?union_size_limit,
```
zuiderkwast, line 712: Ah, now I see why we can't put it in Env: We only have TEnv here.

I think we should pass around the complete Env in many of these places where we only have Env, or use global options (persistent term or so) for all of them. I'm fine with this inconsistency for now, but it's better if we can have a more uniform way of doing options.

src/gradualizer_highlight.erl
```diff
@@ -171,13 +171,13 @@ highlight_line(Line, N, {N, C1}, {N, C2}, Color) ->
     color_and_mark_line(Line, C1, C2, Color);
 highlight_line(Line, N, {L1, C1}, {L2, _C2}, Color) when L1 == N, N < L2 ->
     %% First line
-    color_and_mark_line(Line, C1, length(Line) + 1, Color);
+    color_and_mark_line(Line, C1, ?assert_type(length(Line) + 1, pos_integer()), Color);
```
zuiderkwast, line 174: In the future, I think we could be able to handle the special case `non_neg_integer() + 1` = `pos_integer()`. It seems to appear in many places.

src/typechecker.erl
```diff
@@ -1050,7 +1051,7 @@ expect_list_union([], AccTy, AccCs, _NoAny, _N, _TEnv) ->
 infer_literal_string("") ->
     type(nil);
 infer_literal_string(Str) ->
-    SortedChars = lists:usort(Str),
+    SortedChars = ?assert_type(lists:usort(Str), [A, ...]),
```
zuiderkwast, line 1054: Why type variable `A`? We could use `char()` or just `[any(), ...]` or `nonempty_list()` if we don't care.

```suggestion
    SortedChars = ?assert_type(lists:usort(Str), [char(), ...]),
```

Another idea hit me: If we know by type refinement that it's a non-empty list here, maybe could override the type of lists:usort in the prelude with a spec that preserves the non-emptiness:

```Erlang
-spec usort([A, ...]) -> [A, ...];
           ([])       -> [].

File: src/gradualizer_bin.erl [hunk 5]
```diff
@@ -118,5 +118,5 @@ gcd1(A, 0) -> A;
 gcd1(A, B) ->
   case A rem B of
     0 -> B;
-    X -> gcd1(B, X)
+    X when X > 0 -> gcd1(B, X)
```
zuiderkwast, line 121: The guard looks unnecessary. I know why it's there of course, but it's not ideal that we impose this coding style for users who want to avoid type errors.

Maybe we could override the type spec of `rem` in the prelude so it is closed under `non_neg_integer()`, e.g.

```Erlang
-spec 'rem'(non_neg_integer(), non_neg_integer()) -> non_neg_integer();
           (integer(), integer()) -> integer().
```

WDYT? Actually, I'm not even sure that we use specs for binary operators. Maybe it's handled in some other way.

zuiderkwast (commented):
Looks good! I have some minor comment, but most of them are just ideas for future improvements; no need to fix everything.

src/gradualizer_highlight.erl
erszcz: Indeed, it pops up surprisingly frequently, so it would be nice to have it taken care of without extra assertions.

src/typechecker.erl
erszcz: Hmm, that's clever, but wouldn't it only make sense with intersection types?
Another thing is that this trick would have to be used with a number of list functions that return an empty list for an empty input list. Still, it only has to be done once for the standard library functions for a long term benefit, so good idea.

erszcz: > wouldn't it only make sense with intersection types?
Hmm, maybe not:

```
> typechecker:collect_specs_types_opaques_and_functions([merl:quote("-spec f(a) -> ok; (b) -> nok.")]).
#parsedata{module = undefined,export_all = false,
           exports = [],imports = [],
           specs = [{{f,1},
                     [{type,1,'fun',[{type,1,product,[{atom,1,a}]},{atom,1,ok}]},
                      {type,1,'fun',
                            [{type,1,product,[{atom,1,b}]},{atom,1,nok}]}]}],
           types = [],opaques = [],records = [],functions = []}
```

I was thinking that at some point these return values get merged into a union, but having skimmed over the code now it doesn't seem to be the case. My bad.

src/gradualizer_bin.erl
erszcz: This seems quite useful and flexible, but I looked for it a bit and it seems that for operators, the "specs" are encoded directly in code as [`arith_op_arg_types`](https://github.com/josefs/Gradualizer/blob/bd92da1e47a0fdb8bd22009da10bd73f0422e066/src/typechecker.erl#L2609).

zuiderkwast (approved):
Feel free to add TODOs for the things we discussed, or open issues or PRs. :-)

src/gradualizer_highlight.erl
zuiderkwast: Yes. We already have something for `pos_integer() - 1` <=> `non_neg_integer()` I think, but I'm not sure it works in both directions, i.e. (1) types pushed from result to operands and (2) result type inferred from the operands.

src/gradualizer_bin.erl
zuiderkwast: That's true. Forget about this for now then.

erszcz:
I've created issues for the open threads here. I'll go over my other recent PRs and do the same, so that the points brought up there don't get lost. In the meantime, I think this one is OK to merge - unless you think there are some blockers.

PR #342, opened 2021-09-22
Enable exhaustiveness checking over map variants

erszcz:
This builds on top of #340.

Exhaustiveness checking now works over sum types' map variants. The map types cannot contain non-refinable types themselves - some types are added to the refinable list to increase usefulness of this feature.


src/gradualizer_lib.erl
```diff
@@ -168,6 +168,12 @@ pick_value(?type(record, [{atom, _, RecordName} | Tys]), TEnv) ->
         || ?type(field_type, [{atom, _, FieldName}, Ty]) <- Tys
     ],
     {record, erl_anno:new(0), RecordName, MFields};
+pick_value(?type(map, Assocs), TEnv) ->
+    NewAssocs = [ {AssocTag, Anno, pick_value(KTy, TEnv), pick_value(VTy, TEnv)}
+                  || {type, Anno, AssocTag, [KTy, VTy]} <- Assocs ],
+    {map, erl_anno:new(0), NewAssocs};
+pick_value(?type(boolean), _TEnv) ->
+    {atom, erl_anno:new(0), false};
```
zuiderkwast, line 176: Since boolean is just `false | true`, I guess in the future we should expand this to a union and then be able to pick a value from a union?

src/typechecker.erl
```diff
@@ -3410,6 +3546,46 @@ refine_ty(Ty1, Ty2, TEnv) ->
         _NotDisjoint -> throw(no_refinement) %% imprecision
     end.
 
+-spec refine_map_field_ty({_, _}) -> gradualizer_type:abstract_type().
+%% For the same key K in both M1 and M2 the diff over its field is:
+%% M1 \ M2 |  :=  |  =>
+%% --------+------+------
+%%    :=   | none |  =>
+%%    =>   |  #{} | none
```
zuiderkwast, line 3554: Nice table, but I don't understand exactly what you mean. Is the below equivalent?

```
%% For the same key K in both M1 and M2 the diff over its field (ignoring other keys) is:
%%
%% M1 \ M2    | #{K := V2} | #{K => V2}
%% -----------+------------+---------------------
%% #{K := V1} | none       | #{K => diff(V1, V2)}
%% #{K => V1} | #{}        | none
```

src/gradualizer_lib.erl
erszcz: Hmm, I'm not sure of the extent of it, but some types are passed around denormalised. For example, see https://github.com/josefs/Gradualizer/blob/0b8e6a8fe1100b8e11cd6f822c9093ee76959eaa/src/typechecker.erl#L379 - I think some denormalised types reaching `pick_value` might come from here.

src/typechecker.erl
erszcz: Yes, with the difference that currently the implementation of the top right is:

```erlang
refine_map_field_ty({?type(map_field_exact, KVTy), ?type(map_field_assoc, KVTy)}) ->
    %% M1 = #{x := y}
    %% M2 = #{x => y}
    %% M1 \ M2 = #{x => y}
    [type(map_field_assoc, KVTy)];
 ```
 
So it's just a change of the field "exactness", not of the value type. Should it affect the value type? I don't think so, since `y \ y = none()`, which would give a `x => none()` field, and that doesn't make much sense for variant elimination from a type that's checked for exhaustiveness.

Comment (erszcz) [reply src/typechecker.erl:3554]: So to fix the table:

```
%% For the same key K in both M1 and M2 the diff over its field (ignoring other keys) is:
%%
%% M1 \ M2    | #{K := V} | #{K => V}
%% -----------+-----------+--------------
%% #{K := V}  | none      | #{K => V}
%% #{K => V}  | #{}       | none
```

erszcz: In yet other words, the maps M1 and M2 that we subtract are different because of the set of all fields in each one, but the fields that overlap are required to match with regard to both key and value types.

zuiderkwast: Ok, so you'll update these comments in any way you like? Then I guess we can merge this, unless there's anything else left to do, is there?

erszcz: Updated :+1: I think this one is ready to merge.

PR #343, opened 2021-09-25
Add a GitHub Actions CI workflow

erszcz:
I've noticed that PRs stopped being checked by Travis and realised it might be time to switch to GH Actions. To my surprise, though, the tests didn't pass from the get go in the GH Actions worker environment.

It seems that prelude was not loaded in a deterministic fashion, since on my machine the tests passed just fine, whereas they didn't "here". I imagine there must've been some race condition or a `gradualizer_db` restart in tests happening, since some tests failed as though prelude had not been loaded at all. This made me check when it's loaded and, indeed, it was not tied to `gradualizer_db` lifetime, but to an explicit call from `typechecker:type_check_forms()`. This meant that if `gradualizer_db` died and was restarted by the supervisor after the `import_prelude` call in `typechecker:type_check_forms()`, but before the typechecking finished, inconsistent information would be brought in from the DB, and some of it would use prelude, but some wouldn't.

Loading prelude from `graudalizer_db:init()` solved the problem and also uncovered that two other tests depended on the module as they started failing.

https://github.com/erszcz/Gradualizer/pull/1 contains the whole history. I'm fine with a squash merge.

### DIFF
--- a/src/gradualizer_app.erl
+++ b/src/gradualizer_app.erl
@@ -15,34 +15,6 @@
 %%%===================================================================
 
 start(_StartType, _StartArgs) ->
-    %dbg:tracer(process, {fun
-    %                         ({trace, Pid, call, {M, handle_call, [Msg, From, _State]}}, ok) ->
-    %                             Trace = {trace, Pid, call, {M, handle_call, [Msg, From, '#state{}']}},
-    %                             io:format("~p\n\n", [Trace]);
-    %                         ({trace, Pid, call, {M, Fun, [Msg, _State]}}, ok) when
-    %                               Fun =:= handle_cast;
-    %                               Fun =:= handle_info ->
-    %                             Trace = {trace, Pid, call, {M, Fun, [Msg, '#state{}']}},
-    %                             io:format("~p\n\n", [Trace]);
-    %                         ({trace, Pid, return_from, {M, Fun, Arity}, {ok, Type}}, ok) ->
-    %                             TypeS = typelib:pp_type(Type),
-    %                             Trace = {trace, Pid, return_from, {M, Fun, Arity}, 'Type'},
-    %                             io:format("~p\nType = ~ts\n\n", [Trace, TypeS]);
-    %                         (Trace, ok) ->
-    %                             io:format("~p\n\n", [Trace])
-    %                     end, ok}),
-    %%dbg:p(all, [call, arity]),
-    %dbg:p(all, [call]),
-    %dbg:tpl(gradualizer_db, start_link, []),
-    %dbg:tpl(gradualizer_db, init, []),
-    %dbg:tpl(gradualizer_db, import_prelude, []),
-    %dbg:tpl(gradualizer_db, import_extra_specs, []),
-    %dbg:tpl(gradualizer_db, handle_call, []),
-    %dbg:tpl(gradualizer_db, handle_cast, []),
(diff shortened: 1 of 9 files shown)


src/gradualizer_app.erl
```diff
@@ -15,7 +15,8 @@
 %%%===================================================================
 
 start(_StartType, _StartArgs) ->
-    gradualizer_sup:start_link().
+    Opts = application:get_env(gradualizer, cli_options, []),
```
zuiderkwast, line 18: Gradualizer can also run from the Erlang shell and as a rebar3 plugin. These options are not only CLI, are they?

zuiderkwast (commented):
Very nice! I suppose @josefs would need to enable github actions..?

erszcz:
> Very nice! I suppose @josefs would need to enable github actions..?

Not really, turning GH actions on is done by adding the necessary files. As you can see, this PR is checked with the actions already. It's sufficient to merge it to take effect for the whole repo. Another question is whether @josefs is OK with it.

src/gradualizer_app.erl
erszcz: Hmm, the real takeaway of this change is that `prelude` default value is now baked into `gradualizer_db` itself, and unless overridden, prelude is imported directly after `init`. Otherwise, we were running into inconsistencies if the DB died and was restarted. Specifically, calling `import_prelude` from the `typechecker` wasn't right, because DB could be restarted before `typechecker` finished checking, and nothing would make the DB import prelude again on restart.

Good point, though, these options are "teleported" here from `gradualizer_cli:main()`, so in that case they are CLI options, but I haven't checked how the Rebar3 plugin runs the rest of the code or how it sets the options. As far as the defaults go, it should be working properly.

erszcz: I've renamed the app env var.

I've also taken a look at `src/rebar_prv_gradualizer.erl`. It seems Gradualizer as an app is not even started from there, which means the plugin relied on `application:ensure_all_started(gradualizer)` being called the first time `typechecker:type_check_forms` was called. This is not completely consistent, though, since the plugin tries to respect that each checked app might have its own set of gradualizer options. However, given the DB is started on the first call to `typechecker`, yet not stopped or reset after the first app is checked, the opts of the second and further checked apps aren't really respected (e.g. prelude won't be unloaded from the already running DB if opts of the second checked app contain `{prelude, false}`). In general, there are some details which need considering and aren't considered currently.

All in all, ultimately we might not want to store just `options`, but actually unpack each option from the list and `set_env` it directly. With that approach the Rebar3 plugin could decide which ought to be set "globally" and which can be overridden per checked app. The plugin could also decide if the Gradualizer app as a whole has to be restarted on each checked app.

zuiderkwast: OK. Do you want to take care of it later or now? I can merge now you're happy with it.

zuiderkwast (approved):

src/gradualizer_app.erl
erszcz: Let's leave it for later. The intention of this PR was to unblock CI on other PRs.

zuiderkwast:
The "Build passing" button (or what is it called?) on the README page still links to Travis CI. The code coverage button too maybe...

erszcz:
The build badge should be fixed with #345. Coverage is not enabled in GH actions yet.

PR #345, opened 2021-09-27
Fix the README CI badge, remove coverage badge

erszcz:
Coverage measurement is not enabled yet in GH Actions.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - master
```
zuiderkwast, line 7: There is no main branch afaik.

zuiderkwast:
Thanks!!

PR #347, opened 2021-09-30
Hardcode refinable true in all cases

erszcz:
This tries out the idea from https://github.com/josefs/Gradualizer/pull/346#discussion_r719382243 to assume that every type is refinable. Let's see how that works in practice. A temporary catch-all clause is also added to `pick_value` to handle all the types which don't have specific handlers defined already.


erszcz:
Ok, so in general we have quite a few failures with `refinable(_) -> true`, but with some tweaks to https://github.com/josefs/Gradualizer/blob/59b17f6e71215ce6bf0bcaf6998aa6b8226baef5/src/typechecker.erl#L3314 their number can be reduced from 31 to a few.

Then, it's a question of how specific we want the specs to be - by just making specs in tests more specific, we can get down to 1 failing test case: a spec for a binary. However, some of these specs are starting to be cumbersome, which might be a bit unappealing, but IMO is not a definite blocker.

All in all, removing `refinable` altogether seems possible with some extra work than done here so far, but I'd keep it a separate topic from #346.

zuiderkwast:
No, we don't want to require specific specs. Any() means we don't know anything, tuple() means we know less than if we'd have `{_, _, _}`, which is still unspecific. Map() is also something in between known an unknown. In those cases, we don't want to report any error here.

Maybe "refineable" isn't the right word for this but I guess we need the function.

zuiderkwast:
Thanks for trying this out btw!

erszcz:
I think this one can be closed, then. Nothing to merge here, but it'll stay here as documentation of the experiment.

zuiderkwast:
For the record, here's when we could be able to do for an inexact type like `tuple()` (meaning some unknown tuple type, not all tuples).

```Erlang
%% Here we can give an exhaustiveness error, since no tuple at all has been matched.
-spec f(tuple()) -> ok.
f(42) -> ok.

%% Here we can't give an error.
-spec g(tuple()) -> ok.
g({a, b, c}) -> ok.
```

Issue #355, opened 2021-10-14
Use Env everywhere where only TEnv is available


erszcz:
This is a follow up of [@zuiderkwast's remark that sometimes we only have TEnv where Env would actually be needed](https://github.com/josefs/Gradualizer/pull/341#discussion_r721272086). Refactoring this is a significant chunk of work, though:

```
$ rg Env#env.tenv src/typechecker.erl | wc -l
     153
```

But it should be done sooner or later, as [otherwise some options have to set globally, which is not ideal](https://github.com/josefs/Gradualizer/blob/051040400924dfcfe59f7a8936beb34b43278191/src/gradualizer_app.erl#L19).

PR #356, opened 2021-10-14
Stop infinite recursion in glb

erszcz:
When self-gradualizing with higher `union_size_limit` values, Gradualizer would fall into an infinite loop that I tracked to occur in `gradualizer_fmt.erl` - `test/should_pass/glb_infinite_loop1.erl` is a test module based on that.

~In order to test effectively, I hacked a way to set the limit using the `gradualizer` attribute, but I consider this temporary - I created https://github.com/josefs/Gradualizer/issues/355 to remember this needs a proper solution at some point.~ EDIT: now solved properly.

I think this test only exercises the `union` branches of `glb_ty`, not the `{type, _, 'fun', ...}` branches, but I think all the branches should call `glb` instead of `glb_ty` - otherwise, the `A` map is not updated, so we won't stop recursion as intended.

### DIFF
--- a/src/gradualizer_app.erl
+++ b/src/gradualizer_app.erl
@@ -16,15 +16,7 @@
 
 start(_StartType, _StartArgs) ->
     Opts = application:get_env(gradualizer, options, []),
-    set_union_size_limit(Opts),
     gradualizer_sup:start_link(Opts).
 
 stop(_State) ->
     ok.
-
-set_union_size_limit(Opts) ->
-    case lists:keyfind(union_size_limit, 1, Opts) of
-        false -> ok;
-        {union_size_limit, L} ->
-            persistent_term:put(gradualizer_union_size_limit, L)
-    end.
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -20,9 +20,6 @@
 
 -include("typelib.hrl").
 
-%% Performance hack: Unions larger than this value are replaced by any() in normalization.
--define(union_size_limit, persistent_term:get(gradualizer_union_size_limit, 30)).
-
 -define(verbose(Env, Fmt, Args),
         case Env#env.verbose of
             true -> io:format(Fmt, Args);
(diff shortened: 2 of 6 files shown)


erszcz:
Ok, I'll go for the `TEnv -> Env` refactor first, then. Afterwards, we can rebase this on top of the refactoring.

zuiderkwast:
OK, I was thinking of `add_type_pat` mainly (not a good name btw), so that it follows the same order as `type_check_expr`, but I see now that it already has env in the end. I think it was in the beginning earlier but maybe we've changed that already. Anyway, it's not important.

erszcz:
This is superseded by #368. I'm not sure if we want to merge them separately or just merge #368.

zuiderkwast:
You can decide. You could also rebase and squash the commits in #368 so we have 2 or 3 commits to merge without squash. Feel free to close this PR then.

erszcz:
@zuiderkwast What's the advantage of squash merging instead of just merging?

zuiderkwast:
> What's the advantage of squash merging

A more readable commit history. If we have a lot of changes back and forth, merge commits, etc. in a PR, I think it makes sense to squash them. If we have logically different changes within the same PR, I think it makes sense to keep them separate.

zuiderkwast:
It's not a big deal in this repo though. We already have a lot of commits for partial incomplete changes.

erszcz:
I think that for as along as a PR passes CI and is reviewed, and commit messages make sense, it's not bad.

Some commits are small:
- https://github.com/josefs/Gradualizer/commit/dd02b45eb5a0eb7d29099404fce244fb72267883
- https://github.com/josefs/Gradualizer/commit/c3aa5b162da79e4690a794105f434c1dfb02b50f
- https://github.com/josefs/Gradualizer/commit/121440ab989dd8ac8fcfc78c36ac692a35970a6a

And some are big:
- https://github.com/josefs/Gradualizer/commit/047402e2f8f4da61294c78a412e52a4756cd0a6c
- https://github.com/josefs/Gradualizer/pull/357/commits/6da7e3730f8a8b4a51494bad43dedb570e86f094

Because it makes sense to keep some changes separate, yet merge others together.

I'm trying to keep the commit history clean. I'll squash the obvious candidates, but sometimes I'm intentionally leaving commits separate because, for example, each of them fixes a small, yet different bug, and the commit messages contain the exact error logs the commits fix. Squashing in such a case leads to loss of info that might be useful sometime in the future.

zuiderkwast:
OK, fine. :+1: I'll merge without squash in most cases then (except when you tell me not to). Anyway, it doesn't matter if there is a separate PR or not IMO. If you say this one is superseeded by the other one, just close this one and possibly update the PR title of the other one.

erszcz:
Closing this one. Please see #368.

Issue #351, opened 2021-10-05
Self-gradualization: ebin/gradualizer.beam: The pattern [_F | Fs] on line 200 does not have type: []


erszcz:
A warning of this kind pops up multiple times in the self gradualization log and seems to be connected to list/non-empty-list exhaustiveness checking. When the order of clauses is changed as below:

```erlang
diff --git a/src/gradualizer.erl b/src/gradualizer.erl
index 8e89af7..f48b108 100644
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -193,9 +193,9 @@ add_source_file_and_forms_to_opts(File, Forms, Opts) ->

 %% Extract -gradualizer(Options) from AST
 -spec options_from_forms(gradualizer_file_utils:abstract_forms()) -> options().
+options_from_forms([]) -> [];
 options_from_forms([{attribute, _L, gradualizer, Opts} | Fs]) when is_list(Opts) ->
     Opts ++ options_from_forms(Fs);
 options_from_forms([{attribute, _L, gradualizer, Opt} | Fs]) ->
     [Opt | options_from_forms(Fs)];
-options_from_forms([_F | Fs]) -> options_from_forms(Fs);
-options_from_forms([]) -> [].
+options_from_forms([_F | Fs]) -> options_from_forms(Fs).
```

The warning becomes:

```
ebin/gradualizer.beam: The clause on line 201 at column 1 cannot be reached
```

It refers to `options_from_forms([_F | Fs])`, but it's a false positive - the clause is still necessary, as removing it causes tests to fail.

This was spotted when working on #341.

berbiche:
Hi,

I'm hitting this issue in multiple functions in a large codebase.
Most functions pattern match against a list of tuples:

``` erlang
-spec should_not_error([{integer(), integer()}]) -> integer().
should_not_error([{A, 0} | _]) ->
    A;
%% The pattern [{A, _} | _] does not have type: []
should_not_error([{A, _} | _]) ->
    A;
should_not_error([]) ->
    0.
```

Applying the commit 343d4f6018f23de7a9c45b309e12aa75919be5d1 partially resolves this issue but makes a valid test fail:
``` erlang
%% test/should_pass/user_type_in_pattern_body.erl
-module(user_type_in_pattern_body).

%% Check that the pattern type is normalized when recursively checking a pattern

-compile([export_all, nowarn_export_all]).

-type t()  :: {}.

[...]

PR #359, opened 2021-10-17
WIP: Fix list exhaustiveness checking regressions

erszcz:
List exhaustiveness checking extensions (possibly https://github.com/josefs/Gradualizer/commit/b3e5c31dad0ad4b0e2a3f0b82f44e6721e264311) introduced a number of regressions (`clause ... cannot be reached`, `pattern ... does not have type: []`) on quite common patterns found in everyday Erlang, which can be seen on the current master (051040400924dfcfe59f7a8936beb34b43278191): 

```
$ make gradualize
bin/gradualizer -pa ebin --color ebin
ebin/constraints.beam: The clause on line 43 at column 1 cannot be reached
ebin/gradualizer.beam: The operator '++' on line 125 at column 58 is expected to have type _TyVar-576460752303423293 which is too precise to be statically checked
ebin/gradualizer.beam: The pattern [_F | Fs] on line 200 at column 20 does not have type:
[]
ebin/gradualizer_bin.beam: The clause on line 94 at column 9 cannot be reached
ebin/gradualizer_cli.beam: The pattern ["never" | Args] on line 166 at column 14 does not have type:
[]
ebin/gradualizer_cli.beam: The clause on line 173 at column 1 cannot be reached
ebin/gradualizer_db.beam: The clause on line 356 at column 1 cannot be reached
ebin/gradualizer_db.beam: The pattern [{attribute, _L, record, {Name, Fields}} | Rest] on line 524 at column 21 does not have type:
[]
ebin/gradualizer_file_utils.beam: Call to undefined function epp:open/5 on line 66 at column 16
ebin/gradualizer_file_utils.beam: The pattern {ok, {_Module, [{abstract_code, no_abstract_code}]}} on line 80 at column 9 doesn't have the type {error, beam_lib, chnk_rsn()} | {ok, {module(), []}}
ebin/gradualizer_fmt.beam: The clause on line 495 at column 1 cannot be reached
ebin/gradualizer_highlight.beam: The pattern [$  | Str] on line 184 at column 13 does not have type:
[]
ebin/gradualizer_highlight.beam: The clause on line 188 at column 1 cannot be reached
...
```

This PR fixes these regressions, but is not complete because of a need to normalize list types. This, in turn, will currently trigger an infinite loop on common recursive list types, unless we use a cycle breaking mechanism like the one mentioned in https://github.com/josefs/Gradualizer/pull/324 and extend it to cover all recursive `normalize` clauses.

In this light, looking at the currently open PRs, I think this is the order to review&merge them:
- https://github.com/josefs/Gradualizer/pull/358 - just compile time warnings, no functionality fixes, so easy to review
- https://github.com/josefs/Gradualizer/pull/357 - `Env` is needed to properly pass `union_size_limit` around and be able to define it as a `-gradualizer` attribute in checked files
- https://github.com/josefs/Gradualizer/pull/356 - this fixes one of the inifinite loops in glb, but requires the `union_size_limit` to be defined as an attribute to ensure that tests are self-contained
- an extension of https://github.com/josefs/Gradualizer/pull/324 to cover all `normalize` clauses, so that we can safely add new recursive clauses; EDIT: this is now available as https://github.com/josefs/Gradualizer/pull/361
- this one, which adds normalising lists

Since this cannot work yet, I'm posting it as a draft.


erszcz:
For the record, this should fix https://github.com/josefs/Gradualizer/issues/351 when it's merged.

erszcz:
I've rebased this on top of the current master, but there are two open problems:

1. `normalize(T) != normalize^2(T)`, the same problem as described in https://github.com/josefs/Gradualizer/commit/665d3f1b6a27f1d88047ca3e0f3ea693248d4ce0, which is part of #369
2. Another infinite loop when running `make gradualize` possibly, but not certainly, caused by the above :|

erszcz:
One of the problems in this PR was this:

```
1> Type = fun (T) -> typelib:remove_pos(typelib:parse_type(T)) end.
#Fun<erl_eval.44.40011524>
3> Env = test_lib:create_env([]).
{env,#{},#{},#{},
     #{module => undefined,records => #{},types => #{}},
     false,false,true,30}
6> Ty0 = Type("string()").
{type,0,string,[]}
8> Ty1 = typechecker:normalize(typelib:remove_pos(Ty0), Env).
{type,0,list,[{type,0,char,[]}]}
9> Ty2 = typechecker:normalize(typelib:remove_pos(Ty1), Env).
{type,0,list,
      [{type,0,range,[{integer,0,0},{integer,0,1114111}]}]}
19> Ty3 = typechecker:normalize(typelib:remove_pos(Ty2), Env).
{type,0,list,
      [{type,0,range,[{integer,0,0},{integer,0,1114111}]}]}
22> Ty0 == Ty1.
false
20> Ty1 == Ty2.
false
21> Ty2 == Ty3.
true
```

However, the recent discussion in https://github.com/josefs/Gradualizer/pull/369 prompted me to realise the above problem is solvable by expanding builtin aliases "to the bottom", i.e. until there's nothing to expand anymore. It means I can finally fix the last broken test in this draft and finally make it a PR ready for review.

erszcz:
Ahhh, there's still an infinite loop when calling `make graudalize` due to some recursion, so CI was going to time out... thinking how to sidestep that 🤔

zuiderkwast:
> ```
> 8> Ty1 = typechecker:normalize(typelib:remove_pos(Ty0), Env).
> {type,0,list,[{type,0,char,[]}]}
> 9> Ty2 = typechecker:normalize(typelib:remove_pos(Ty1), Env).
> {type,0,list,
>       [{type,0,range,[{integer,0,0},{integer,0,1114111}]}]}
> ```

9 is the problem here IMO. Normalize shouldn't look at the element type inside the list.

erszcz:
> 9 is the problem here IMO. Normalize shouldn't look at the element type inside the list.

Hmm, so do you suggest `normalize` should not recurse to types being referenced in the type being normalised at all? I think `normalize` code from a commit before any of my changes suggests the contrary - please see https://github.com/josefs/Gradualizer/pull/359/commits/f044c35f2701a76a6eb26d8154c0c6e76980eb49.

What I had in mind when writing https://github.com/josefs/Gradualizer/pull/359#issuecomment-1025134693 was that `Ty1 == Ty2` if false, i.e. depending on "how far you expand a builtin type" it might or might not return the same value in a subsequent call to `normalize` on it. However, that's fixed with the change to `expand_builtin_aliases` in https://github.com/josefs/Gradualizer/pull/359/commits/f044c35f2701a76a6eb26d8154c0c6e76980eb49.

erszcz:
BTW, I think this PR is ready for review and merge in the current shape.

zuiderkwast:
> > 9 is the problem here IMO. Normalize shouldn't look at the element type inside the list.
> 
> Hmm, so do you suggest `normalize` should not recurse to types referenced in the type being normalised at all? I think `normalize` code from a commit before any of my changes suggests the contrary - please see

Yes, I believe making normalize recursive was a mistake. It was made recursive in 766ba271 which added the clause `normalize({type, _, record, [{atom, _, Name}|Fields]}, TEnv)`. Before that, care had been taken to make sure it's not recursive. Later, c23477c2 added safeguards to stop infinite recursion, which is unnecessary if normalize isn't recursive in the first place.

zuiderkwast:
> I believe making normalize recursive was a mistake

I have to revise this statement slightly. It needs to be recursive if we want to normlize `t1()` to `foo` given these definitions:

```Erlang
-type t1() :: t2().
-type t2() :: t3().
-type t3() :: foo.
```

... so I supposed it's good that we have the `Unfolded` parameter here, so we can detect loops and return an error in this case, as we talked about before.

test/should_fail/exhaustive_list_variants.erl
```diff
@@ -0,0 +1,12 @@
...
+-export([list_variants/1]).
+
+-type list_variant_t() :: {non_list, integer()}
+                        | {list, [integer()]}.
+
+-spec list_variants(list_variant_t()) -> ok.
+list_variants(T) ->
+    case T of
+        {non_list, _} -> ok
+    end.
```
zuiderkwast, line 12: This seems to have nothing to do with lists. What am I missing?

test/should_pass/list_exhaustiveness_checking_regressions.erl
```diff
@@ -0,0 +1,17 @@
...
+-export([f/2, g/1, h/1]).
+
+f( Needle, [Needle | _]) -> ok;
+f( Needle, [_ | Haystack]) -> f(Needle, Haystack);
+f(_Needle, []) -> ok.
+
+-spec g([a | b]) -> ok.
+g([a | _Haystack]) -> ok;
+g([_ | Haystack]) -> g(Haystack);
+g([]) -> ok.
```
zuiderkwast, line 12: Good test cases.

Without the 2nd or 3rd clause, do we detect the unexhaustiveness? If yes, are there should-fail tests for this already?

src/typelib.erl
```diff
@@ -48,7 +48,11 @@ pp_type(Type = {type, _, bounded_fun, _}) ->
     "fun(" ++ S ++ ")";
 pp_type({var, _, TyVar}) ->
     %% TODO: In type(), TyVar should be an atom but we use a string.
-    TyVar;
+    %% EDIT: Apparently, not everywhere :/
```
zuiderkwast, line 51: For explicit type variables, it's an atom. For implicit ones (auto-generated for fun parameters), it's a string. Perhaps it's good to describe that, before we forget it again. :-)

src/typechecker.erl
```diff
@@ -4174,14 +4197,15 @@ add_type_pat(CONS = {cons, P, PH, PT}, ListTy, Env, VEnv) ->
...
         {elem_ty, ElemTy, Cs1} ->
-            {_PatTy1, _UBound1, VEnv2, Cs2} =
+            {PatTy1, _UBound1, VEnv2, Cs2} =
                 add_type_pat(PH, normalize(ElemTy, Env), Env, VEnv),
             TailTy = normalize(type(union, [ListTy, type(nil)]), Env),
             {_PatTy2, _Ubound2, VEnv3, Cs3} = add_type_pat(PT, TailTy, Env, VEnv2),
+            PatTy = type(nonempty_list, [PatTy1]),
             NonEmptyTy = rewrite_list_to_nonempty_list(ListTy),
-            {NonEmptyTy, NonEmptyTy, VEnv3, constraints:combine([Cs1, Cs2, Cs3])};
+            {PatTy, NonEmptyTy, VEnv3, constraints:combine([Cs1, Cs2, Cs3])};
```
zuiderkwast, line 4208: This is hard do understand when you haven't looked at it for a while. :-)

I think some comments would be helpful, e.g.

* %% PatTy1 is the type exhausted by the cons-pattern head (PH)

* %% _PatTy2 is the type exhausted by the cons-pattern tail (PT), which is irrelevant (why??? example???)

* %% PatTy is the type exhausted by the cons pattern.

src/typechecker.erl
```diff
@@ -762,6 +762,13 @@ normalize_rec({type, _, union, Tys} = Type, Env, Unfolded) ->
                 Ts  -> type(union, Ts)
             end
     end;
+normalize_rec({type, _, list, [ElemTy]} = Type, Env, Unfolded) ->
+    case maps:get(mta(Type, Env), Unfolded, no_type) of
+        {type, NormType} -> NormType;
+        no_type ->
+            UnfoldedNew = maps:put(mta(Type, Env), {type, Type}, Unfolded),
+            type(list, [normalize_rec(typelib:remove_pos(ElemTy), Env, UnfoldedNew)])
+    end;
```
zuiderkwast, line 771: Please don't recurse into the element type. (I don't think we should recurse into tuple, map and record elements either. If we do, I think it's the wrong place to do it.)

Also don't add `remove_pos`. It is recursive on the syntactic level and should only be done when a type comes into the system (from the parser or from gradualizer_db). I'm working on a small fix to remove all redundant remove_pos calls.

Doesn't the refinement work without this?

src/typechecker.erl
```diff
@@ -834,9 +841,9 @@ expand_builtin_aliases({type, Ann, maybe_improper_list, []}) ->
 expand_builtin_aliases({type, Ann, nonempty_list, []}) ->
     {type, Ann, nonempty_list, [{type, Ann, any, []}]};
 expand_builtin_aliases({type, Ann, string, []}) ->
-    {type, Ann, list, [{type, Ann, char, []}]};
+    {type, Ann, list, [expand_builtin_aliases({type, Ann, char, []})]};
 expand_builtin_aliases({type, Ann, nonempty_string, []}) ->
-    {type, Ann, nonempty_list, [{type, Ann, char, []}]};
+    {type, Ann, nonempty_list, [expand_builtin_aliases({type, Ann, char, []})]};
```
zuiderkwast, line 846: I don't mind this, but I don't think it's necessary if it's only used by normalize. It shouldn't look inside the element type.

src/typechecker.erl
```diff
@@ -3401,18 +3408,32 @@ check_clauses(Env, ArgsTy, ResTy, Clauses, Caps) ->
...
+
+                           ArgsTy =/= any
+                           andalso lists:all(fun (Ty) -> refinable(Ty, Env) end, ArgsTy),
+
+                           lists:all(fun no_guards/1, Clauses),
+
+                           is_list(RefinedArgsTy)
+                           andalso lists:any(fun (T) -> T =/= type(none) end, RefinedArgsTy))
+    of
+        true ->
```
zuiderkwast, line 3421: TBH, I don't think this change improves on clarity.

Adding line breaks (and comments?) can be useful, but moving the logic to a function below mostly hides what's happening from the view.

Another style I'd be fine with is

```Erlang
if
    Env#env.exhaus andalso
    ... andalso
    ... ->
        %% do exhaustiveness-checking
    true ->
        %% no exhaustiveness-checking
```

src/typechecker.erl
```diff
@@ -3812,6 +3833,8 @@ refinable(RefinableTy, Env, Trace)
                 not_found -> false
             end
     end;
+refinable(?type(range, _), _Env, _Trace) ->
+    true;
```
zuiderkwast, line 3837: Makes sense. No test for this? I would expect to see a known problem solved or a new test added.

test/should_fail/exhaustive_list_variants.erl
erszcz: `list_variant_t()` has a variant which is basically a tagged list - hence its name. This tests exhaustiveness checking on types which have such variants. Well, the name could say `list_variants_exhaustiveness_test`, but it seemed self-explanatory - maybe it's not, if so I'll change the name.

src/typechecker.erl
erszcz: To be honest, to figure it out I had to look at an execution trace of a specific minimal example. I wouldn't be able to even start without that. Only then names like `PH` started making sense. I was not sure either why `PatTy1` was ignored, given it turned out to be useful.

In this light:

> %% _PatTy2 is the type exhausted by the cons-pattern tail (PT), which is irrelevant (why??? example???)

I'm not 100% sure it's irrelevant, but it didn't seem so yet. Maybe we just haven't found an example where it's relevant. That's why I think self-gradualizing is an important step for the project - this entire PR based on false positives found thanks to `make gradualize`.

I'll add some comments.

src/typelib.erl
erszcz: Yeah, this was not very consistent...
https://github.com/josefs/Gradualizer/commit/0e898384d640363efb13cd30e8d4bd472df4bec9#diff-c46e88595065283f0502281ada54f0a29daeff24290d35ad89c44161760b2753R263-R265

But your comment clears it up :+1:

src/typechecker.erl
erszcz: To be clear, I want the intermediate result printed as a separate trace message, to understand which path is taken and why:

```erlang
{trace,<0.9.0>,call,{typechecker,is_non_exhaustive,[true,true,true,true]}}

{trace,<0.9.0>,return_from,{typechecker,is_non_exhaustive,4},true}
```

That's a tiny snippet, but consider:

```
{trace,<0.9.0>,call,
    {typechecker,type_check_forms,
        [[{attribute,
              {1,1},
              file,
              {"test/should_fail/nonempty_list_match_in_head_nonexhaustive.erl",
               1}},
          {attribute,{1,2},module,nonempty_list_match_in_head_nonexhaustive},
          {attribute,{3,2},export,[{f,1}]},
          {attribute,
              {12,2},
              type,
              {t,{type,
                     {12,14},
                     union,
                     [{type,{12,14},tuple,[]},
                      {type,{12,19},list,[{user_type,{12,20},t,[]}]}]},
                 []}},
          {attribute,
              {14,2},
              spec,
              {{f,1},
               [{type,
                    {14,8},
                    'fun',
                    [{type,{14,8},product,[{user_type,{14,9},t,[]}]},
                     {atom,{14,17},ok}]}]}},
          {function,
              {15,1},
              f,1,
              [{clause,
                   {15,1},
                   [{cons,{15,3},{var,{15,4},'_'},{var,{15,6},'_'}}],
                   [],
                   [{atom,{15,13},ok}]}]},
          {eof,{16,1}}],
         []]}}

...

{trace,<0.9.0>,return_from,
       {typechecker,mta,2},
       {type,0,list,[{user_type,0,t,[]}]}}

{trace,<0.9.0>,return_from,
       {typechecker,stop_refinable_recursion,3},
       {proceed,{set,3,16,16,8,80,48,
                     {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
                     {{[],[],[],[],[],[],
                       [{nonempty_list_match_in_head_nonexhaustive,t,0}],
                       [],[],[],[],[],[],
                       [{type,0,union,
                              [{type,0,tuple,[]},
                               {type,0,list,[{user_type,0,t,[]}]}]}],
                       [{type,0,list,[{user_type,0,t,[]}]}],
                       []}}}}}

{trace,<0.9.0>,call,
    {typechecker,refinable,
        [{user_type,0,t,[]},
         {env,
             #{{f,1} =>
                   [{type,0,bounded_fun,
                        [{type,0,'fun',
                             [{type,0,product,[{user_type,0,t,[]}]},
                              {atom,0,ok}]},
                         []]}]},
             #{},#{},
             #{module => nonempty_list_match_in_head_nonexhaustive,
               records => #{},
               types =>
                   #{{t,0} =>
                         {[],
                          {type,0,union,
                              [{type,0,tuple,[]},
                               {type,0,list,[{user_type,0,t,[]}]}]}}}},
             false,false,true,30},
         {set,3,16,16,8,80,48,
             {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
             {{[],[],[],[],[],[],
               [{nonempty_list_match_in_head_nonexhaustive,t,0}],
               [],[],[],[],[],[],
               [{type,0,union,
                    [{type,0,tuple,[]},{type,0,list,[{user_type,0,t,[]}]}]}],
               [{type,0,list,[{user_type,0,t,[]}]}],
               []}}}]}}

{trace,<0.9.0>,call,
    {typechecker,stop_refinable_recursion,
        [{user_type,0,t,[]},
         {env,
             #{{f,1} =>
                   [{type,0,bounded_fun,
                        [{type,0,'fun',
                             [{type,0,product,[{user_type,0,t,[]}]},
                              {atom,0,ok}]},
                         []]}]},
             #{},#{},
             #{module => nonempty_list_match_in_head_nonexhaustive,
               records => #{},
               types =>
                   #{{t,0} =>
                         {[],
                          {type,0,union,
                              [{type,0,tuple,[]},
                               {type,0,list,[{user_type,0,t,[]}]}]}}}},
             false,false,true,30},
         {set,3,16,16,8,80,48,
             {[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},
             {{[],[],[],[],[],[],
               [{nonempty_list_match_in_head_nonexhaustive,t,0}],
               [],[],[],[],[],[],
               [{type,0,union,
                    [{type,0,tuple,[]},{type,0,list,[{user_type,0,t,[]}]}]}],
               [{type,0,list,[{user_type,0,t,[]}]}],
               []}}}]}}

{trace,<0.9.0>,call,
    {typechecker,mta,
        [{user_type,0,t,[]},
         {env,
             #{{f,1} =>
                   [{type,0,bounded_fun,
                        [{type,0,'fun',
                             [{type,0,product,[{user_type,0,t,[]}]},
                              {atom,0,ok}]},
                         []]}]},
             #{},#{},
             #{module => nonempty_list_match_in_head_nonexhaustive,
               records => #{},
               types =>
                   #{{t,0} =>
                         {[],
                          {type,0,union,
                              [{type,0,tuple,[]},
                               {type,0,list,[{user_type,0,t,[]}]}]}}}},
             false,false,true,30}]}}

{trace,<0.9.0>,return_from,
       {typechecker,mta,2},
       {nonempty_list_match_in_head_nonexhaustive,t,0}}

{trace,<0.9.0>,return_from,{typechecker,stop_refinable_recursion,3},stop}

{trace,<0.9.0>,return_from,{typechecker,refinable,3},true}

{trace,<0.9.0>,return_from,{typechecker,refinable,3},true}

{trace,<0.9.0>,return_from,{typechecker,'-refinable/3-fun-0-',3},true}

{trace,<0.9.0>,return_from,{typechecker,refinable,3},true}

{trace,<0.9.0>,return_from,{typechecker,refinable,3},true}

{trace,<0.9.0>,return_from,{typechecker,refinable,2},true}

{trace,<0.9.0>,return_from,
       {typechecker,'-check_exhaustiveness/6-fun-2-',2},
       true}

{trace,<0.9.0>,call,
       {typechecker,'-fun.no_guards/1-',
                    [{clause,{15,1},
                             [{cons,{15,3},{var,{15,4},'_'},{var,{15,6},'_'}}],
                             [],
                             [{atom,{15,13},ok}]}]}}

{trace,<0.9.0>,call,
       {typechecker,no_guards,
                    [{clause,{15,1},
                             [{cons,{15,3},{var,{15,4},'_'},{var,{15,6},'_'}}],
                             [],
                             [{atom,{15,13},ok}]}]}}

{trace,<0.9.0>,return_from,{typechecker,no_guards,1},true}

{trace,<0.9.0>,return_from,{typechecker,'-fun.no_guards/1-',1},true}

{trace,<0.9.0>,call,
       {typechecker,'-check_exhaustiveness/6-fun-0-',[{user_type,0,t,[]}]}}

{trace,<0.9.0>,call,{typechecker,type,[none]}}

{trace,<0.9.0>,call,{typechecker,type,[none,[]]}}

{trace,<0.9.0>,return_from,{typechecker,type,2},{type,0,none,[]}}

{trace,<0.9.0>,return_from,{typechecker,type,1},{type,0,none,[]}}

{trace,<0.9.0>,return_from,
       {typechecker,'-check_exhaustiveness/6-fun-0-',1},
       true}

{trace,<0.9.0>,call,{typechecker,is_non_exhaustive,[true,true,true,true]}}

{trace,<0.9.0>,return_from,{typechecker,is_non_exhaustive,4},true}

{trace,<0.9.0>,exception_from,
       {typechecker,check_exhaustiveness,6},
       {throw,{nonexhaustive,{15,1},[{tuple,0,[]}]}}}

{trace,<0.9.0>,exception_from,
       {typechecker,check_clauses,5},
       {throw,{nonexhaustive,{15,1},[{tuple,0,[]}]}}}

{trace,<0.9.0>,exception_from,
       {typechecker,check_clauses_fun,3},
       {throw,{nonexhaustive,{15,1},[{tuple,0,[]}]}}}

{trace,<0.9.0>,exception_from,
       {typechecker,type_check_function,2},
       {throw,{nonexhaustive,{15,1},[{tuple,0,[]}]}}}

{trace,<0.9.0>,return_from,
       {typechecker,'-type_check_forms/2-fun-0-',5},
       [{nonexhaustive,{15,1},[{tuple,0,[]}]}]}

{trace,<0.9.0>,return_from,
       {typechecker,type_check_forms,2},
       [{nonexhaustive,{15,1},[{tuple,0,[]}]}]}

test/should_fail/nonempty_list_match_in_head_nonexhaustive.erl: Nonexhaustive patterns on line 15 at column 1
	{}
```

This trace, from calling `typechecker,type_check_forms` to returning, is ~22k lines long. And that's running the typechecker on a minimised test example.

src/typechecker.erl
erszcz: This changed was prompted by our discussion in the other issue - https://github.com/josefs/Gradualizer/pull/369#issuecomment-1024977834. Specifically:

>> Do you have an example in which this turns into infinite recursion?

> I think the problem here is not infinite recursion - ultimately builtin type expansion is well defined and ends at some point. However, I remember seeing a buggy check result, which when traced was caused by a mismatch between a string() and a [char()] - or was it [char()] vs [0..16#10ffff]? - the point is it was a mismatch between a more and less expanded builtin type. I'll find an exact example.

We have to expand "to the bottom", as otherwise we'll get half-expanded builtin types which don't compare equal, where they should, and therefore false positives in the output.

erszcz: BTW, check the commit message - this change is introduced by "Fix exhaustive_refinable_map_variants":

```
$ fd exhaustive_refinable_map_variants
test/should_fail/exhaustive_refinable_map_variants.erl
```

src/typechecker.erl
erszcz: As above, it's tested by `test/should_fail/exhaustive_refinable_map_variants.erl`. Specifically, it's exercised by a `string()` being ultimately expanded to a `{type,0,list,[{type,0,range,[{integer,0,0},{integer,0,1114111}]}]}`.

src/typechecker.erl
erszcz: To be even more clear, I don't care that much about the indentation in between the new function arguments. The important part, of course, is extracting the compound condition to a new function, so it's traceable.

test/should_pass/list_exhaustiveness_checking_regressions.erl
erszcz: Hmmm, good point, I don't think there are explicit tests for this.

BTW, we don't detect anything in the case of `f`, but it's because it lacks a spec. However, it's like that intentionally, since we were getting the regressions (false positives) on such spec-less functions when running `make gradualize`.

src/typechecker.erl
erszcz: > Doesn't the refinement work without this?

It doesn't in the current shape of the PR :( I was specifically trying that out after your statement about `normalize` (https://github.com/josefs/Gradualizer/pull/359#issuecomment-1025886673):

> Yes, I believe making normalize recursive was a mistake. It was made recursive in 766ba27 [...]

I was thinking about it a bit. This PR fixes regressions introduced when extending the exhaustiveness checking mechanism to non-trivial types. `normalize` was already recursive when I started that work. This PR makes regressions which were evident when running `make gradualize` finally go away. Therefore, I propose merging it without first reverting to non-recursive `normalize`. `normalize` is already recursive on `master`, this PR fixes known issues and doesn’t break anything in return, so I think it’s a reasonable step.

Then, I’d like to focus on getting back to recursive type problems. At this point it might turn out it’s actually best to return `normalize` to being non-recursive (or non-recursive where possible) and push subsequent `normalize` calls down where they’re actually needed. However, I think it might be quite a significant effort, so I think it makes sense to make it a next step, given the work in this PR already brings visible improvement. In other words: let's chunk the work.

erszcz:
Thanks for the review, good points. Please note I'm still working out a bug that still in this code - another infinite loop :/ I'll make it a draft again until it's done.

test/should_fail/exhaustive_list_variants.erl
zuiderkwast: ... but the list is just matched against the `_` pattern, which matches all types (lists or not) so I still can't see why any list-specific code would kick in...

zuiderkwast: ... apart from pick_value. Is that it?

erszcz:
> Please note I'm still working out a bug in this code - another infinite loop :/

I've pushed https://github.com/josefs/Gradualizer/pull/359/commits/9731f6ec3fb645de46b9724f03ba410a6433541f which reproduces the problem. I'm tracing the code and looking for what's going on, but so far it seems there's a mutual recursion between `glb` and `normalize`, which continuously expands the recursive type https://github.com/josefs/Gradualizer/commit/9731f6ec3fb645de46b9724f03ba410a6433541f#diff-21a419b75f6024ef182cc2ff13d28b44ca89b043732813904b58a26e2aa9bc8dR12.

This is an example of the mistake of treating more and less expanded instances of the same type as distinct types which we touched on in https://github.com/josefs/Gradualizer/pull/369#issuecomment-1024450091. As you can see

```
-type t() :: {} | [t()].
```

does not have recursion on the top level according to your definition, but it's still problematic. Do you think such types should be supported?

Trace snippet (cover your eyes ;p):

```
{trace,<0.9.0>,call,
 {typechecker,glb,
  [{type,0,list,
    [{type,0,union,
      [{type,0,list,
        [{type,0,union,
          [{type,0,list,
            [{type,0,union,
              [{type,0,list,
                [{type,0,union,
                  [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
               {type,0,tuple,[]}]}]},
           {type,0,tuple,[]}]}]},
       {type,0,tuple,[]}]}]},
   {type,0,union,
    [{type,0,list,
      [{type,0,union,
        [{type,0,list,
          [{type,0,union,
            [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
         {type,0,tuple,[]}]}]},
     {type,0,tuple,[]}]},
   #{{{type,0,list,
       [{type,0,union,
         [{type,0,list,
           [{type,0,union,
             [{type,0,list,
               [{type,0,union,
                 [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
              {type,0,tuple,[]}]}]},
          {type,0,tuple,[]}]}]},
      {type,0,list,
       [{type,0,union,
         [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]}} =>
      0,
     {{type,0,list,
       [{type,0,union,
         [{type,0,list,
           [{type,0,union,
             [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
          {type,0,tuple,[]}]}]},
      {type,0,union,[{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}} =>
      0,
     {{type,0,nonempty_list,
       [{type,0,union,
         [{type,0,list,
           [{type,0,union,
             [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
          {type,0,tuple,[]}]}]},
      {type,0,nonempty_list,[{user_type,0,t,[]}]}} =>
      0,
     {{type,0,union,
       [{type,0,list,
         [{type,0,union,
           [{type,0,list,
             [{type,0,union,
               [{type,0,list,
                 [{type,0,union,
                   [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
                {type,0,tuple,[]}]}]},
            {type,0,tuple,[]}]}]},
        {type,0,tuple,[]}]},
      {type,0,union,
       [{type,0,list,
         [{type,0,union,
           [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
        {type,0,tuple,[]}]}} =>
      0,
     {{type,0,union,
       [{type,0,list,
         [{type,0,union,
           [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
        {type,0,tuple,[]}]},
      {user_type,0,t,[]}} =>
      0},
   {env,
    #{{f,1} =>
       [{type,0,bounded_fun,
         [{type,0,'fun',[{type,0,product,[{user_type,0,t,[]}]},{atom,0,ok}]},
          []]}]},
    #{},#{},
    #{module => nonempty_list_match_in_head_nonexhaustive,records => #{},
      types =>
       #{{t,0} =>
          {[],
           {type,0,union,
            [{type,0,tuple,[]},{type,0,list,[{user_type,0,t,[]}]}]}}}},
    false,false,true,30}]}}

{trace,<0.9.0>,call,
 {typechecker,normalize,
  [{type,0,list,
    [{type,0,union,
      [{type,0,list,
        [{type,0,union,
          [{type,0,list,
            [{type,0,union,
              [{type,0,list,
                [{type,0,union,
                  [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
               {type,0,tuple,[]}]}]},
           {type,0,tuple,[]}]}]},
       {type,0,tuple,[]}]}]},
   {env,
    #{{f,1} =>
       [{type,0,bounded_fun,
         [{type,0,'fun',[{type,0,product,[{user_type,0,t,[]}]},{atom,0,ok}]},
          []]}]},
    #{},#{},
    #{module => nonempty_list_match_in_head_nonexhaustive,records => #{},
      types =>
       #{{t,0} =>
          {[],
           {type,0,union,
            [{type,0,tuple,[]},{type,0,list,[{user_type,0,t,[]}]}]}}}},
    false,false,true,30}]}}

{trace,<0.9.0>,return_from,
 {typechecker,normalize,2},
 {type,0,list,
  [{type,0,union,
    [{type,0,list,
      [{type,0,union,
        [{type,0,list,
          [{type,0,union,
            [{type,0,list,
              [{type,0,union,
                [{type,0,list,
                  [{type,0,union,
                    [{type,0,list,[{user_type,0,t,[]}]},{type,0,tuple,[]}]}]},
                 {type,0,tuple,[]}]}]},
             {type,0,tuple,[]}]}]},
         {type,0,tuple,[]}]}]},
     {type,0,tuple,[]}]}]}}

{trace,<0.9.0>,return_to,{typechecker,glb,4}}
```

test/should_fail/nonempty_list_match_in_head_nonexhaustive.erl
```diff
@@ -0,0 +1,15 @@
+-module(nonempty_list_match_in_head_nonexhaustive).
+
+-export([f/1]).
+
+%% For the infinite loop bug to happen all the conditions must be met:
+%% - the function matches on a non-empty list
+%% - the type is a union of something and a list of itself - if it's not a union,
+%%   typechecker doesn't loop
+%% - the type is recursive - a list of something else doesn't trigger the infinite loop
+%% - the pattern is bound to a variable (doesn't matter if used or unused)
```
zuiderkwast, line 10: Interesting.

Since you found out that also maps are normalized inside their elements, is there a similar case for maps? `t() :: {} | #{t() => t()}`?

Btw, I think an issue is a better place for such detailed bug information. Especially when the bug is fixed, such description makes less sense in the source code. A link to the issue is enough as motivation for the test.

zuiderkwast:
> ```
> -type t() :: {} | [t()].
> ```
>
> does not have recursion on the top level according to your definition, but it's still problematic. Do you think such types should be supported?

Yes, it is a tree type with leaf and a list of branches. We should support this. A perhaps more realistic variant of this:

```Erlang
-type tree(A) :: {leaf, A} | [tree(A)].
```

src/typechecker.erl
zuiderkwast: I get it. :-)

zuiderkwast: In that case, I'd prefer functions with more meaningful names and parameters. Adding functions should also improve readability of the code, not only tracing.

Idea:

```Erlang
case all_refinable(ArgsTy, Env) andalso
     no_clause_has_guards(Clauses) andalso
     some_type_is_not_none(RefinedArgsTy) of
    true when Env#env.exhaust ->
        ...
```

where the implementations are something like

```Erlang
all_refineable(any, _Env) -> false;
all_refineable(Types, Env) -> lists:all(fun (Ty) -> refinable(Ty, Env) end, Types).

no_clause_has_guards(Clauses) ->
    lists:all(fun no_guards/1, Clauses).

some_type_is_not_none(Types) ->
    lists:any(fun (T) -> T =/= type(none) end, Types).
```

It's still good for tracing, isn't it?

Btw, documenting the what this function does and what the parameters are would be very useful:

```Erlang
%% Checks if clauses completely covers the types of the arguments and wtf about argstys refinedargsty(?)
check_exhaustiveness(Env, ArgsTy, Clauses, RefinedArgsTy, VarBindsList, Css)
```

erszcz: Good idea, I've refactored the code and added some docs 👍

test/should_fail/exhaustive_list_variants.erl
erszcz: Yes, `typechecker:refinable()` and `gradualizer_lib:pick_value()`. Now I also extended this file to trigger exhaustiveness checking of (non)empty lists in the list variant.

test/should_pass/list_exhaustiveness_checking_regressions.erl
erszcz: There are now in `test/should_fail/exhaustive_list_variants.erl`.

erszcz:
> I still think one way to solve it is to prevent normalize from going into the elements. Then, whatever breaks needs to be handled differently.

I've been thinking of it a bit more and while I have nothing against converting `normalize` back to Head Normal Form, I think it might not solve the recursion problem on an infinite recursive type.

> For example, when the typechecker checks something with T1 and T2 and figures out that both are lists, then it needs to normalize the element types of both lists before comparing the element types. WDYT?

I think the current problem we're experiencing is `glb` trying to compare instances of the same recursive type expanded to different depths. `normalize` not looking at the element type doesn't move us forward, as the loop is not in `normalize` itself (please check the trace snippet I shared above, it's not `normalize_rec` looping, it's always a top-level call to `normalize` returning the expanded user type `t()`). This means we're constantly asking to expand the `t()` (the element type), whereas at that point it's already "expanded enough" (i.e. it's expanded further than the simplest recursive form) - `glb` simply doesn't see the equivalence of the less and more expanded instances of `t()`.

I think we're experiencing this problem because of the lack of explicitness in marking a recursive type as such. Let me quote a fragment of TaPL chapter 20, Recursive Types:

> The recursive equation specifying this infinite tree type [there's a diagram just above this paragraph] is similar to the equation specifying the recursive `factorial` function on page 52. Here, as there, it is convenient to make this equation into a proper definition by moving the "loop" over to the right-hand side of the = sign. We do this by introducing an explicit recursion operator μ for types:
> ```
> NatList = μX. <nil:Unit, cons:{Nat,X}>;
> ```

In other words, they introduce an explicit μ binder for recursive types similar to the λ binder for lambda abstraction terms. In Erlang syntax that would be something like:

```
NatList = {mu, [{var, _, 'X'}], {type, _, union, [{type, _, nil},
                                                  {type, _, cons, {{type, _, non_neg_integer, []}, {var, _, 'X'}}}]}}.
```

Such a term identifies a `NatList` independently of its length:

```
[] :: type(nil) :: NatList.
[1] :: type(cons, {type(non_neg_integer(), type(nil)}) :: NatList.
[1,2] :: type(cons, {type(non_neg_integer()), type(cons, {non_neg_integer(), type(nil)})}) :: NatList.
```

The example TaPL implementations show how to expand/contract such a recursive type to reach the simplest recursive case and therefore [allow for comparing different instances of the recursive type](https://github.com/erszcz/tapl-erlang/blob/b364c24416533826daaa3bd5970e5560ffa8838b/equirec/core.ml#L43-L47). However, in TaPL example languages, [this μ type operator is also present in the concrete syntax of type annotations (as `Rec X. ...`)](https://github.com/erszcz/tapl-erlang/blob/b364c24416533826daaa3bd5970e5560ffa8838b/fullequirec/test.f#L93), which makes the implementation easier.

I'm not sure we can reliably compare extended/contracted recursive type instances without such a device in the abstract type representation. WDYT?

erszcz:
I've finally translated and studied a bit [the TaPL example typechecker with recursive types to Erlang](https://github.com/erszcz/tapl-erlang/tree/e3fa689d2e9b0a91eb6e72245e4904965fddc01f/apps/fullequirec). It convinces me that we should be good with Head Normal Form. The μ binder, encoded as `TyRec` in:

```
type ty =
    TyVar of int * int
  | TyId of string
  | TyArr of ty * ty
  | TyUnit
  | TyRecord of (string * ty) list
  | TyRec of string * ty
  | TyVariant of (string * ty) list
  | TyBool
  | TyString
  | TyFloat
  | TyNat
```

is roughly equivalent to our `{user_type, _, _, _}`. I think it's worth pursuing #376 before getting back to this PR.

erszcz:
Superseded by #404.

Issue #247, opened 2020-01-09
Remove unnecessary clauses matching ann_type in various places


josefs:
Since #228 we don't use `ann_type` in our internal representation of types. But we still have a whole bunch of clauses which still deals with `ann_types`. These should be removed.

zuiderkwast:
Closed by #253.

PR #363, opened 2021-10-24
Property based tests to find bugs

erszcz:
This PR sketches a property based testing framework for Gradualizer. It's based on a vendored version of [Proper's Erlang abstract code generator](https://github.com/proper-testing/proper/blob/9f6a6501430479bed66d08cd795cd34d36ec83aa/src/proper_erlang_abstract_code.erl) used to [property test the Erlang compiler](https://github.com/erlang/otp/blob/f7d00dd023b15ab787a5f420a263dd320914213a/lib/compiler/test/property_test/compile_prop.erl#L46).

On top of that, it uses an ad-hoc construction to define type envs of mutually recursive types - this already allows to find timeouts in `normalize` runs which I think are the infinite loops found by other testing means such as #324. There's more work needed on that front.

I've not included any `src/` fixes found when running the properties so far in this PR, so that it's relatively easy to see if the properties defined here actually find bugs. However, there's already a number of them that I've stashed and will PR separately.

This builds on top of https://github.com/josefs/Gradualizer/pull/357, since it deals with creating environments for the properties a little bit. However, if we want to make sure the #357 itself does not introduce new bugs, I can rebase this PR onto the current master which does not include #357 yet.


src/typechecker.erl
```diff
@@ -1041,6 +1045,9 @@ expect_list_type(Ty, _, _) ->
 rewrite_list_to_nonempty_list({type, Ann, T, [ElemTy]})
   when T == list orelse T == nonempty_list ->
     {type, Ann, nonempty_list, [ElemTy]};
+rewrite_list_to_nonempty_list({type, Ann, T, []})
+  when T == string orelse T == nonempty_string ->
```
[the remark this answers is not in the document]
erszcz: This was a bit more tricky. The denormalized `string()` came from https://github.com/josefs/Gradualizer/blob/b09458159e72ef16598f4ff8499ef64a2a816061/src/typechecker.erl#L4142 since when typechecking `{match, _, _, _}` `add_type_pat` is called consecutively - https://github.com/josefs/Gradualizer/blob/b09458159e72ef16598f4ff8499ef64a2a816061/src/typechecker.erl#L4218-L4221 - and output of the first call is used as input to the second one. If `Pat2` happened to be a string, then we would've gotten a denormalized `string()` type back. Fixed in the last commit.

src/typechecker.erl
```diff
@@ -720,6 +722,8 @@ glb_ty({type, _, Name, Args1}, {type, _, Name, Args2}, A, Env)
 glb_ty(_Ty1, _Ty2, _A, _Env) -> {type(none), constraints:empty()}.
 
 -spec has_overlapping_keys(type(), env()) -> boolean().
+has_overlapping_keys({type, _, map, any}, _Env) ->
+    true;
```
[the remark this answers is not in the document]
erszcz: Just adding `{ann_type, ...}` seems to solve the problem :)

rebar.config
```diff
@@ -0,0 +1,4 @@
+{deps,
+ [
+  {proper, {git, "https://github.com/proper-testing/proper.git", {branch, "master"}}}
+ ]}.
```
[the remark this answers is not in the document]
erszcz: > How about some simple hand-written lines for git cloning proper in the Makefile?

Hmmm, to be honest I'm not convinced - I don't really see the benefit of using a custom Makefile in Gradualizer. A lot of work has been put into Rebar3 and not having to write custom Makefile rules is IMO one of the positive outcomes.

Proper is now a test-only dependency, though, so it won't pollute other projects when Gradualizer itself is pulled in as a dependency.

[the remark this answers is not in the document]
erszcz: For some context:

```
$ rm -rf _build
21:53:21 erszcz @ x6 : ~/work/erszcz/gradualizer (property-based-tests %)
$ time r3 compile
...

real	0m3.370s
user	0m4.288s
sys	0m0.669s
```

```
$ make clean
$ time make
...

real	0m15.742s
user	0m11.484s
sys	0m4.863s
```

src/typechecker.erl
```diff
@@ -782,6 +780,8 @@ normalize({op, _, _, _Arg} = Op, _Env) ->
     erl_eval:partial_eval(Op);
 normalize({op, _, _, _Arg1, _Arg2} = Op, _Env) ->
     erl_eval:partial_eval(Op);
+normalize({ann_type, Ann, [Var, Ty]}, Env) ->
+    {ann_type, Ann, [Var, normalize(typelib:remove_pos(Ty), Env)]};
```
zuiderkwast, line 784: `remove_pos` is deep so I don't see why it's needed here. It's only done in normalize where we expand remote and user-defined types.

src/typechecker.erl
```diff
@@ -4154,7 +4151,7 @@ add_type_pat(CONS = {cons, P, PH, PT}, ListTy, Env, VEnv) ->
 add_type_pat(String = {string, P, _}, Ty, Env, VEnv) ->
     case subtype(type(string), Ty, Env) of
         {true, Cs} ->
-            {type(none), type(string), VEnv, Cs};
+            {type(none), normalize(type(string), Env), VEnv, Cs};
```
zuiderkwast, line 4154: :+1:

test/property_test/gradualizer_erlang_abstract_code.erl
```diff
@@ -0,0 +1,3117 @@
+%%% This file is part of PropEr.
```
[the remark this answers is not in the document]
zuiderkwast: Good points about the abstract code and this proper file. I think it would be good to add a note about this in the LICENCE file. (Not necessarily in this PR.)

.github/workflows/build-and-test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     name: Erlang ${{matrix.otp}}
     strategy:
       matrix:
-        otp: ['22.3.4.20', '23.3.4.7', '24.1']
+        otp: ['23.3.4.7', '24.1']
```
zuiderkwast, line 16: Don't we support 22 anymore?

zuiderkwast (approved):
LGTM

.github/workflows/build-and-test.yml
erszcz: There's a problem with building property based tests on 22 - https://github.com/erszcz/Gradualizer/runs/3983416094?check_suite_focus=true. I rechecked with the latest version of this branch again, just to make sure, but no luck - https://github.com/erszcz/Gradualizer/runs/4056122650?check_suite_focus=true. The only info about this I could find is http://erlang.org/pipermail/erlang-questions/2011-June/059722.html, but it's a dead end.

src/typechecker.erl
erszcz: Fixed :+1:

erszcz:
I applied fixes for the latest comments. I think this is ready to merge unless we need CI on Erlang 22. I also modified the PR title to make it clear it doesn't yet detect any infinite loops - that needs some further work.

.github/workflows/build-and-test.yml
zuiderkwast: I don't know if we need CI on OTP 22. Maybe it's better to have it if it's easy? Is it hard to just exclude the property tests on the OTP 22 build?

erszcz: Ok, it seems we can quite easily skip property tests on 22, but have that version in the matrix - done in the latest commit.

.github/workflows/build-and-test.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       - name: "Run tests"
         run: "make tests"
       - name: "Run property-based tests"
+        if: ${{ matrix.otp != '22.3.4.20' }}
```
zuiderkwast, line 27: :+1:

src/gradualizer_bin.erl
```diff
@@ -65,11 +72,13 @@ bin_element_view({bin_element, _Anno, _Expr, default, Specifiers}) ->
...
+bin_element_view({bin_element, Anno, _Expr, SizeSpec, Specifiers} = BinElem) ->
     %% Non-default size, possibly a constant expression
-    try erl_eval:expr(SizeSpec, []) of
-        {value, Sz, _VarBinds} ->
-            {Sz * get_unit(Specifiers), 0}
+    try
+        case erl_eval:expr(SizeSpec, []) of
+            {value, Sz, _VarBinds} when is_integer(Sz) ->
+                {Sz * get_unit(Specifiers), 0}
+        end
```
zuiderkwast, line 81: The exception is in `erl_eval:expr` right? If yes, please revert this unnecessary change. Then I think it's ready to merge.

zuiderkwast, line 81:
```suggestion
    try erl_eval:expr(SizeSpec, []) of
        {value, Sz, _VarBinds} when is_integer(Sz) ->
            {Sz * get_unit(Specifiers), 0}
```

erszcz: The expression might be in `erl_eval:expr` or a `case_clause`.

This change is intentional, please see the commit message for https://github.com/josefs/Gradualizer/pull/363/commits/8db6300fe62dcf5865b0163c8731715c0ccb3f60. From the official docs:

> In construction, `Size` is any expression that evaluates to an integer.

The point is that the `SizeSpec` must evaluate to int, so we use this syntax to ensure that `is_integer(Sz)` holds - if not, we handle the `case_clause`, too.

zuiderkwast: Ah, ok, makes sense.

The case with only one clause can be written as `=` then. Not really important though.

    try
        {value, Sz, _VarBinds} = erl_eval:expr(SizeSpec, []),
        true = is_integer(Sz),
        {Sz * get_unit(Specifiers), 0}

zuiderkwast:
@erszcz Do you have a message for a squash commit? Can I use the top comment of the PR?

src/typechecker.erl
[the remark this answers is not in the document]
zuiderkwast: Good. In the past I wanted the `ann_type` annotations to be deleted by normalize. I don't know why we didn't do it. It's good that at least normalizing the child of the `ann_type` solves it.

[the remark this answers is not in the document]
zuiderkwast: See #228, #247, #253. Specifically #253 deleted the `ann_type` clause in normalize and many other places and instead strips the `ann_type`s in `remove_pos`. Do you know how an ann_type can make it past remove_pos to normalize?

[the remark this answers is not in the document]
erszcz: Ahhh, that explains a lot!

There are multiple properties defined and most of them call internal typechecker functions. The properties should encapsulate the information on invariants which should hold for the inputs to these internal typechecker functions (like, "should params to this function be normalized already or will it normalize them?"). Possibly, some don't do it accurately, yet.

That being said, the problem was originally found by `prop_glb`. Let's see if it shows up if the property in question is modified to call `typelib:remove_pos/1` on the arguments before passing them to `glb/3`...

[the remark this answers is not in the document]
erszcz: Eagle eye, @zuiderkwast! https://github.com/josefs/Gradualizer/pull/363/commits/d72d57e7ea1c72bcbc1fefd76cb87998bdb95248 illustrates what I had in mind in the previous comment.

BTW, if I understand you correctly, then every property that uses a generated type as input should actually call `typelib:remove_pos` on the type, shouldn't it?

test/property_test/gradualizer_prop.erl
```diff
@@ -69,8 +69,8 @@ prop_glb() ->
 
 prop_glb_(Type1, Type2) ->
     Env = test_lib:create_env([]),
-    Type1_ = typechecker:normalize(Type1, Env),
-    Type2_ = typechecker:normalize(Type2, Env),
+    Type1_ = typelib:remove_pos(typechecker:normalize(Type1, Env)),
+    Type2_ = typelib:remove_pos(typechecker:normalize(Type2, Env)),
```
zuiderkwast, line 73: The other way around. Remove pos before normalize.

erszcz:
I've modified the properties to always call `typelib:remove_pos` on generated types.

As for the squash message, maybe this?

> This PR sketches a property based testing framework for Gradualizer. It's based on a vendored version of Proper's Erlang abstract code generator used to property test the Erlang compiler.

zuiderkwast:
Great!

PR #361, opened 2021-10-20
Stop more cases of infinite recursion

erszcz:
This builds on top of https://github.com/josefs/Gradualizer/pull/356 and https://github.com/josefs/Gradualizer/pull/324 and breaks all cycles summed up in https://github.com/josefs/Gradualizer/pull/324#issuecomment-944074679. It also terminates on the original example posted in https://github.com/josefs/Gradualizer/issues/322. Together with the fixes from https://github.com/erszcz/Gradualizer/tree/start-app-in-rebar-plugin it also terminates on the example project from https://github.com/josefs/Gradualizer/issues/360.

Apparently, this breaks a few existing tests, so I'm leaving it as a draft for now.

Issue #360, opened 2021-10-20
function_clause error in typechecker:unary_op_arg_type/2 with --use_beams


jesperes:
Trying to run gradualizer on my "Advent of Code" solutions (https://github.com/jesperes/aoc_erlang) triggered a `function_clause` error when running `rebar3 gradualizer --use_beams`:

```erlang
===> Uncaught error: function_clause
===> Stack trace to the error location:
[{typechecker,unary_op_arg_type,
              ['-',{var,0,"_TyVar-576460752303423391"}],
              [{file,"/home/jespereskilson/dev/aoc_erlang/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,2735}]},
 {typechecker,type_check_unary_op_in,5,
              [{file,"/home/jespereskilson/dev/aoc_erlang/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,2728}]},
 {typechecker,type_check_expr_in,3,
              [{file,"/home/jespereskilson/dev/aoc_erlang/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,2051}]},
 {typechecker,'-type_check_call_ty/4-lc$^0/1-0-',2,
              [{file,"/home/jespereskilson/dev/aoc_erlang/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,1864}]},
 {typechecker,'-type_check_call_ty/4-lc$^0/1-0-',2,
              [{file,"/home/jespereskilson/dev/aoc_erlang/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,1865}]},
 {typechecker,type_check_call_ty,4,
              [{file,"/home/jespereskilson/dev/aoc_erlang/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,1865}]},
 {typechecker,do_type_check_expr,2,
              [{file,"/home/jespereskilson/dev/aoc_erlang/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,1434}]},
 {typechecker,type_check_expr,2,
              [{file,"/home/jespereskilson/dev/aoc_erlang/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,1311}]}]
```

To reproduce:

* Clone https://github.com/jesperes/aoc_erlang/commit/573a16409e9cf2b29ccb5fdfb809c2da407d845c
* Do `rebar3 gradualizer --use_beams`

[...]

PR #368, opened 2021-11-07
Property based tests with recursive types

erszcz:
This continues the effort to find infinite loops with property based tests. Moreover, it subsumes #356 and #361, i.e. the branches those two PRs refer to are already merged into this PR. Proptests allowed to find more infinite loops than the ones we were already aware of and had examples to trigger.

However, this PR does not solve all the problems found ~and therefore CI will most likely fail~ EDIT: I split the workflows, so it's clear that proptests fail, but ordinary tests do not. I'll send another draft PR with some more research and a description of the unsolved issues. That being said, the chunk of work in this PR seems reasonably sound, so I'm posting it for review and eventual merge.

This should finally enable merging https://github.com/josefs/Gradualizer/pull/359 and also not looping on the example project from https://github.com/josefs/Gradualizer/issues/360.


erszcz:
I've cleaned up this PR and prepared for review and merge.

I closed the two PRs which are already included here. Property based tests allow for more thorough testing of infinite loops than was done with unit tests in those PRs (the relevant unit tests are still there).

As can be seen in the test results of the last commits, the number of property repetitions is crucial - with 100 repetitions there's 1 failing property and 11 passing, with 5000 repetitions it's 4 failing, 8 passing. The failures are further explored in #369.

That being said, I think these changes make Gradualizer significantly more robust than the master branch, therefore I think it's good to bring them in.

erszcz:
@zuiderkwast @josefs Do you think you could find some time for a review? I consider this PR ready, unless you have comments on what to change / fix.

zuiderkwast:
Oh, but the build failed. I thought you wanted to fix that. Don't we want the builds to be green?

erszcz:
Maybe I wasn't clear on that, sorry. I don't think it's that easy to fix property based tests in this PR - https://github.com/josefs/Gradualizer/pull/369 elaborates on why that's the case.

That being said, I think this PR is a good milestone to merge into master anyway, since it subsumes all the already known code examples which cause Gradualizer to loop indefinitely and more discovered by proptests.

I could make proptests light up green even if they fail, but I'm not sure it's the right thing to do. If we see them failing, but ordinary tests passing, it's like a "known problem" to solve - which is the approach already used in Gradualizer tests anyway.

zuiderkwast:
I think it's better to allow proptests to fail for now.

The known problems are similar. They don't fail our builds.

josefs:
I agree with Victor. We can let the prop tests fail for now. Another option is to divide them up in to passing and failing tests as we have to unit tests.

erszcz:
> We can let the prop tests fail for now.

So that's how it's right now - all "Build and test" CI jobs pass, whereas "Property based tests" do not. Or did you have "light up green even if they fail" in mind?

zuiderkwast:
Yes, make it green

zuiderkwast:
What matters is the overall colour/status so we can easily spot whenever something is broken IMO

erszcz:
Done. The property tests fail, but they don't fail the entire PR.

.github/workflows/proptests.yml
```diff
@@ -0,0 +1,24 @@
...
+      matrix:
+        otp: ['23.3.4.7', '24.1']
+    steps:
+      - uses: actions/checkout@v2
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: ${{matrix.otp}}
+      - name: "Run property-based tests"
+        run: "PROP_NUMTESTS=5000 rebar3 ct --suite test/gradualizer_prop_SUITE.erl"
+        continue-on-error: true
```
zuiderkwast, line 24: :trophy:

zuiderkwast (approved):

PR #369, opened 2021-11-07
Recursive types - unsolved problems

erszcz:
This is the description of further infinite loop / recursive type problem I referred to in #368.

https://github.com/josefs/Gradualizer/commit/665d3f1b6a27f1d88047ca3e0f3ea693248d4ce0 contains the description of the problem together with an example (a type generated automatically for proptests). I'll copy the description here for convenience.

When a union type is normalized, the topmost occurences of user types in its definition are expanded. If the user types are recursive and are unions themselves, a subsequent normalization leads to their further expansion. This means, that normalize^1(Ty) is NOT equal to normalize^2(Ty), where f^2(x) := f(f(x)). In other words, subsequent normalizations of such a type make the type grow indefinitely.

We have to keep this behaviour in mind when doing recursive computations on types, to avoid infinite loops when trying to process the ever-growing type. I tried acting accordingly in `glb` and `compatible`.

In `glb` it initially led to a few broken test cases, but I managed to fix them by "manually" controlling when a user type should be expanded by switching a flag in Env which says "from now on, do not expand user types" - so that until we return from `glb` expansion doesn't happen.

The same approach doesn't work in `compatible`, though, as for example `test/should_pass/remote_types.erl` stops typechecking due to `user_types:my_tuple()` not getting expanded:

```
test/should_pass/remote_types.erl: The variable on line 9 at column 9 is expected to have type user_types:my_tuple() but it has type {atom(), integer()}

%% test resolving remote tuple referencing a user type in the remote module
-spec f(local_tuple()) -> user_types:my_tuple().
f(A) -> A.
        ^
```

I'm not sure how to proceed yet. As it is now, the property tests will not repeatedly pass (they sometimes might) due to the above problem.


erszcz:
I've skimmed over the _Recursive Types_ chapter of _Types and Programming Languages_ by Pierce. Here's what I think:

1. `normalize`, to work as expected, cannot return a different value when called on an already normalised value.
1. This means it cannot expand a self-referencing recursive type further. Before the infinite-loop fix it would expand indefinitely, i.e. `normalize` would not terminate. Now it expands only "one step ahead", but it still treats more and less expanded representations of a recursive type as distinct types. So `normalize` doesn't loop indefinitely anymore, but any operation which calls `normalize` in its subsequent recursive calls (like `glb`) does loop indefinitely.
1. [Recursive types in TaPL are explicitly modelled as such](https://github.com/erszcz/tapl-erlang/blob/b364c24416533826daaa3bd5970e5560ffa8838b/equirec/syntax.ml#L12) also [in the concrete syntax](https://github.com/erszcz/tapl-erlang/blob/b364c24416533826daaa3bd5970e5560ffa8838b/equirec/test.f#L6) - it makes it easier, or feasible at all, to handle them.
1. The only types which might be really recursive are user types. `normalize` is used to both expand them to their definition, which opens the path for infinite recursion, and to simplify them to some canonical form. If we just returned them as is, there'd be no risk of cycling, but it would make it impossible to compare them structurally, i.e. compute `glb` or the subtype relation.
1. Expansion of builtin types also causes problems sometimes, when fully expanded, and not fully expanded values are compared. This might happen if a user type is normalised the first time (expanded to definition) early and then definition is expanded again deeper in the type checker.

Especially the 4th point is interesting, since when applied to Erlang it would mean a `-type` attribute defining a recursive type would use a keyword to mark the type as recursive explicitly. Of course, the `-type` syntax is already defined, so using an additional keyword is not possible. Instead, we could detect if a type definition is recursive when encountering it the first time. This would be straightforward for _simple recursive types_, i.e. self-referencing ones:

```erlang
-type int_list(T) :: nil | {cons, integer(), int_list()}.
```

 It might be hard for mutually recursive types:

```erlang
-type t1() :: t1 | t2().
-type t2() :: t2 | t1().
```

Unfortunately, TaPL does not provide an example typechecker with recursive types and subtyping - both features are demonstrated, but in separate examples - and that's exactly what we're trying to achieve in Gradualizer.

I think we might be able to fix `normalize` for real by representing recursive types explicitly as such, that is by extending [`abstract_type()`](https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/gradualizer_type.erl#L201) and handling appropriately.

zuiderkwast:
Just want to explain the reasoning from when we implemented normalize.

> 1. `normalize`, to work as expected, cannot return a different value when called on an already normalised value.

Agree. Normalize should be stable.

> 2. This means it cannot expand a self-referencing recursive type further. Before the infinite-loop fix it would expand indefinitely, i.e. `normalize` would not terminate. Now it expands only "one step ahead", but it still treats more and less expanded representations of a recursive type as distinct types. So `normalize` doesn't loop indefinitely anymore, but any operation which calls `normalize` in its subsequent recursive calls (like `glb`) does loop indefinitely.

That's right. That's why we only normalize on the top level. This is called Head Normal Form.

Any function which does deep normalization (like glb) must detect recursion and stop. The way it's done so far is to pass along a set of types already seen, as I'm sure you have seen.

> 4. The only types which might be really recursive are user types. `normalize` is used to both expand them to their definition, which opens the path for infinite recursion, and to simplify them to some canonical form. If we just returned them as is, there'd be no risk of cycling, but it would make it impossible to compare them structurally, i.e. compute `glb` or the subtype relation.

Normalize only operates on the top-level union and we forbid recursion on the top-level union, so a type like `-type t() :: 1 | t().` is not allowed (although we might not detect this; not sure). Recursion on the top-level union is also quite meaningless.

> 5. Expansion of builtin types also causes problems sometimes, when fully expanded, and not fully expanded values are compared. This might happen if a user type is normalized the first time (expanded to definition) early and then definition is expanded again deeper in the type checker.

Do you have an example in which this turns into infinite recursion?

> ```erlang
> -type t1() :: t1 | t2().
> -type t2() :: t2 | t1().
> ```

This is recursion on top-level, which we don't support. An improvement would be to detect it in normalize and report an error.

erszcz:
@zuiderkwast Thanks for the explanations, they're very insightful.

> That's right. That's why we only normalize on the top level. This is called Head Normal Form.

I think that's not the case, though, unless I misunderstand "on the top level". [`normalize` is called subsequently in recursive calls of, for example, `glb`](https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L539). This means that part of an already normalised type is normalised again when we recurse down in `glb`. That's one of the cases when we still run into an infinite loop on recursive types, even with the preventive measures in both `normalize` and `glb`, because currently `normalize` is not stable.

> Normalize only operates on the top-level union and we forbid recursion on the top-level union, so a type like `-type t() :: 1 | t().` is not allowed (although we might not detect this; not sure).
> [...] This is recursion on top-level, which we don't support. An improvement would be to detect it in normalize and report an error.

Hmm, I was not aware of that. I've not seen such a check in code. This is an important next step, then.

> Do you have an example in which this turns into infinite recursion?

I think the problem here is not infinite recursion - ultimately builtin type expansion is well defined and ends at some point. However, I remember seeing a buggy check result, which when traced was caused by a mismatch between a `string()` and a `[char()]` - or was it `[char()]` vs `[0..16#10ffff]`? - the point is it was a mismatch between a more and less expanded builtin type. I'll find an exact example.

zuiderkwast:
> I think that's not the case, though, unless I misunderstand "on the top level". [`normalize` is called subsequently in recursive calls of, for example, `glb`](https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L539). This means that part of an already normalised type is normalised again when we recurse down in `glb`.

I mean that `normalize` only works on the top level of its argument. If it's called by glb on other parts of a type, then that's a problem of glb, I think. Glb needs to reember what it already handled.

Is the problem is about type variables?

In which case is normalize not stable?

erszcz:
@zuiderkwast Before I dive in the code again, could you specify recursion at the top level a bit more? For example, given these types:

1. `-type t() :: 1 | t().`
2. `-type t() :: 1 | {t()}.`

Are both recursive at the top level or only the first one?

zuiderkwast:
> @zuiderkwast Before I dive in the code again, could you specify recursion at the top level a bit more? For example, given these types:
> 
>     1. `-type t() :: 1 | t().`
> 
>     2. `-type t() :: 1 | {t()}.`
> 
> 
> Are both recursive at the top level or only the first one?

Only the first one. Union is the only operation we accept on toplevel. Tuple and list are sublevel stuff.

erszcz:
Closing as this was a dead end anyway and the master branch has diverged a lot.

PR #372, opened 2021-12-17
Use erl_lint to detect compile errors

### LINKED ISSUE #288: Should not pass on malformed expression.

### LINKED ISSUE #291: Missing included path and function signature being present or not.

### LINKED ISSUE #371: internal crash on comparison of unbound variable to non-contractive type
Given this code:

**Steps to repro**:

`./bin/gradualizer sample.erl`

where `sample.erl` contains:

```erl
-module(sample).

-type loop() :: loop().

-spec main(any()) -> loop().
main(ok) -> X.
```

**Actual behavior**

crash:

```
escript: exception error: bad key: 'X'
  in function  map_get/2
     called as map_get('X',#{})
     *** argument 1: not present in map
  in call from typechecker:do_type_check_expr_in/3 (src/typechecker.erl, line 2248)
  in call from typechecker:type_check_expr_in/3 (src/typechecker.erl, line 2241)
  in call from typechecker:check_clause/5 (src/typechecker.erl, line 3443)
  in call from typechecker:'-check_clauses/5-fun-0-'/5 (src/typechecker.erl, line 3400)
  in call from lists:foldl/3 (lists.erl, line 1285)
  in call from typechecker:check_clauses/5 (src/typechecker.erl, line 3398)
  in call from typechecker:check_clauses_fun/3 (src/typechecker.erl, line 3372)
  ```

**Expected behavior**

The same as for the example above with `-type loop() :: any()`:

```
sample.erl: Unknown variable 'X' on line 6 at column 13.
```

I doubt this kind of thing comes up much in the wild, I found it playing around with intentionally weird code using recursive types.

### PR DESCRIPTION
**Pull request by zuiderkwast**
When checking `.erl` files, use `erl_lint` to detect compile errors. This is basically what the `basic_validation` compiler pass does.

It detects use of undefined variables, undefined functions, undefined records, missing include files and some things you could expect the parser to detect such as `foo() -> ok; bar() ->ok.` (Note the semicolon.) and `f(X + 1) -> X.`.

This pass is not used when checking beam files, since we already know that this abstract form compiles.

Fixes #291. Fixes #288. Fixes #371.

### DIFF
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -56,21 +56,47 @@ type_check_file(File) ->
 %% @doc Type check a source or beam file
 -spec type_check_file(file:filename(), options()) -> ok | nok | [{file:filename(), any()}].
 type_check_file(File, Opts) ->
-    ParsedFile =
-        case filename:extension(File) of
-            ".erl" ->
-                Includes = proplists:get_all_values(i, Opts),
-                gradualizer_file_utils:get_forms_from_erl(File, Includes);
-            ".beam" ->
-                gradualizer_file_utils:get_forms_from_beam(File);
-            Ext ->
-                throw({unknown_file_extension, Ext})
-        end,
-    case ParsedFile of
-        {ok, Forms} ->
+    case filename:extension(File) of
+        ".erl" ->
+            Includes = proplists:get_all_values(i, Opts),
+            case gradualizer_file_utils:get_forms_from_erl(File, Includes) of
+                {ok, Forms} ->
+                    lint_and_check_forms(Forms, File, Opts);
+                Error ->
+                    throw(Error)
+            end;
+        ".beam" ->
+            case gradualizer_file_utils:get_forms_from_beam(File) of
+                {ok, Forms} ->
+                    type_check_forms(File, Forms, Opts);
+                Error ->
+                    throw(Error)
+            end;
+        Ext ->
+            throw({unknown_file_extension, Ext})
+    end.
+
+%% @doc Runs an erl_lint pass, to check if the forms can be compiled at all,
+%% before running the type checker.
+-spec lint_and_check_forms([erl_parse:abstract_form()], file:filename(), options()) ->
+          ok | nok | [{file:filename(), any()}].
+lint_and_check_forms(Forms, File, Opts) ->
+    case erl_lint:module(Forms, File, [return_errors]) of
+        {ok, _Warnings} ->
             type_check_forms(File, Forms, Opts);
-        Error ->
-            throw(Error)
+        {error, Errors, _Warnings} ->
+            %% If there are lint errors (i.e. compile errors like undefined
+            %% variables) we don't even try to type check.
+            case proplists:get_bool(return_errors, Opts) of
+                true ->
+                    [{Filename, ErrorInfo} || {Filename, ErrorInfos} <- Errors,
+                                              ErrorInfo <- ErrorInfos];
+                false ->
+                    [gradualizer_fmt:print_errors(ErrorInfos,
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by zuiderkwast** [comment]
For reviewers: Tests for the lint errors are moved to a separate module, because if there are lint errors, we don't run the type checker and thus we don't detect non-lint errors in the same module. With one module for lint errors (e.g. undefined local type) and another for type errors (e.g. undefined remote type) we cover all of them.

File: src/gradualizer.erl [hunk 1]
```diff
@@ -56,21 +56,47 @@ type_check_file(File) ->
...
+            end;
+        ".beam" ->
+            case gradualizer_file_utils:get_forms_from_beam(File) of
+                {ok, Forms} ->
+                    type_check_forms(File, Forms, Opts);
+                Error ->
+                    throw(Error) % TODO: Don't use throw. These are user errors.
+            end;
+        Ext ->
+            throw({unknown_file_extension, Ext}) %% TODO: Don't use throw here.
```
josefs, line 76: Can you create a ticket and add the ticket number to the TODOs so that we keep track of that we need to remove the throws in the future?

josefs (approved):
Thanks. I like changes that remove code. 
The thing that makes me a little uneasy about this is the performance implications of this change. We're adding a whole extra traversal over the AST. But on the other hand, we haven't really cared much about the performance of Gradualizer yet and I haven't gotten any feedback that it's slow.

test/misc/lint_errors.erl
```diff
@@ -0,0 +1,43 @@
+%% This module doesn't even compile. There errors here are caught by erl_lint.
```
erszcz, line 1:
```suggestion
%% This module doesn't even compile. The errors here are caught by erl_lint.
```

erszcz (approved):
Looks good to me!

src/gradualizer.erl
zuiderkwast: Too much effort to describe it, including context and consequences, in an issue. I deleted the TODOs instead.

zuiderkwast:
> We're adding a whole extra traversal over the AST. But on the other hand, we haven't really cared much about the performance of Gradualizer yet and I haven't gotten any feedback that it's slow.

We have had performance issues and even infinite loops, so we have added hacks to limit the depth of certain things.

I think this pass is very quick (linear maybe), besides it can be regarded as parser post-processing and we do use the lexer, preprocessor and parser passes already.

Good that it's accepted. Now we closed a number of issues. :-)

PR #375, opened 2022-02-01
Call remove_pos only where a type comes into the system

zuiderkwast:
Remove_pos normalizes the syntactical representation of a type, by recursively removing the location annotation (line and column). This is to be able to compare types using equality and pattern matching in the type checker.

This should only be done once on each type and only when the type comes into the type checker, i.e. when types are extracted from a parse tree.

This PR sorts out the mess, makes sure it's done where it should and removes it where it shouldn't be done.

Note that gradualizer_db doesn't use this, so we call remove_pos on the types returned by gradualizer_db.

### DIFF
--- a/src/gradualizer_db.erl
+++ b/src/gradualizer_db.erl
@@ -491,7 +491,7 @@ collect_types(Module, Forms) ->
                  Info     = #typeinfo{exported = Exported,
                                       opaque   = (Attr == opaque),
                                       params   = Params,
-                                      body     = Body},
+                                      body     = typelib:remove_pos(Body)},
                  {Id, Info}
              end || Form = {attribute, _, Attr, {Name, Body, Vars}} <- Forms,
                     Attr == type orelse Attr == opaque,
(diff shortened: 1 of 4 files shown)


erszcz:
This looks good, but I think there are a few places where `remove_pos` is now redundant:
- https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L772
- possibly https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L3060
- possibly https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L3074

Apart from the above, there are a few cases where it's called on a value returned from various `gradualizer_db` functions - for example:
https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L785

Is there a reason to keep types with their position in the DB or could we also remove it there, so remove the redundancy at call sites?

zuiderkwast:
> I think there are a few places where `remove_pos` is now redundant

Thanks, the tests still pass when I remove these.

> Is there a reason to keep types with their position in the DB or could we also remove it there, to remove the redundancy at call sites?

The reason I didn't do it already is that this is no mess so far. Remove_pos was simply not done there, without exceptions. The mess was only in typechecker. But since gradualizer_db does other stuff such as `absform:normalize_function_type_list/1` and `absform:normalize_record_field/1`, I think it can do remove_pos too. It will make everything simpler. :+1: I'll do it tomorrow.

erszcz (approved):
That's neat!

PR #376, opened 2022-02-01
Don't let normalize recurse into element types

zuiderkwast:
... or "Make normalize head normal form again"

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -735,6 +735,10 @@ has_overlapping_keys({type, _, map, Assocs}, Env) ->
 %% * Expand user-defined and remote types on head level (except opaque types)
 %% * Replace built-in type synonyms
 %% * Flatten unions and merge overlapping types (e.g. ranges) in unions
+%%
+%% This function only normalizes the head level type or union of types to
+%% compute the Head Normal Form. In particular, it doesn't normalize the element
+%% types of a list, tuple, map or record type.
 -spec normalize(type(), tenv()) -> type().
 normalize(Ty, Env) ->
     normalize_rec(Ty, Env, #{}).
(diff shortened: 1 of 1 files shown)


erszcz:
Hmm, I don't think it's as simple as that. What about
https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L799-L800
https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L802
https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L805 to name a few? I think all of these will lead to broken test cases, though I haven't checked.

zuiderkwast:
You're right, I've lost it. :sob: I'll have to study this source code from scratch.

Anyway, Head Normal Form is about not recursing fully down the tree. I'm pretty sure it was like that once upon a time, but I don't know since when and why it's not the case anymore.

erszcz:
> https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L802 
> 
> https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L805

Apparently, the above can be removed with no tests breaking. It's not that nice with this one, though:

> https://github.com/josefs/Gradualizer/blob/0657d76e7de893a2bd7e49e0fac186b732fea434/src/typechecker.erl#L799-L800

Quite a lot started failing when I tried that :/

erszcz:
I think this can be closed now, WDYT @zuiderkwast?

Issue #379, opened 2022-02-08
`remote_type` position handling is incorrect


erszcz:
https://github.com/esl/gradient/issues/31 shows that removing position information leads to `undef, remote_type` being thrown with the position always being equal to 0. This results in warnings like this:

```
Undefined remote type gradualizer_type:abstract_type/0  # when self-gradualizing in the editor
lib/tip_calculator.ex: Undefined remote type Access.get/2 on line 0  # from the issue linked above
```

Having traced the problem I see two things:

```
{trace,<0.93.0>,call,
 {typechecker,normalize_rec,
  [{remote_type,0,
    [{atom,3,'Elixir.Access'},
     {atom,3,get},
     [{atom,0,'Elixir.List'},
      {remote_type,0,[{atom,3,'Elixir.String'},{atom,3,t},[]]}]]},
  ...
```

1. We could use the position from module name or type name (`3` above), which would solve the problem of incorrect line in the warning, BUT...
2. The above means that `typelib:remove_pos` for `remote_type` is actually buggy, since it doesn't remove the position everywhere, so remote type references coming from different source lines do not compare equal :/ I'm not sure if it's a real problem, given we don't see any negative consequences.
3. If we make the `remove_pos` clause for `remote_type` just return the argument type (i.e. not remove the position), we get 11 test failures with a lot of `position_not_removed` errors thrown.

What could we do with this?

It seems that in order to easily compare types we need this information removed, but in order to provide this particular diagnostic we need the position present.

I imagine we could do a first pass before `type_check_forms` to just check for remote types being defined - in this pass the position would not have to be removed. However, that's costly as the entire module would have to be scanned one more time (apart from `erl_lint` and `type_check_forms`).

zuiderkwast:
The location we're reporting in errors is where the error happens in the module being checked. It can be within an expression but in this case it is within a -type or a -spec.

Can we have a separate pass where we check for undefined remote types, user types and records, before we call remove_pos?

PR #380, opened 2022-02-08
Detect undefined remote types in a separate pass

erszcz:
This tries fixing #379 by introducing a separate pass over the AST.

Remaining tasks:
- [x] fix `typelib:remove_pos` to actually remove pos from all `remote_type` elements
- [x] fix the failing test case

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4834,8 +4834,8 @@ check_remote_type({call, _, {remote, _, {atom, _, Module}, {atom, _, Fun}}, Args
     end;
 check_remote_type({remote_type, P, [{atom, _, M}, {atom, _, N}, Args]}, Acc) ->
     case gradualizer_db:get_exported_type(M, N, Args) of
-        {ok, _} ->
-            Acc;
+        {ok, Type} ->
+            check_undefined_types([Type]) ++ Acc;
         opaque ->
             Acc;
         not_exported ->
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4775,6 +4775,9 @@ type_check_forms(Forms, Opts) ->
               (_Function, Errors) ->
                   Errors
           end, [], ParseData#parsedata.functions),
+    %% `TypeCheckErrors' will contain undefined remote type errors missing position information.
+    %% We don't want to report them, as they're not useful to the user - we report
+    %% `UndefinedTypeErrors' instead.
     AllErrors = filter_undefined_type_errors(Env, TypeCheckErrors) ++ UndefinedTypeErrors,
     lists:reverse(AllErrors).
 
@@ -4816,6 +4819,26 @@ create_fenv(Specs, Funs) ->
       ]
      ).
 
+%% @doc Check for undefined remote types in a separate pass over `Forms'.
+%%
+%% The main typechecker pass defined in `type_check_forms/2' removes position information from types
+%% as soon as they're encountered. It's done to make types occuring on different source
+%% lines represented the same, therefore easy to compare.
+%%
+%% However, in case of references to undefined types (that is, programmer mistakes) it means that
+%% an error which ought to carry this position information to point the user at the source line
+%% to fix is missing.
+%%
+%% This pass over the AST does not remove position information, so invalid remote type references
+%% can be reported properly.
+%%
+%% See https://github.com/josefs/Gradualizer/issues/379 for more information.
+%% @end
+-spec check_undefined_types(Forms) -> Errors when
+      Forms :: gradualizer_file_utils:abstract_forms(),
+      Errors :: [ Error ],
+      Error :: {not_exported, remote_type, erl_anno:anno(), {module(), atom(), non_neg_integer()}}
+             | {undef, remote_type, erl_anno:anno(), {module(), atom(), non_neg_integer()}}.
 check_undefined_types(Forms) ->
     gradualizer_lib:fold_ast(fun check_remote_type/2, [], Forms).
 
@@ -4825,6 +4848,7 @@ check_remote_type({attribute, _, SpecType, {_, Forms}}, Acc)
     check_undefined_types(Forms) ++ Acc;
 check_remote_type(?top(), Acc) ->
     Acc;
+%% Check remote calls, as the remote function specs can also refer to undefined remote types.
 check_remote_type({call, _, {remote, _, {atom, _, Module}, {atom, _, Fun}}, Args}, Acc) ->
     case gradualizer_db:get_spec(Module, Fun, length(Args)) of
         {ok, Types} ->
@@ -4835,6 +4859,7 @@ check_remote_type({call, _, {remote, _, {atom, _, Module}, {atom, _, Fun}}, Args
 check_remote_type({remote_type, P, [{atom, _, M}, {atom, _, N}, Args]}, Acc) ->
     case gradualizer_db:get_exported_type(M, N, Args) of
         {ok, Type} ->
+            %% A remote type might expand to another remote type, so let's check that, too.
             check_undefined_types([Type]) ++ Acc;
         opaque ->
             Acc;
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -766,7 +766,7 @@ normalize_rec({user_type, P, Name, Args} = Type, Env, Unfolded) ->
         {type, NormType} -> NormType;
         no_type ->
             UnfoldedNew = maps:put(mta(Type, Env), {type, Type}, Unfolded),
-            case gradualizer_lib:get_type_definition(Type, Env, []) of
+            case gradualizer_lib:get_type_definition(Type, Env, [remove_pos]) of
                 {ok, T} ->
                     normalize_rec(T, Env, UnfoldedNew);
                 opaque ->
@@ -779,7 +779,7 @@ normalize_rec(T = ?top(), _Env, _Unfolded) ->
     %% Don't normalize gradualizer:top().
     T;
 normalize_rec({remote_type, P, [{atom, _, M}, {atom, _, N}, Args]}, Env, Unfolded) ->
-    case gradualizer_db:get_exported_type(M, N, Args) of
+    case gradualizer_db:get_exported_type(M, N, Args, [remove_pos]) of
         {ok, T} ->
             normalize_rec(T, Env, Unfolded);
         opaque ->
@@ -1724,7 +1724,7 @@ do_type_check_expr(Env, {'fun', P, {function, Name, Arity}}) ->
 do_type_check_expr(Env, {'fun', P, {function, M, F, A}}) ->
     case {get_atom(Env, M), get_atom(Env, F), A} of
         {{atom, _, Module}, {atom, _, Function}, {integer, _, Arity}} ->
-            case gradualizer_db:get_spec(Module, Function, Arity) of
+            case gradualizer_db:get_spec(Module, Function, Arity, [remove_pos]) of
                 {ok, BoundedFunTypeList} ->
                     Ty = bounded_type_list_to_type(Env, BoundedFunTypeList),
                     {Ty, #{}, constraints:empty()};
@@ -2526,7 +2526,7 @@ do_type_check_expr_in(Env, ResTy, Expr = {'fun', P, {function, Name, Arity}}) ->
 do_type_check_expr_in(Env, ResTy, Expr = {'fun', P, {function, M, F, A}}) ->
     case {get_atom(Env, M), get_atom(Env, F), A} of
         {{atom, _, Module}, {atom, _, Function}, {integer, _,Arity}} ->
-            case gradualizer_db:get_spec(Module, Function, Arity) of
+            case gradualizer_db:get_spec(Module, Function, Arity, [remove_pos]) of
                 {ok, BoundedFunTypeList} ->
                     FunTypeList =
                         unfold_bounded_type_list(Env, BoundedFunTypeList),
@@ -3090,7 +3090,7 @@ type_check_fun(Env, {atom, P, Name}, Arity) ->
     {Types, #{}, constraints:empty()};
 type_check_fun(_Env, {remote, P, {atom,_,Module}, {atom,_,Fun}}, Arity) ->
     % Module:function call
-    case gradualizer_db:get_spec(Module, Fun, Arity) of
+    case gradualizer_db:get_spec(Module, Fun, Arity, [remove_pos]) of
         {ok, Types} -> {Types, #{}, constraints:empty()};
         not_found   -> throw({call_undef, P, Module, Fun, Arity})
     end;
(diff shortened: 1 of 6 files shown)


zuiderkwast (commented):
Seems you like the idea of a separate pass. :-)

I think it is slightly more complex than this though. What about remote types that expand to other remote types? There may even be circular dependencies between remote types.

E.g. [here](https://github.com/josefs/Gradualizer/pull/380/files#diff-2a1f43e88f03f8c1f907ab604658c19b600eec97d4d3ac72682410319beb8890R771) is a call to `gradualizer_lib:get_type_definition`, which expands types and calls remove pos internally.

Instead of a completely separate pass, how about checking for undefined remote types (and undefined user types and records) just after importing a type or spec from gradualizer_lib and just before removing the pos from these types? That code is currently inside gradualizer_lib:get_type_definition but we can split it out again if you prefer.

gradualizer_lib has already expanded the type and called remove_pos. Whenever we encounter a remote type

> * [ ] fix `typelib:remove_pos` to actually remove pos from all remote_type elements

Doesn't it already? I.e. it removes the line and column, but it keeps the filename annotation in user types and records, which is good because it these are different in different modules.

erszcz:
> Seems you like the idea of a separate pass. :-)

Well, it was relatively easy to pull off, at the very least to check if it solves the problem :)

> I think it is slightly more complex than this though. What about remote types that expand to other remote types?

Good point!

> Instead of a completely separate pass, how about [...]

I was thinking of incorporating this check in the existing pass, but it wasn't clear to me where we could do it. I'll try your suggestion out.

> > * [ ]  fix `typelib:remove_pos` to actually remove pos from all remote_type elements
> 
> Doesn't it already?

It doesn't remove pos from `Mod` and `Name`: https://github.com/josefs/Gradualizer/blob/b21659436a320253ce36763f3d326b5af7c598d7/src/typelib.erl#L119-L121

which do have it according to the trace from #379:

```
{trace,<0.93.0>,call,
 {typechecker,normalize_rec,
  [{remote_type,0,
    [{atom,3,'Elixir.Access'},
     {atom,3,get},
     [{atom,0,'Elixir.List'},
      {remote_type,0,[{atom,3,'Elixir.String'},{atom,3,t},[]]}]]},
...
```

zuiderkwast:
> It doesn't remove pos from `Mod` and `Name`:

Good point! It should, so that two identical remote types will compare as identical.

erszcz:
@zuiderkwast

> I think it is slightly more complex than this though. What about remote types that expand to other remote types?

It seems to me this case is covered by tests in [`undefined_errors_test.erl`](https://github.com/josefs/Gradualizer/blob/b21659436a320253ce36763f3d326b5af7c598d7/test/undefined_errors_test.erl), specifically by [`undefined_errors: normalize_remote_type()`](https://github.com/josefs/Gradualizer/blob/b21659436a320253ce36763f3d326b5af7c598d7/test/misc/undefined_errors.erl#L19) which calls `undefined_errors_helper:und_ty()`, [the spec of which in turn refers back to `undefined_errors:undefined_type()`](https://github.com/josefs/Gradualizer/blob/b21659436a320253ce36763f3d326b5af7c598d7/priv/test/undefined_errors_helper.erl#L11). The implementation from yesterday couldn't catch it, but today I added a clause which looks for remote function calls and checks types used in specs of those remote functions.

If that's what you had in mind, I think we could live with this solution in a separate pass, as there are plenty of other bugs waiting to be fixed, and this is good enough to make CI green. WDYT?

erszcz:
> It seems to me this case is covered by tests in `undefined_errors_test.erl`.

Ok, I'm taking that back, it's not tested yet. The one I described above is testing reference to a remote type via a spec, not via remote type expansion.

erszcz:
> What about remote types that expand to other remote types?

Ok, with the latest changes we also explicitly handle that 👍

erszcz:
@zuiderkwast Any feedback on this one? I think it's ready for merging.

> What about remote types that expand to other remote types?

We handle that.

> Instead of a completely separate pass, how about checking for undefined remote types (and undefined user types and records) just after importing a type or spec from gradualizer_lib and just before removing the pos from these types?

I considered that, but I don't think that anytime before actually calling `type_check_forms` we can say with certainty that all files which ought to be imported into the DB, are already imported. In other words, if we check for an undefined remote type in `gradualizer_db` when importing a file, it might be the case that the type is undefined now, but it's defined in the next file queued up for importing. Unless I misunderstood you?

src/typechecker.erl
```diff
@@ -4812,6 +4819,67 @@ create_fenv(Specs, Funs) ->
...
+%%
+%% See https://github.com/josefs/Gradualizer/issues/379 for more information.
+%% @end
+-spec check_undefined_types(Forms) -> Errors when
+      Forms :: gradualizer_file_utils:abstract_forms(),
+      Errors :: [ Error ],
+      Error :: {not_exported, remote_type, erl_anno:anno(), {module(), atom(), non_neg_integer()}}
+             | {undef, remote_type, erl_anno:anno(), {module(), atom(), non_neg_integer()}}.
+check_undefined_types(Forms) ->
+    gradualizer_lib:fold_ast(fun check_remote_type/2, [], Forms).
```
zuiderkwast, line 4843: Correct me if I'm wrong, but I can't see that we handle the following:

* remote types which expand to structures containing user types or records which are undefined
* remote types which expand to structures containing more remote types
* user types and records belong to a remote module (i.e. they come from expanding a remote type) which refer to other remote types, user types and records
* cycles of the above

Some of it could be solved if gradualizer_db would perform linting when parsing a module, but I don't think it does (yet). (Btw, that's another topic: Unify the gradualizer_db and typechecker code to parse and extracts stuff from a parse tree.)

erszcz: What would the difference between:

> remote types which expand to structures containing user types or records which are undefined

and

> user types and records belong to a remote module (i.e. they come from expanding a remote type) which refer to other remote types, user types and records

be? Could you give an example?

I've added a number of tests - could you confirm if they capture what you had in mind above? Apparently, there was no need to change the check itself.

zuiderkwast: > remote types which expand to structures containing user types or records which are undefined

```Erlang
-module(foo).
-export_type([t/0]).
-type t() :: {undef(), #undef{}}.
```

> user types and records belong to a remote module (i.e. they come from expanding a remote type) which refer to other remote types, user types and records

```Erlang
-module(bar).
-export_type([t/0]).
-type t() :: [w() | r()].
-type w() :: {foo:undef()}.
-record(r, {f :: foo:undef()}).
```

If these are covered by tests, then I suspect the line and column are either zero (because I can't see that they're covered by the new pass, but only by the regular typechecking pass) or if they're non-zero, maybe line and column refer to the location in foo or bar where the error is rather than the module being type checked. We can probably report the line and column where foo:t() or bar:t() is used in the module being checked rather than where the error is in the foo and bar modules.

erszcz: You were right, this work so far only improved one particular situation when the reference was to a remote type which was not exported. All the other reports were still based on type representation fetched from `gradualizer_db` where position information was already lost, so the reported lines were 0.

erszcz:
["Store position info in gradualizer_db"](https://github.com/josefs/Gradualizer/pull/380/commits/81da18befe79b31e22421d0272bd19867d69e2b1) changes where type information is removed in `gradualizer_db`. Instead of removing it on import, it's kept in the DB and only removed before returning from the DB. The caller can decide whether to remove the position information or keep it. The main typechecker pass asks for removing it, the `check_undefined_types` pass asks to keep it.

Results from current master:
```
> gradualizer:type_check_file("test/misc/undefined_errors.erl", []).
test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type2/0 on line 0
test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type3/0 on line 0
test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type5/0 on line 0
test/misc/undefined_errors.erl: Undefined remote type undefined_errors_helper:not_exported_type/0 on line 0
test/misc/undefined_errors.erl: Undefined type undefined_type1/0 on line 0
test/misc/undefined_errors.erl: Undefined type undefined_type4/0 on line 0
test/misc/undefined_errors.erl: The atom on line 25 at column 41 is expected to have type #undefined_record1{} but it has type ok

-spec remote_struct_with_record() -> undefined_errors_helper:expands_to_struct_with_undefined_record().
remote_struct_with_record() -> {struct, ok}.
                                        ^^

test/misc/undefined_errors.erl: Call to undefined function undefined_errors_helper:undefined_call/0 on line 28 at column 41
test/misc/undefined_errors.erl: The function call on line 32 at column 20 is expected to have type #defined_record{} but it has type #undefined_record2{}

-record(defined_record, {a, b, c}).
-spec remote_record() -> #defined_record{}.
remote_record() -> undefined_errors_helper:und_rec().
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

nok
>
> gradualizer:type_check_file("priv/test/undefined_errors_helper.erl", []).
priv/test/undefined_errors_helper.erl: record undefined_record1 undefined on line 14 at column 61
priv/test/undefined_errors_helper.erl: record undefined_record2 undefined on line 16 at column 20
priv/test/undefined_errors_helper.erl: type undefined_type1() undefined on line 9 at column 14
priv/test/undefined_errors_helper.erl: type undefined_type4() undefined on line 13 at column 60
nok
```

Results with "Store position info in gradualizer_db":
```
> gradualizer:type_check_file("test/misc/undefined_errors.erl", []).
test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type2/0 on line 11
test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type3/0 on line 12
test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type5/0 on line 19
test/misc/undefined_errors.erl: Undefined remote type undefined_errors_helper:not_exported_type/0 on line 37 at column 25
test/misc/undefined_errors.erl: Undefined type undefined_type1/0 on line 0
test/misc/undefined_errors.erl: Undefined type undefined_type4/0 on line 0
test/misc/undefined_errors.erl: The atom on line 25 at column 41 is expected to have type #undefined_record1{} but it has type ok

-spec remote_struct_with_record() -> undefined_errors_helper:expands_to_struct_with_undefined_record().
remote_struct_with_record() -> {struct, ok}.
                                        ^^

test/misc/undefined_errors.erl: Call to undefined function undefined_errors_helper:undefined_call/0 on line 28 at column 41
test/misc/undefined_errors.erl: The function call on line 32 at column 20 is expected to have type #defined_record{} but it has type #undefined_record2{}

-record(defined_record, {a, b, c}).
-spec remote_record() -> #defined_record{}.
remote_record() -> undefined_errors_helper:und_rec().
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

nok
>
> gradualizer:type_check_file("priv/test/undefined_errors_helper.erl", []).
priv/test/undefined_errors_helper.erl: record undefined_record1 undefined on line 14 at column 61
priv/test/undefined_errors_helper.erl: record undefined_record2 undefined on line 16 at column 20
priv/test/undefined_errors_helper.erl: type undefined_type1() undefined on line 9 at column 14
priv/test/undefined_errors_helper.erl: type undefined_type4() undefined on line 13 at column 60
nok
```

Diff of the above:
```
--- remote-types.before.log	2022-02-15 14:44:11.000000000 +0100
+++ remote-types.after.log	2022-02-15 14:44:13.000000000 +0100
@@ -1,8 +1,8 @@
 > gradualizer:type_check_file("test/misc/undefined_errors.erl", []).
-test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type2/0 on line 0
-test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type3/0 on line 0
-test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type5/0 on line 0
-test/misc/undefined_errors.erl: Undefined remote type undefined_errors_helper:not_exported_type/0 on line 0
+test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type2/0 on line 11
+test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type3/0 on line 12
+test/misc/undefined_errors.erl: Undefined remote type undefined_errors:undefined_type5/0 on line 19
+test/misc/undefined_errors.erl: Undefined remote type undefined_errors_helper:not_exported_type/0 on line 37 at column 25
 test/misc/undefined_errors.erl: Undefined type undefined_type1/0 on line 0
 test/misc/undefined_errors.erl: Undefined type undefined_type4/0 on line 0
 test/misc/undefined_errors.erl: The atom on line 25 at column 41 is expected to have type #undefined_record1{} but it has type ok
```

As can be seen above we can still improve on:
```
 test/misc/undefined_errors.erl: Undefined type undefined_type1/0 on line 0
 test/misc/undefined_errors.erl: Undefined type undefined_type4/0 on line 0
```

and there 1 test failing, but the direction is good :+1:

zuiderkwast:
I just added `remove_pos` calls to `gradualizer_db` in a PR very recently. :) TBH I don't like the direction with added complexity.

When we check `mod.erl` which refers to `remote:t()` which has some error, I think the error location should be that within mod.erl, so the position in gradualizer_db is not necessary. Those error positions will be returned when `remote.erl` is checked.

erszcz:
> I just added remove_pos calls to gradualizer_db in a PR very recently. :)

I realise that and I think it was a good move. I'm just thinking it might be needed in slightly different places in `gradualizer_db`, but I'm not fixed on this being the right or the only way to solve the problem.

> When we check mod.erl which refers to remote:t() which has some error, I think the error location should be that within mod.erl, so the position in gradualizer_db is not necessary. Those error positions will be returned when remote.erl is checked.

Hmmm, so do you think we should provide this position information after we catch the thrown `{undef, remote_type, ...}`, but before we return from `typechecker:type_check_forms`? That might work 🤔

erszcz:
Superseded by #384.

erszcz:
> It doesn't necessarily have to be a catch. If we pass around the current location [...]

Obviously. I looked if we could pass the spec down from `type_check_function`, but apparently it's not available there.

Using the function's own position info would be a bit lame, as it's not the same as that of the spec (it might be completely off, there's no requirement on specs actually being close to the functions they specify, see https://github.com/erlang/otp/blob/60497f1b78d9fc7f02e16b8dba035548abbf4e9f/lib/stdlib/src/otp_internal.hrl#L32-L36 for an example).

So the spec is only looked up in `#env.fenv` downstream of `type_check_function`. `#env.fenv` on the other hand is populated as early as in `create_env`. All in all, it seemed to be easier to catch and rethrow on the site of fetching the spec from `#env.fenv` - please see #384 for the actual solution. It's significantly simpler than this PR.

PR #383, opened 2022-02-14
Temporary measure: break cycles by timing out

erszcz:
Current master branch is known to fall into infinite loops on some forms. It's evidenced by the property tests and real-world issues like #360. This PR tries to improve the UX and facilitate dogfooding by a **temporary measure** of forcibly breaking infinite loops after a timeout and therefore turning situations of hogging the CPU or never terminating into warning messages.

When Gradualizer is run interactively from the CLI or as a Rebar plugin, it's easy to conclude that it's stuck on a given file. However, it's still problematic when it gets stuck, as it prevents us from assessing the number of forms in the project which cause it to loop.

When Gradualizer is run as a background integration task, the only means of realising that it's stuck is the lack of diagnostics and the sound of fans cooling down the CPU.

This PR improves the situation in both of the above cases:
- it allows to finish the CLI / Rebar plugin run and assess the number of `Timeout` warnings to understand how many forms causing Gradualizer to fail are present in real-world code,
- it allows to finish a background run and only mark some forms in the editor with the `Timeout` diagnostic, providing valuable feedback on other forms which do not time out and are successfully type checked.

### DIFF
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -342,6 +342,20 @@ format_type_error({bad_type_annotation, TypeLit}, Opts) ->
       [format_location(TypeLit, brief, Opts),
        pp_expr(TypeLit, Opts),
        format_location(TypeLit, verbose, Opts)]);
+format_type_error({internal_error, form_check_timeout, Form}, Opts) ->
+    io_lib:format(
+      "~sTimeout checking form~s~n",
+      [format_location(Form, brief, Opts),
+       case proplists:get_value(fmt_location, Opts, ?FMT_LOCATION_DEFAULT) of
+           brief ->
+               "";
+           verbose ->
+               io_lib:format("~s~n~p~n"
+                             "This is most likely a bug in Gradualizer.~n"
+                             "Please report it at https://github.com/josefs/Gradualizer/issues",
+                             [format_location(Form, verbose, Opts),
+                              Form])
+       end]);
 format_type_error({Location, Module, ErrorDescription}, Opts)
   when is_integer(Location) orelse is_tuple(Location),
        is_atom(Module) ->
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -344,17 +344,17 @@ format_type_error({bad_type_annotation, TypeLit}, Opts) ->
        format_location(TypeLit, verbose, Opts)]);
 format_type_error({internal_error, form_check_timeout, Form}, Opts) ->
     io_lib:format(
-      "~sTimeout checking form~s~n",
+      "~sTimeout checking ~s~n",
       [format_location(Form, brief, Opts),
        case proplists:get_value(fmt_location, Opts, ?FMT_LOCATION_DEFAULT) of
            brief ->
-               "";
+               "form";
            verbose ->
-               io_lib:format("~s~n~p~n"
+               io_lib:format("~s~s~n"
                              "This is most likely a bug in Gradualizer.~n"
                              "Please report it at https://github.com/josefs/Gradualizer/issues",
-                             [format_location(Form, verbose, Opts),
-                              Form])
+                             [form_info(Form),
+                              format_location(Form, verbose, Opts)])
        end]);
 format_type_error({Location, Module, ErrorDescription}, Opts)
   when is_integer(Location) orelse is_tuple(Location),
@@ -371,6 +371,9 @@ format_type_error({none, Module, ErrorDescription}, _Opts)
     io_lib:format("~s~n", [Module:format_error(ErrorDescription)]);
 format_type_error(type_error, _) ->
     io_lib:format("TYPE ERROR~n", []).
+
+form_info({function, _, Name, Arity, _}) ->
+    io_lib:format("function ~s/~p", [Name, Arity]).
 
 -spec format_expr_type_error(gradualizer_type:abstract_expr(),
 			typelib:extended_type(),
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -1,5 +1,9 @@
 -module(gradualizer_fmt).
--export([format_location/2, format_type_error/2, print_errors/2, handle_type_error/2]).
+-export([format_location/2,
+         format_type_error/2,
+         print_errors/2,
+         handle_type_error/2,
+         form_info/1]).
 
 -include("typelib.hrl").
 
--- a/src/gradualizer_task.erl
+++ b/src/gradualizer_task.erl
@@ -144,13 +144,13 @@ do_apply(TaskOwnerInfo, {Mod, Fun, Args} = MFA) ->
     try
         erlang:apply(Mod, Fun, Args)
     catch
-        error: Value ->
+        error:Value:Stacktrace ->
             task_exit(TaskOwnerInfo, MFA,
-                      {Value, erlang:get_stacktrace()});
-        throw: Value ->
+                      {Value, Stacktrace});
+        throw:Value:Stacktrace ->
             task_exit(TaskOwnerInfo, MFA,
-                      {{nocatch, Value}, erlang:get_stacktrace()});
-        exit: Value ->
+                      {{nocatch, Value}, Stacktrace});
+        exit:Value ->
             task_exit(TaskOwnerInfo, MFA, Value)
     end.
 
--- a/src/gradualizer_task.erl
+++ b/src/gradualizer_task.erl
@@ -1,0 +1,184 @@
+%% @doc Task is a helper for running asynchronous operations.
+%%
+%% This is a port of the Elixir Task module borrowed from https://github.com/redink/task.
+%% Original Elixir Task documentation can be found at https://hexdocs.pm/elixir/Task.html.
+%% @end
+-module(gradualizer_task).
+
+-export([async/3,
+         async/4,
+         async/1,
+         async/2,
+         await/1,
+         await/2]).
+
+-export([async_opt/4,
+         async_opt/5,
+         async_opt/2,
+         async_opt/3]).
+
+-export([safe_await/2,
+         safe_await/3]).
+
+-export([async_do/3]).
+
+-spec async(function()) -> {pid(), reference()}.
+async(Fun) when erlang:is_function(Fun) ->
+    async(erlang, apply, [Fun, []]).
+
+-spec async(atom(), function()) -> {pid(), reference()}.
+async(Node, Fun) when erlang:is_function(Fun) ->
+    async(Node, erlang, apply, [Fun, []]).
+
+-spec async(atom(), atom(), [term()]) -> {pid(), reference()}.
+async(Mod, Fun, Args) ->
+    Me  = erlang:self(),
+    Pid = proc_lib:spawn_link(?MODULE, async_do,
+                              [Me, get_info(Me), {Mod, Fun, Args}]),
+    Ref = erlang:monitor(process, Pid),
+    erlang:send(Pid, {Me, Ref}),
+    {Pid, Ref}.
+
+-spec async(atom(), atom(), atom(), [term()]) -> {pid(), reference()}.
+async(Node, Mod, Fun, Args) ->
+    Me  = erlang:self(),
+    Pid = proc_lib:spawn_link(Node, ?MODULE, async_do,
+                              [Me, get_info(Me), {Mod, Fun, Args}]),
+    Ref = erlang:monitor(process, Pid),
+    erlang:send(Pid, {Me, Ref}),
+    {Pid, Ref}.
+
+-spec async_opt(function(), [term()]) -> {pid(), reference()}.
+async_opt(Fun, Opts) when erlang:is_function(Fun) ->
+    async_opt(erlang, apply, [Fun, []], Opts).
+
+-spec async_opt(atom(), function(), [term()]) -> {pid(), reference()}.
+async_opt(Node, Fun, Opts) when erlang:is_function(Fun) ->
+    async_opt(Node, erlang, apply, [Fun, []], Opts).
+
+-spec async_opt(atom(), atom(),
+                [term()], [term()]) -> {pid(), reference()}.
+async_opt(Mod, Fun, Args, Opts) ->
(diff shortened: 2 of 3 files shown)


erszcz:
> Current master branch is known to fall into infinite loops on some forms.

I have taken another look at it and I have to correct the above. I couldn't find evidence about the current master branch looping on real world code. Only on synthetic, generated recursive types in property tests.

Below is a comparison between 3 branches:
- `upstream-master`, i.e. this repo's master branch
- `erszcz-master`, i.e. my forks master branch with #359 and this PR merged in
- `erszcz-gradient`, my branch with just this PR merged in, currently used by https://github.com/esl/gradient

The test file contains 3 problems I identified when dogfooding with Gradualizer as a background diagnostics provider for ErlangLS or by running `make gradualize`:
- functions `f`, `g`, and `h` are the regressions introduced when extending exhaustiveness checking to non-trivial types - these are already mentioned in #359
- function `type_/1` which `case` matches on a relatively big and complex union type `type_()` - this checks if exhaustiveness checking, also on types with `string()`s and lists, works
- function `format_binding_type/2` which returns a deep list: `io_lib:chars()` which causes #359 to loop

The results are that:
- `upstream-master` doesn't loop on `io_lib:chars()`, but it has the regressions
- `erszcz-master` fixes the regressions, but it loops on `io_lib:chars()`, therefore to maintain usability it requires this PR
- `erszcz-gradient` doesn't loop on `io_lib:chars()`, but it has the regressions; as expected from the above two points, the fact that it has this PR merged in doesn't change anything

I also did a similar test of the code from #360 - the results are here: https://gist.github.com/erszcz/4323e97860e67da290f52e2b0998a863 (search for timeout). The only timing out function here is `inputs:get_input_filename/2`. This function operates on file names, specifically it refers to type `file:name_all()`, which similarly to `io_lib:chars()` also is a deep list. This suggests #359 loops on deep lists.

---

The test file in full detail:

```erlang
-module(syntax).

f( Needle, [Needle | _]) -> ok;
f( Needle, [_ | Haystack]) -> f(Needle, Haystack);
f(_Needle, []) -> ok.

-spec g([a | b]) -> ok.
g([a | _Haystack]) -> ok;
g([_ | Haystack]) -> g(Haystack);
g([]) -> ok.

-spec h([a | b]) -> ok.
h([a | _Haystack]) -> ok;
h([]) -> ok;
h([_ | Haystack]) -> h(Haystack).

%%
%%' Datatypes
%%

-type ty() :: {var, index(), context_size()}
            | {id, string()}
            | {arr, ty(), ty()}
            | unit
            | {record, [{string(), ty()}]}
            | {rec, string(), ty()}
            | {variant, [{string(), ty()}]}
            | bool
            | string
            | float
            | nat.

-type info() :: {pos_integer(), pos_integer()}.

-type token() :: {atom(), info()}
               | {atom(), info(), string()}.

-type index() :: non_neg_integer().
-type context_size() :: non_neg_integer().

-type term_() :: {true, info()}
               | {false, info()}
               | {if_, info(), term_(), term_(), term_()}
               | {case_, info(), term_(), [{string(), {string(), term_()}}]}
               | {tag, info(), string(), term_(), ty()}
               | {var, info(), index(), context_size()}
               | {abs, info(), string(), ty(), term_()}
               | {app, info(), term_(), term_()}
               | {let_, info(), string(), term_(), term_()}
               | {fix, info(), term_()}
               | {string, info(), string()}
               | {unit, info()}
               | {ascribe, info(), term_(), ty()}
               | {proj, info(), term_(), string()}
               | {record, info(), [{string(), term_()}]}
               | {float, info(), float()}
               | {timesfloat, info(), term_(), term_()}
               | {zero, info()}
               | {succ, info(), term_()}
               | {pred, info(), term_()}
               | {is_zero, info(), term_()}
               | {inert, info(), ty()}.
%% TAPL `term' type, but `term()' is a builtin type in Erlang,
%% hence the name `term_()'.

-type binding() :: name_bind
                 | ty_var_bind
                 | {var_bind, ty()}
                 | {tm_abb_bind, term_(), ty() | none}
                 | {ty_abb_bind, ty()}.

-type context() :: [{string(), binding()}].

-type command() :: {eval, info(), term()}
                 | {bind, info(), string(), binding()}.

%%.
%%' Constructors
%%

-spec ty(ty()) -> ty().
ty(Ty) ->
    case Ty of
        {var, _, _} -> Ty;
        %{id, _} -> Ty;
        {arr, _, _} -> Ty;
        unit -> Ty;
        {record, _} -> Ty;
        {rec, _, _} -> Ty;
        {variant, _} -> Ty;
        bool -> Ty;
        string -> Ty;
        float -> Ty;
        nat -> Ty
    end.

-spec info(token()) -> info().
info({_, Info}) -> Info;
info({_, Info, _Chars}) -> Info.

%% This function might seem useless, since it "doesn't do anything",
%% but in fact it gives us two guarantees:
%% - thanks to Gradualizer it provides compile-time warnings if we build an invalid term_()
%% - it fails fast (aka crashes) at runtime if we try to build an invalid term_()
-spec term_(term_()) -> term_().
term_(T) ->
    case T of
        {true, _} -> T;
        %{false, _} -> T;
        {if_, _, _, _, _} -> T;
        {case_, _, _, _} -> T;
        {tag, _, _, _, _} -> T;
        {var, _, _, _} -> T;
        {abs, _, _, _, _} -> T;
        {app, _, _, _} -> T;
        {let_, _, _, _, _} -> T;
        {fix, _, _} -> T;
        {string, _, _} -> T;
        {unit, _} -> T;
        {ascribe, _, _, _} -> T;
        {proj, _, _, _} -> T;
        {record, _, _} -> T;
        {float, _, _} -> T;
        {timesfloat, _, _, _} -> T;
        {zero, _} -> T;
        {succ, _, _} -> T;
        {pred, _, _} -> T;
        {is_zero, _, _} -> T;
        {inert, _, _} -> T
    end.

-spec binding(binding()) -> binding().
binding(B) ->
    case B of
        name_bind -> B;
        ty_var_bind -> B;
        {var_bind, _} -> B;
        {tm_abb_bind, _, _} -> B;
        {ty_abb_bind, _} -> B
    end.

-spec command(command()) -> command().
command(C) ->
    case C of
        {eval, _, _} -> C;
        {bind, _, _, _} -> C
    end.

%%.

-spec format_binding_type(context(), binding()) -> io_lib:chars().
%-spec format_binding_type(context(), binding()) -> string().
format_binding_type(Ctx, B) ->
    case B of
        name_bind -> "";
        ty_var_bind -> "";
        {var_bind, TyT} ->
            io_lib:format(": ~ts", [zxcqwe]);
        {tm_abb_bind, T, MaybeTyT} ->
            case MaybeTyT of
                none -> [": ", "asdzxc"];
                TyT -> [": ", "qwe123"]
            end;
        {ty_abb_bind, _} -> ":: *"
    end.
```

Results:

```
$ for GRADUALIZER in $(find bin -name gradualizer.\*); do ($GRADUALIZER --version; $GRADUALIZER syntax_example.erl) | tee $(basename $GRADUALIZER).log; echo; done
Gradualizer v0.1.3-108-g5be0f94
syntax_example.erl: The clause on line 5 at column 1 cannot be reached
syntax_example.erl: The pattern [_ | Haystack] on line 9 at column 3 does not have type:
[]
syntax_example.erl: The clause on line 15 at column 1 cannot be reached
syntax_example.erl: Nonexhaustive patterns on line 84 at column 9
	{id, []}
syntax_example.erl: Nonexhaustive patterns on line 108 at column 9
	{false, {0, 0}}

Gradualizer v0.1.3-117-g26acc93
syntax_example.erl: The clause on line 5 at column 1 cannot be reached
syntax_example.erl: The pattern [_ | Haystack] on line 9 at column 3 does not have type:
[]
syntax_example.erl: The clause on line 15 at column 1 cannot be reached
syntax_example.erl: Nonexhaustive patterns on line 84 at column 9
	{id, []}
syntax_example.erl: Nonexhaustive patterns on line 108 at column 9
	{false, {0, 0}}

Gradualizer v0.1.3-150-gef2dbb7
syntax_example.erl: Nonexhaustive patterns on line 84 at column 9
	{id, []}
syntax_example.erl: Nonexhaustive patterns on line 108 at column 9
	{false, {0, 0}}
syntax_example.erl: Timeout checking function format_binding_type/2 on line 153 at column 1
```

src/gradualizer_task.erl
```diff
@@ -0,0 +1,184 @@
+%% @doc Task is a helper for running asynchronous operations.
+%%
+%% This is a port of the Elixir Task module borrowed from https://github.com/redink/task.
+%% Original Elixir Task documentation can be found at https://hexdocs.pm/elixir/Task.html.
+%% @end
+-module(gradualizer_task).
```
zuiderkwast, line 6: If this would be part of the standard library (as it is in Elixir) I'd be fine using it. Porting it and including it here seems like lots of added complexity though. Isn't it enough to use spawn_link or spawn_monitor and set a timeout?

```Erlang
%% Pseudo-code
TypecheckerPid = spawn_link(fun () -> type_check(Froms, self()) end),
receive
    {typechecker_done, Result} -> Result;
after Timeout ->
    kill(TypecheckerPid),
    {timeout, Forms)
end.
```

erszcz: It worked pretty well for me, but we now have a hand rolled `async/3` if you prefer 🤷‍♂️

erszcz: @zuiderkwast you must've got some sixth sense 😆 It turned out the `task.erl` would return with a timeout, so other forms would be checked and reported, but at the same time the async task would go rouge and hog the CPU in an infinite loop anyway. I've now tested in ErlangLS with the hand-rolled `spawn_monitor` and timeout and the CPU is way less busy 👍

erszcz:
@zuiderkwast This is ready with the fix you suggested. Do you have any further comments?

src/gradualizer_fmt.erl
```diff
@@ -342,6 +346,20 @@ format_type_error({bad_type_annotation, TypeLit}, Opts) ->
       [format_location(TypeLit, brief, Opts),
        pp_expr(TypeLit, Opts),
        format_location(TypeLit, verbose, Opts)]);
+format_type_error({internal_error, form_check_timeout, Form}, Opts) ->
```
zuiderkwast, line 349: I think `{timeout, Form}` would be enough, to keep the error format simpler.

src/gradualizer_fmt.erl
```diff
@@ -342,6 +346,20 @@ format_type_error({bad_type_annotation, TypeLit}, Opts) ->
...
+      [format_location(Form, brief, Opts),
+       case proplists:get_value(fmt_location, Opts, ?FMT_LOCATION_DEFAULT) of
+           brief ->
+               "form";
+           verbose ->
+               io_lib:format("~s~s~n"
+                             "This is most likely a bug in Gradualizer.~n"
+                             "Please report it at https://github.com/josefs/Gradualizer/issues",
+                             [form_info(Form),
+                              format_location(Form, verbose, Opts)])
```
zuiderkwast, line 361: The `fmt_location` isn't meant to be a generic `--verbose` flag. It was added because someone was afraid of just changing the location output to the [GNU standard](https://www.gnu.org/prep/standards/html_node/Errors.html) file:line:col.

I think it should be possible to use brief location and still be use a verbose message. (Furthermore, I think "on line x column y" should be deleted, but that's a different story.)

src/typechecker.erl
```diff
@@ -4734,46 +4737,112 @@ get_record_info_type({call, Anno, {atom, _, record_info},
...
+                     [gradualizer_fmt:form_info(Function)]),
+            erlang:raise(error, Error, Trace);
+        {errors, Errors1} ->
+            ?verbose(Env, "Task returned from ~s with ~p~n",
+                     [gradualizer_fmt:form_info(Function), Errors1]),
+            Errors1
+    end.
+
+-spec timeout(fun(() -> any()), non_neg_integer(), fun(({'DOWN', _, _, _, _}) -> any())) -> any().
+timeout(TaskF, Timeout, DownF) ->
```
zuiderkwast, line 4795: This abstraction is smaller than the full Elixir `task` :+1: but it's still an abstraction only used once. I'm not a fan of that. (Small functions are good, but wrapping code in extra funs only used once is mainly added complexity.)

I'd rather put the spawn_monitor in type_check_form_with_timeout and replace the case..of with this receive here. The down fun isn't necessary.

src/typechecker.erl
```diff
@@ -4734,46 +4737,101 @@ get_record_info_type({call, Anno, {atom, _, record_info},
...
+type_check_form_with_timeout(Function, Errors, StopOnFirstError, Env, Opts) ->
+    %% TODO: make FormCheckTimeOut configurable
+    FormCheckTimeOut = ?form_check_timeout_ms,
+    ?verbose(Env, "Spawning async task...~n", []),
+    TaskF = fun () ->
+                    type_check_form(Function, Errors, StopOnFirstError,
+                                    Env, Opts)
+            end,
+    Self = self(),
+    {Pid, MRef} = spawn_monitor(fun () -> Self ! TaskF() end),
```
zuiderkwast, line 4771: I suppose it's a matter of taste, so I won't demand it.

I just note that it's possible to skip one more fun here:

```suggestion
    Self = self(),
    Task = fun () ->
                   Self ! type_check_form(Function, Errors, StopOnFirstError,
                                          Env, Opts)
           end,
    {Pid, MRef} = spawn_monitor(Task)
```

src/gradualizer_fmt.erl
erszcz: I think the `brief` format plays better with tools which want to integrate with Gradualizer, both due to the easier to parse location format and due to the shorter message. I imagine these could have independent switches in the future, but for the time being I'd try to keep the message brief too, if we ask for the brief location format.

That being said, we shouldn't completely omit a vital part of the message, so I've extended the brief text. Before:
Screenshot 2022-02-22 at 18 05 46

After:
Screenshot 2022-02-22 at 18 14 28

src/typechecker.erl
erszcz: Ahh, good catch. It was necessary with the auxiliary named function hiding message passing - it's not anymore.

PR #385, opened 2022-02-16
Provide the git commit sha when asked for --version in CLI

erszcz:


Makefile
```diff
@@ -55,10 +55,8 @@ ebin/typechecker.beam: src/typelib.hrl
 ebin/gradualizer_fmt.beam: src/typelib.hrl
 
 # .app file
-# TODO When we start using git tags, insert version from
-# `git describe --tags`
 ebin/gradualizer.app: src/gradualizer.app.src | ebin
-	cp $< $@
+	sed s/git/`git describe --tags --always`/ $< > $@
```
zuiderkwast, line 59: Isn't it risky to replace all occurences of "git"? What if we add these three letters somewhere else in the app file?

Maybe ``sed 's/{vsn, *"git"}/{vsn, `git describe --tags --always`}/'`` is more safe..? (I don't know what rebar3 and erlang.mk does, but I suppose at least rebar3 parses the file and replaces only this occurrence.)

erszcz: Good point, it definitely doesn't hurt to make the pattern more specific :+1:

zuiderkwast (approved):

Issue #387, opened 2022-02-19
Map pattern in function head leads to type error if spec is not present


erszcz:
We saw a few reports of this happening, one of them being https://github.com/esl/gradient/issues/37#issuecomment-1043412181. A test case in Erlang:

```
$ cat test.erl
-module(test).

-export([f/1, g/1, h/1, i/1]).

f({_} = _Tuple) -> ok.

g(#{} = _Map) -> ok.

h([_|_] = _List) -> ok.

-record(r, {a, b}).
i(#r{} = _Record) -> ok.
$ gradualizer test.erl
test.erl: The pattern #{} on line 7 at column 3 doesn't have the type any()
```

Interestingly, it happens only with maps 🤔

erszcz:
Fixed by #389.

PR #389, opened 2022-02-23
Fix for map pattern "doesn't have type any()" warning

erszcz:
Fix for #387.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4805,11 +4805,11 @@ type_check_form_with_timeout(Function, Errors, StopOnFirstError, Env, Opts) ->
            end,
     {Pid, MRef} = spawn_monitor(Task),
     Result = receive
-                 {crash, Crash, St} ->
+                 {crash, Crash, STrace} ->
                      ?verbose(Env, "Task reported crash on ~s~n",
                               [gradualizer_fmt:form_info(Function)]),
                      io:format("Crashing...~n"),
-                     erlang:raise(throw, Crash, St);
+                     erlang:raise(throw, Crash, STrace);
                  {error_trace, Error, Trace} ->
                      ?verbose(Env, "Task reported error with trace from ~s~n",
                               [gradualizer_fmt:form_info(Function)]),
@@ -4844,19 +4844,14 @@ type_check_form(Function, Errors, StopOnFirstError, Env, Opts)
         {_VarBinds, _Cs} ->
             {errors, Errors}
     catch
-        throw:Throw:St ->
-            % Useful for debugging
-            % io:format("~p~n", [erlang:get_stacktrace()]),
-            if
-                CrashOnError ->
-                    {crash, Throw, St};
-                not CrashOnError ->
-                    {errors, [Throw | Errors]}
-            end;
-        error:Error:ST ->
+        throw:Throw:STrace when CrashOnError ->
+            {crash, Throw, STrace};
+        throw:Throw ->
+            {errors, [Throw | Errors]};
+        error:Error:STrace ->
             %% A hack to hide the (very large) #env{} in
             %% error stacktraces. TODO: Add an opt for this.
-            Trace = case ST of
+            Trace = case STrace of
                         [{M, F, [#env{}|Args], Pos} | RestTrace] ->
                             [{M, F, ['*environment excluded*'|Args], Pos} | RestTrace];
                         Trace0 ->
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4317,7 +4317,9 @@ expect_map_type({var, _, Var}, _Env) ->
     Cs = constraints:add_var(Var, constraints:upper(Var, type(map, any))),
     {assoc_tys, any, Cs};
 expect_map_type(?type(map, AssocTys), _Env) ->
-    {assoc_tys, AssocTys, constraints:empty()}.
+    {assoc_tys, AssocTys, constraints:empty()};
+expect_map_type(Ty, _Env) ->
+    {type_error, Ty}.
 
 %% Rewrite map_field_assoc to map_field_exact to return in pattern types.
 %%
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4249,6 +4249,13 @@ add_type_pat({record, P, Record, Fields}, Ty, Env, VEnv) ->
         {type_error, _Type} ->
             throw({type_error, record_pattern, P, Record, Ty})
     end;
+%% TODO implement expect_list_type
+%add_type_pat({map, _, _} = MapPat, MapTy, _Env, VEnv) ->
+%    case expect_map_type(normalize(ListTy, Env), Env) of
+%    end;
+%% this is just a workaround:
+add_type_pat({map, _, _} = MapPat, ?type(any), _Env, VEnv) ->
+    {type(none), type(any), add_any_types_pat(MapPat, VEnv), constraints:empty()};
 add_type_pat({map, _, _} = MapPat, {var, _, Var} = TyVar, _Env, VEnv) ->
     %% FIXME this is a quite rudimentary implementation
     %% - variables from the map pattern become any()
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4250,7 +4250,8 @@ add_type_pat({record, P, Record, Fields}, Ty, Env, VEnv) ->
             throw({type_error, record_pattern, P, Record, Ty})
     end;
 add_type_pat({map, P, AssocPats} = MapPat, MapTy, Env, VEnv) ->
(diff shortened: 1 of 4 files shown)


src/typechecker.erl
```diff
@@ -4249,6 +4249,13 @@ add_type_pat({record, P, Record, Fields}, Ty, Env, VEnv) ->
         {type_error, _Type} ->
             throw({type_error, record_pattern, P, Record, Ty})
     end;
+%% TODO implement expect_list_type
```
zuiderkwast, line 4252:
```suggestion
%% TODO implement expect_map_type
```

src/typechecker.erl
```diff
@@ -4846,7 +4846,7 @@ type_check_form(Function, Errors, StopOnFirstError, Env, Opts)
     catch
         throw:Throw:St ->
             % Useful for debugging
-            % io:format("~p~n", [erlang:get_stacktrace()]),
+            % io:format("~p~n", [St]),
             if
                 CrashOnError ->
                     {crash, Throw, St};
```
zuiderkwast, line 4852: I think these commands can be deleted. We have CrashOnError which returns the stacktrace instead. I assume these comments are from before that.

Btw, the if expression could be incorporated in the catch clauses as guards, e.g.

```Erlang
        throw:Throw:St when CrashOnError ->
            {crash, Throw, St};
        throw:Throw ->
            {errors, [Throw | Errors]};
        error:Error:ST ->
            ...
```

erszcz:
Ok, I think it's good enough for now - at least it fixes #387. Constraints could be improved according to the FIXME comment, but with not constraint solver it's not testable anyway, so will like require some attention in the future.

erszcz:
@zuiderkwast what do you think?

src/typechecker.erl
```diff
@@ -4308,11 +4302,32 @@ add_type_pat(OpPat = {op, _Anno, _Op, _Pat}, Ty, Env, VEnv) ->
...
+expect_map_type(?top(), _Env) ->
+    {assoc_tys, [type(map_field_assoc, [top(), top()])], constraints:empty()};
+expect_map_type({var, _, Var}, _Env) ->
+    %% FIXME this is a quite rudimentary implementation
+    %% - variables from the map pattern become any()
+    %% - the constraint could contain the map keys
+    Cs = constraints:add_var(Var, constraints:upper(Var, type(map, any))),
+    {assoc_tys, any, Cs};
+expect_map_type(?type(map, AssocTys), _Env) ->
+    {assoc_tys, AssocTys, constraints:empty()}.
```
zuiderkwast, line 4320: I think we need a catch-all cause for non-map types here. Before, those types fell through in `add_type_pat` to `throw({type_error, pattern, P, MapPat, MapTy})` but now `add_type_pat` catches all map patterns and calls `expect_map_type` instead.

Maybe we should have a test case for something like this? (IIRC how to reach add_type_pat)

```Erlang
f() ->
    G = g(),
    h(G).

-spec g() -> {tup, le}.
g() -> {tup, le}.

h(#{k := v} = _Map) ->
    ok.
```

zuiderkwast (commented):
:+1: I didn't take my time to fully understand it, but it looks like how we handle lists and other stuff.

src/typechecker.erl
erszcz: Good point, thanks :+1: I don't return `{type_error, ...}` and all the other `expect_*_type` functions do.

erszcz: I've added https://github.com/erszcz/Gradualizer/blob/f5b2468090636f38b8a1029287b6b862094b0c3c/test/should_fail/map_pattern_fail.erl#L21-L30 but it doesn't seem to trigger anything even with`expect_map_type` not returning `{type_error, ...}`...

The `g` and `h` specs are required for the warning to show up:
```
     22 not_a_map_passed_as_map() ->
     23     G = g(),
  W  24     h(G).     ■ The variable G is expected to have type map() but it has type {tup, le}
     25
     26 -spec g() -> {tup, le}.
     27 g() -> {tup, le}.
     28
     29 -spec h(map()) -> ok.
     30 h(#{k := v} = _Map) ->
     31     ok.
```

So... it seems to just work 🤔

erszcz: In other words, the presence of this:
```
diff --git a/src/typechecker.erl b/src/typechecker.erl
index ebecda1..4d2fb80 100644
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4317,7 +4317,9 @@ expect_map_type({var, _, Var}, _Env) ->
     Cs = constraints:add_var(Var, constraints:upper(Var, type(map, any))),
     {assoc_tys, any, Cs};
 expect_map_type(?type(map, AssocTys), _Env) ->
-    {assoc_tys, AssocTys, constraints:empty()}.
+    {assoc_tys, AssocTys, constraints:empty()};
+expect_map_type(Ty, _Env) ->
+    {type_error, Ty}.
 
 %% Rewrite map_field_assoc to map_field_exact to return in pattern types.
 %%
```

Doesn't seem to make any difference whatsoever (I checked all combinations of specs present/missing).

zuiderkwast: If we have coverage on the other `expect_map_type` clauses, then we must be able to get coverage on the last clause too.

We really should know how to trigger the different parts of the code. We shouldn't rely on guess work and trial-and-error.

erszcz: > We really should know how to trigger the different parts of the code. We shouldn't rely on guess work and trial-and-error.

Exactly, that's why I'm thoroughly testing it on examples (what could be called "trial-and-error", but I'd argue). The thing is I couldn't find any that would trigger this new clause. If there's no example triggering the code, the code's not needed - that's what I'm trying to say. Anyway, I'll trace execution and see if I can come up with something that would hit the new clause.

src/typechecker.erl
```diff
@@ -4308,11 +4302,32 @@ add_type_pat(OpPat = {op, _Anno, _Op, _Pat}, Ty, Env, VEnv) ->
...
+
 %% Rewrite map_field_assoc to map_field_exact to return in pattern types.
 %%
 %% Similarly to map field type inference on map creation - if a pattern matches,
 %% then the map field is exact (:=), not assoc (=>).
 %% There isn't even syntax for optional fields in map patterns.
+rewrite_map_assocs_to_exacts(?top()) ->
+    top();
+rewrite_map_assocs_to_exacts({var, _, _Var}) ->
+    type(none);
```
zuiderkwast, line 4330: What are these clauses doing here? Var :: none() seems very weird to me.

I though we only call this function on explicit map types.

src/typechecker.erl
zuiderkwast: Keep up the good work. :+1: If you can hit the clause just above, then I don't understand how it can be impossible to cover the next clause.....

zuiderkwast: Of course guess work and testing all kinds of examples is useful to figure out how it works. We just shouldn't commit code that we don't know how it works. That's what I meant.

src/typechecker.erl
erszcz: That was the case before introducing `expect_map_type`. Now multiple of the former `add_type_pat` clauses fall into case clause for `{assoc_tys, ...}`, therefore all their return values have to be handled here, too, unless we want to explicitly branch there 🤷‍♂️ . That's just a trade-off between where to branch, nothing new introduced here.

erszcz: Ahh, and it's suggested when it can be `none` here:
https://github.com/josefs/Gradualizer/blob/c3ee078616a88fc25655cb4bc40e81cda2f01cd3/src/typechecker.erl#L4036

zuiderkwast: Yeah there are some strange types used for refinement. If we spill that over to other functions, we'd have to describe it there too... Better keep the semantics of each function simple, so maybe branch before this function is called then..?

src/typechecker.erl
erszcz: Mystery solved:

```
 expect_map_type(?type(map, AssocTys), _Env) ->
-    {assoc_tys, AssocTys, constraints:empty()}.
+    {assoc_tys, AssocTys, constraints:empty()};
+expect_map_type(Ty, _Env) ->
+    {type_error, Ty}.
```

protects against a buggy spec as shown by https://github.com/josefs/Gradualizer/pull/389/commits/7aed2b2a3c17aac5c602ff2df7d99c2fc8913ddd.

src/typechecker.erl
erszcz: Cleaned up now :+1:

erszcz:
@zuiderkwast Ok, looks good to me, but let me know what you think.

zuiderkwast (approved):
Yes, LGTM!

Issue #339, opened 2021-08-05
Exhaustiveness check for tuple expressions in case_of fails


berbiche:
Hi,

I did not find a similar issue opened.

I encountered an issue where the refined type for a `case of` is `any()` while the type used
to check for `refinable` is a `tuple` if too many fields are checked against.

I have not done a deep dive of the issue yet but the sample code below will reproduce the issue.

``` erlang
-type union_a() :: a | b | c | d | e.
-type union_b() :: m | n.

-spec union_atom(undefined | union_a(), undefined | union_b()) -> {}.
union_atom(X, Y) ->
    %% The second field does not seem to matter
    case {X, Y} of
        {a, m} -> {};
        {b, m} -> {};
        {c, m} -> {};
        {d, m} -> {};
        %% Commenting out the next match will NOT produce the error
        {e, m} -> {};
        %% Swapping the match above with any of the next matches also reproduce the error
        %% This seems to indicate that the number of matches is the problem
        %{a, n} -> {};
        %{b, n} -> {};
        %{_, undefined} -> {};
        %{undefined, _} -> {};
        {_, _} -> {}
    end.

%% The minimum amount of fields in the union
%% to reproduce the issue seems to be (3x5) or (5x3)
-spec union_atom_inline(union_a(), union_b() | o) -> {}.
union_atom_inline(X, Y) ->
    case {X, Y} of
        {a, m} -> {};
        {b, m} -> {};
        {c, m} -> {};
        {d, m} -> {};
        %% Commenting out the next match will NOT produce the error
        {e, m} -> {};
        {_, _} -> {}
    end.
```

Replacing the above `case of` with function patterns will not result in an error.
Changing the last match for `_` also fixes the issue.

Stacktrace:
```
escript: exception error: no function clause matching
                 gradualizer_lib:pick_value({type,0,any,[]},
                                            #{module => exhaustive,
                                              records => #{},
                                              types =>
                                               #{{union_a,0} =>
                                                  {[],
                                                   {type,0,union,
                                                    [{atom,0,a},
                                                     {atom,0,b},
                                                     {atom,0,c},
                                                     {atom,0,d},
                                                     {atom,0,e}]}},
                                                 {union_b,0} =>
                                                  {[],
                                                   {type,0,union,
                                                    [{atom,0,m},
                                                     {atom,0,n}]}}}}) (src/gradualizer_lib.erl, line 137)
[...]

PR #392, opened 2022-03-09
Make binary() aka String.t() in Elixir refinable

erszcz:
`String.t()` is the default string type in Elixir just as `string()` is the default in Erlang. However, when Erlang `string()` is a _charlist_, i.e. a list of integers, the Elixir default string type is actually a binary. This makes binaries ubiquitous in Elixir code, much more so than in Erlang. To make exhaustiveness checking useful and relevant in Elixir (for example in cases such as #391), it has to support the binary type.

Alas, this brings in some problems, the most obvious one being a broken test case. Therefore, until this is solved, I'm making this a draft.

### DIFF
--- a/src/gradualizer_db.erl
+++ b/src/gradualizer_db.erl
@@ -608,7 +608,9 @@ get_beam_map() ->
                 {match, [Mod]} ->
                     {true, {list_to_atom(Mod), Filename}};
                 nomatch ->
-                    false
+                    false;
+                _ ->
+                    erlang:error({unreachable, "check re:run/3 opts above - this should not happen"})
             end
         end,
         BeamFiles),
--- a/src/gradualizer_lib.erl
+++ b/src/gradualizer_lib.erl
@@ -193,8 +193,8 @@ pick_value(?type(nonempty_list, Ty), Env) ->
     {cons, erl_anno:new(0), H, {nil, erl_anno:new(0)}};
 pick_value(?type(nil), _Env) ->
     {nil, erl_anno:new(0)};
-pick_value(?type(binary, _), _Env) ->
-    {bin, erl_anno:new(0), []};
+pick_value(?type(binary, _) = B, Env) ->
+    pick_binary_value(B, Env);
 %% The ?type(range) is a different case because the type range
 %% ..information is not encoded as an abstract_type()
 %% i.e. {type, Anno, range, [{integer, Anno2, Low}, {integer, Anno3, High}]}
@@ -222,6 +222,22 @@ pick_value(Type, Env)
             throw({undef, Kind, Anno, {Name, length(Args)}})
     end.
 
+pick_binary_value(?type(binary, []), _Env) ->
+    {bin, erl_anno:new(0), []};
+pick_binary_value(?type(binary, [{integer, _, 0}, {integer, _, _}]), _Env) ->
+    {bin, erl_anno:new(0), []};
+pick_binary_value(?type(binary, [{integer, _, M}, {integer, _, _}]), _Env) ->
+    Elements = case M rem 8 of
+                   0 ->
+                       S = [ $a || _ <- lists:seq(1, M div 8) ],
+                       [{bin_element, 0, {string, 0, S}, default, default}];
+                   _ ->
+                       Eights = [ {bin_element, 0, {integer, 0, 0}, default, default}
+                                  || _ <- lists:seq(1, M div 8) ],
+                       Tail = {bin_element, 0, {integer, 0, 0}, {integer, 0, M rem 8}, default},
+                       Eights ++ [Tail]
+               end,
+    {bin, erl_anno:new(0), Elements}.
 
 %% ------------------------------------------------
 %% Functions for working with abstract syntax trees
--- a/src/gradualizer_lib.erl
+++ b/src/gradualizer_lib.erl
@@ -193,6 +193,8 @@ pick_value(?type(nonempty_list, Ty), Env) ->
     {cons, erl_anno:new(0), H, {nil, erl_anno:new(0)}};
 pick_value(?type(nil), _Env) ->
     {nil, erl_anno:new(0)};
+pick_value(?type(binary, _), _Env) ->
+    {bin, erl_anno:new(0), []};
 %% The ?type(range) is a different case because the type range
 %% ..information is not encoded as an abstract_type()
 %% i.e. {type, Anno, range, [{integer, Anno2, Low}, {integer, Anno3, High}]}
--- a/src/gradualizer_lib.erl
+++ b/src/gradualizer_lib.erl
@@ -226,16 +226,13 @@ pick_binary_value(?type(binary, []), _Env) ->
     {bin, erl_anno:new(0), []};
 pick_binary_value(?type(binary, [{integer, _, 0}, {integer, _, _}]), _Env) ->
     {bin, erl_anno:new(0), []};
-pick_binary_value(?type(binary, [{integer, _, M}, {integer, _, _}]), _Env) ->
+pick_binary_value(?type(binary, [{integer, _, M}, {integer, _, _}]), _Env) when M > 0 ->
     Elements = case M rem 8 of
                    0 ->
                        S = [ $a || _ <- lists:seq(1, M div 8) ],
                        [{bin_element, 0, {string, 0, S}, default, default}];
                    _ ->
-                       Eights = [ {bin_element, 0, {integer, 0, 0}, default, default}
-                                  || _ <- lists:seq(1, M div 8) ],
-                       Tail = {bin_element, 0, {integer, 0, 0}, {integer, 0, M rem 8}, default},
-                       Eights ++ [Tail]
+                       [{bin_element, 0, {integer, 0, 0}, {integer, 0, M}, default}]
                end,
     {bin, erl_anno:new(0), Elements}.
 
--- a/src/gradualizer_lib.erl
+++ b/src/gradualizer_lib.erl
@@ -193,8 +193,10 @@ pick_value(?type(nonempty_list, Ty), Env) ->
     {cons, erl_anno:new(0), H, {nil, erl_anno:new(0)}};
 pick_value(?type(nil), _Env) ->
     {nil, erl_anno:new(0)};
-pick_value(?type(binary, _) = B, Env) ->
-    pick_binary_value(B, Env);
+pick_value(?type(binary, [{integer, _, M}, {integer, _, _}]), _Env) when M > 0 ->
+    {bin, erl_anno:new(0), [{bin_element, 0, {integer, 0, 0}, {integer, 0, M}, default}]};
+pick_value(?type(binary, _), _Env) ->
+    {bin, erl_anno:new(0), []};
 %% The ?type(range) is a different case because the type range
 %% ..information is not encoded as an abstract_type()
 %% i.e. {type, Anno, range, [{integer, Anno2, Low}, {integer, Anno3, High}]}
@@ -221,20 +223,6 @@ pick_value(Type, Env)
         not_found ->
             throw({undef, Kind, Anno, {Name, length(Args)}})
     end.
-
-pick_binary_value(?type(binary, []), _Env) ->
-    {bin, erl_anno:new(0), []};
-pick_binary_value(?type(binary, [{integer, _, 0}, {integer, _, _}]), _Env) ->
-    {bin, erl_anno:new(0), []};
-pick_binary_value(?type(binary, [{integer, _, M}, {integer, _, _}]), _Env) when M > 0 ->
-    Elements = case M rem 8 of
-                   0 ->
-                       S = [ $a || _ <- lists:seq(1, M div 8) ],
-                       [{bin_element, 0, {string, 0, S}, default, default}];
-                   _ ->
-                       [{bin_element, 0, {integer, 0, 0}, {integer, 0, M}, default}]
-               end,
-    {bin, erl_anno:new(0), Elements}.
 
 %% ------------------------------------------------
 %% Functions for working with abstract syntax trees
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3725,6 +3725,10 @@ refine_ty(?type(binary, [{integer, _, 0}, {integer, _, 8}]),
           ?type(binary, [{integer, _, 0}, {integer, _, 0}]), _, _Env) ->
     %% binary() \ <<>> => nonempty_binary()
     type(binary, [{integer, 0, 8}, {integer, 0, 8}]);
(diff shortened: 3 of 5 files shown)


zuiderkwast:
I don't really know the exact meaning of `refinable/3`. @josefs added it, for exhaustiveness-checking, but I don't think it's used for type refinement.

For refinement, we can only partially refine a type if a pattern matches a subtype of the type (e.g. a singleton so that we can subtract it from the type (e.g. `x|y \ x ==> y`). However, a free variable or `_` matches everything and this should be possible to use regardless of whether a type has singleton subtypes or not.

What I'm trying to say is that if we can make binary() refinable, we can probably remove refinable alltogether.

(Additionally, we could maybe refine by the empty binary which is a singleton: `binary() \ <<>> ==> <<_:8, _:_*8>>` (i.e. nonempty-binary).)

erszcz:
> I don't really know the exact meaning of refinable/3. @josefs added it, for exhaustiveness-checking, but I don't think it's used for type refinement.

It's not used for type refinement per se. It's used solely as a gatekeeper to exhaustiveness checking. We had a conversation about it in some other issue already. In short, "if a type is refinable we can check exhaustiveness."

> we could maybe refine by the empty binary which is a singleton: binary() \ <<>> ==> <<_:8, _:_*8>> (i.e. nonempty-binary).

Good idea :+1:

erszcz:
> We had a conversation about it in some other issue.
https://github.com/josefs/Gradualizer/pull/347

berbiche:
Related? https://github.com/josefs/Gradualizer/issues/339#issuecomment-895786061

zuiderkwast:
> Related? [#339 (comment)](https://github.com/josefs/Gradualizer/issues/339#issuecomment-895786061)

Maybe. I don't remember anymore. :-)

@erszcz Do you think everything except any() should be refineable?

erszcz:
> @erszcz Do you think everything except any() should be refineable?

I think so. From the opposite perspective, would we want any types to disable exhaustiveness checking? Ideally not, I think. Therefore everything except `any()` should be refinable. Whether it's feasible for everything except `any()` is another question, but I think we're already quite close.

zuiderkwast:
Good! So if you keep any other types as non-refineable, you can add a comment about why? (missing implementation in other parts of the code, etc.)

erszcz:
This is now ready for review. I've followed a similar pattern to that of checking empty/nonempty lists, i.e. we only check simple patterns and for complex ones exhaustiveness checking is disabled. In general making `binary()`/`String.t()` refinable should unblock exhaustiveness checking on complex types having it as a component.

`make gradualize` status is as follows - before:

```
$ wc -l gradualize.002f5f7.log
     307 gradualize.002f5f7.log
```

with this PR:

```
$ wc -l gradualize.3108bcd.log
     310 gradualize.3108bcd.log
```

The increase in self gradualization errors is caused by a new warning being emitted:

```
ebin/gradualizer_db.beam: Nonexhaustive patterns on line 608 at column 17
	match
```

This warning is now emitted since `re:run/3` present on the reported line returns complex terms containing binaries. This means we were not able to check its return value for exhaustiveness prior to this change and we are able to do so now.

erszcz:
With the latest fix we're back to

```
$ wc -l gradualize.b185fb5.log
     307 gradualize.b185fb5.log
```

src/gradualizer_db.erl
```diff
@@ -608,7 +608,9 @@ get_beam_map() ->
                 {match, [Mod]} ->
                     {true, {list_to_atom(Mod), Filename}};
                 nomatch ->
-                    false
+                    false;
+                _ ->
+                    erlang:error({unreachable, "check re:run/3 opts above - this should not happen"})
```
zuiderkwast, line 613: Debugging code? We already get a case clause error without this, so I don't think we should add this explicit code. Or is this just to silence a gradualizer error?

src/gradualizer_lib.erl
```diff
@@ -193,6 +193,8 @@ pick_value(?type(nonempty_list, Ty), Env) ->
     {cons, erl_anno:new(0), H, {nil, erl_anno:new(0)}};
 pick_value(?type(nil), _Env) ->
     {nil, erl_anno:new(0)};
+pick_value(?type(binary, _), _Env) ->
+    {bin, erl_anno:new(0), []};
```
zuiderkwast, line 197: `<<>>` is not a value of every bitstring type. The type might have a fixed or minimum length >= 0.

src/typechecker.erl
```diff
@@ -3721,6 +3721,10 @@ refine_ty(?type(nonempty_list, [Ty1]), ?type(nonempty_list, [Ty2]), Trace, Env)
         RefTy ->
             type(nonempty_list, [RefTy])
     end;
+refine_ty(?type(binary, [{integer, _, 0}, {integer, _, 8}]),
+          ?type(binary, [{integer, _, 0}, {integer, _, 0}]), _, _Env) ->
+    %% binary() \ <<>> => nonempty_binary()
+    type(binary, [{integer, 0, 8}, {integer, 0, 8}]);
```
zuiderkwast, line 3727: The `8` here can be turned into a variable N to make this a bit more generic. We just need a guard `when N > 0`.

src/typechecker.erl
```diff
@@ -4426,6 +4438,19 @@ is_list_pat_exhaustive({nil, _}) -> true;
 is_list_pat_exhaustive({cons, _, {var, _, _}, {var, _, _}}) -> true;
 is_list_pat_exhaustive(_) -> false.
 
+%% TODO: This is incomplete!
+%% To properly check pattern exhaustiveness we have to consider bound variables.
```
zuiderkwast, line 4442: A nice way to remember this is to add a known problem where we e.g. have a bound variable V in `<<..., V/binary>>`.

zuiderkwast (commented):
Can you explain in which cases we can do exhaustiveness checking for bitstring types and patterns and in which cases we can't?

For example, the following is exhaustive but I don't think we can capture it, can we?

```Erlang
-spec f(<<_:1, _:_*0>>) -> ok.
f(<<0:1>>) -> ok;
f(<<1:1>>) -> ok.
```

Do we get a false positive for the above?

What about the `<<..., _/binary>>` and `<<..., _/bitstring>>` patterns? What are the guarantees?

src/gradualizer_db.erl
erszcz: > is this just to silence a gradualizer error?

It is, exactly :+1: We don't get the warning for remaining/unhandled `re:run/3` return values anymore as we handle whatever else might be returned explicitly.

In general, I'm trying out approaches to deal with warnings that will reported by Gradualizer. A clause like this seems reasonable as it doesn't require extra options in the typechecker or pragmas in code and is also a reminder to the readers that the catch-all is intentional.

erszcz:
> Can you explain in which cases we can do exhaustiveness checking for bitstring types and patterns and in which cases we can't?

This is probably a bit flaky currently. We only do exhaustiveness checking in case of patterns covering:
- empty binaries
- nonempty binaries, that is binaries with exactly two segments (i.e. `bin_element` tuples) and the `bytes`/`binary` specifier for the last segment; first segment size is not taken into consideration

It's based on how we do it for lists.

> For example, the following is exhaustive but I don't think we can capture it, can we?
> 
> ```erlang
> -spec f(<<_:1, _:_*0>>) -> ok.
> f(<<0:1>>) -> ok;
> f(<<1:1>>) -> ok.
> ```
> 
> Do we get a false positive for the above?

We don't as the patterns are single-segment, so they disable exhaustiveness checking:

```
$ cat t.erl
-module(t).

-spec f(<<_:1, _:_*0>>) -> ok.
f(<<0:1>>) -> ok;
f(<<1:1>>) -> ok.
$ ./bin/gradualizer --verbose t.erl
...
Exhaustiveness checking: false
Popping clauses controls
Task returned from function f/1 with []
```

> What about the `<<..., _/binary>>` and `<<..., _/bitstring>>` patterns? What are the guarantees?

`bitstring` disables exhaustiveness checking. `binary` doesn't, unless the ellipsis is a single segment, as mentioned above.

src/gradualizer_db.erl
zuiderkwast: It effectively silences the warning, but it goes against good Erlang style by adding defensive code... Well well, it's partly because of re:run/3 returning different stuff depending on options.

zuiderkwast:
So anything except a few cases disables exhaustiveness checking. That is, it is best effort and we don't try to be sound regarding exhaustiveness checking in genereal. It's good to be clear about that, since we want to be sound (gradually sound, the graduality property, whatever) in the main parts of the type system.

I repeat the wish for a page about exhaustiveness checking in the https://github.com/josefs/Gradualizer/wiki.

src/typechecker.erl
```diff
@@ -4426,6 +4442,19 @@ is_list_pat_exhaustive({nil, _}) -> true;
...
+%% Pattern: <<>>
+is_bin_pat_exhaustive([]) ->
+    true;
+%% Pattern: <<_, _/bytes>>
+is_bin_pat_exhaustive([{bin_element, _, {var, _, _}, _, _},
```
zuiderkwast, line 4451: This function doesn't check if a pattern is exhaustive (what the name indicates) but rather if the pattern can be used for exhaustiveness checking or if exhaustiveness checking should be disabled.

I think we should come up with a better name for this function. (We can negate the return values or change them if it fits the function name better.)

erszcz: Yes, you're right. I was trying to avoid too long and complicated names, but it seems unavoidable if we want to capture the meaning correctly.

src/gradualizer_lib.erl
erszcz: Updated. There's no distinction between minimum and fixed length binaries when printing, but there's a special case for empty binaries.

src/typechecker.erl
erszcz: Done ✅

erszcz:
Unless you have more comments in response to today's changes, I think the only remaining thing is the wiki page. I'll create one once I find the next suitable time slot.

src/gradualizer_lib.erl
```diff
@@ -222,6 +222,22 @@ pick_value(Type, Env)
...
+                   0 ->
+                       S = [ $a || _ <- lists:seq(1, M div 8) ],
+                       [{bin_element, 0, {string, 0, S}, default, default}];
+                   _ ->
+                       Eights = [ {bin_element, 0, {integer, 0, 0}, default, default}
+                                  || _ <- lists:seq(1, M div 8) ],
+                       Tail = {bin_element, 0, {integer, 0, 0}, {integer, 0, M rem 8}, default},
+                       Eights ++ [Tail]
+               end,
+    {bin, erl_anno:new(0), Elements}.
```
zuiderkwast, line 240: Interesting, but why not simply `<<0:M>>`?

erszcz: Good point. This mislead me:

```
12> <<1:48>>.
<<0,0,0,0,0,1>>
```

But indeed, it's just:

```
13> merl:quote("<<1:48>>").
{bin,1,
     [{bin_element,1,{integer,1,1},{integer,1,48},default}]}
```

I'll simplify it.

src/gradualizer_lib.erl
```diff
@@ -220,6 +222,19 @@ pick_value(Type, Env)
...
     end.
 
+pick_binary_value(?type(binary, []), _Env) ->
+    {bin, erl_anno:new(0), []};
+pick_binary_value(?type(binary, [{integer, _, 0}, {integer, _, _}]), _Env) ->
+    {bin, erl_anno:new(0), []};
+pick_binary_value(?type(binary, [{integer, _, M}, {integer, _, _}]), _Env) when M > 0 ->
+    Elements = case M rem 8 of
+                   0 ->
+                       S = [ $a || _ <- lists:seq(1, M div 8) ],
```
zuiderkwast, line 232: Good to simplify. Is there a need for `<<"aaaa...">>` or can we just use `<<0:M>>` here too? Is there is a nonempty string type in Elixir that would benefit from picking a string-looking value?

I think the minimal binary-picking would be just two clauses and no helper function needed, but it might have missed something.

```Erlang
pick_value(?type(binary, [{integer, _, M}, {integer, _, _}]), _Env) when M > 0 ->
    {bin, erl_anno:new(0), [{bin_element, 0, {integer, 0, 0}, {integer, 0, M}, default}]};
pick_value(?type(binary, _), _Env) ->
    {bin, erl_anno:new(0), []};
```

src/typechecker.erl
erszcz: The latter actually leads to:

```
t2.erl: The variable on line 13 at column 11 is expected to have type non_neg_integer() but it has type <<_:_*8>>
```

but indeed, they don't lead to any exhaustiveness false positives, which is a good thing.

> since the real logic is in type_diff?

For the case when `should_bin_pat_enable_exhaustiveness_check()` returns true and the examples I could think of or you provide - yes. If `should_bin_pat_enable_exhaustiveness_check()` returns false we disable exhaustiveness checking for the clauses list completely, therefore avoid any false positives.

---

Ultimately, because of https://github.com/josefs/Gradualizer/pull/392#issuecomment-1145871376 we could just disable exhaustiveness checking for binary patterns completely (i.e. keep `refinable(binary()) == true`, but disable ex-check each time we see a binary pattern), and the goal to be able to include them in complex types would be met anyway. That would be as simple as possible, but the tradeoff would be that even the simplest binary patterns (for example an empty binary) would disable the ex-check on the complex type, too.

src/gradualizer_lib.erl
zuiderkwast: Btw, I'd use `rem` in a guard, so the case clauses can be turned into function clauses, i.e. `when M > 0, M rem 8 =:= 0`.

zuiderkwast: Oh, you removed this. :-) Well, I'll leave it to you to decide if you want the `<<"aaa...">>` back or not. I don't mind, just want to understand the use cases.

src/typechecker.erl
zuiderkwast: Thanks for the explanation.

Btw, I think my g/2 example could easily be corrected, e.g. `<<Y:1/binary, _/binary>>`.

zuiderkwast: If you leave out the pattern `<<>>` in the examples `f/2` and the corrected `g/2`, you'd get "Values not covered: `<<>>`" and when you add a clause for `<<>>`, the error dissapears, which can make the user believe all values are covered. Hopefully, they have RTFM and found that exhaustiveness-checking is best effort (i.e. unsound) anyway, so all is fine. :-)

zuiderkwast (approved):
Feel free to merge when you have decided which variant of pick value you prefer.

zuiderkwast:
Nice! But I miss the IMO most important info about the guarantees/soundness.

PR #393, opened 2022-03-18
fix: pp_type 'var' always assume the var is a string

berbiche:
A 'var' type variable's name is not always a string as opposed to what the comment indicate in `pp_type`.

This commit fixes the logic to handle both cases in `pp_type`.

This commit also removes the final period in one error message because most error messages don't have a
final period.

---

The odd thing about this bug is that I hit this case through `format_type_error({type_error, record_pattern, ...`:
> /app/src/censored: The record pattern for record #censored on line 707 at column 19 is expected to have type A

where `A` is typed as a `var` because it stems from a `lists:map()`.

The error message should be improved for generics if it's possible.

### DIFF
--- a/src/typelib.erl
+++ b/src/typelib.erl
@@ -47,8 +47,11 @@ pp_type(Type = {type, _, bounded_fun, _}) ->
                           [{capture, all_but_first, list}, dotall]),
     "fun(" ++ S ++ ")";
 pp_type({var, _, TyVar}) ->
-    %% TODO: In type(), TyVar should be an atom but we use a string.
-    TyVar;
+    %% See gradualizer_type:af_type_variable/0 and typechecker:new_type_var/0
+    if
+        is_atom(TyVar) -> atom_to_list(TyVar);
+        is_list(TyVar) -> TyVar
+    end;
 pp_type(Type) ->
     %% erl_pp can handle type definitions, so wrap Type in a type definition
     %% and then take the type from that.
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -303,7 +303,7 @@ format_type_error({type_error, check_clauses}, _Opts) ->
 format_type_error({type_error, record_pattern, Anno, Record, Ty}, Opts) ->
     io_lib:format(
       "~sThe record pattern for record #~p~s is expected to have"
-      " type ~s.~n",
+      " type ~s~n",
       [format_location(Anno, brief, Opts),
        Record,
        format_location(Anno, verbose, Opts),


src/typelib.erl
```diff
@@ -48,7 +48,10 @@ pp_type(Type = {type, _, bounded_fun, _}) ->
     "fun(" ++ S ++ ")";
 pp_type({var, _, TyVar}) ->
     %% TODO: In type(), TyVar should be an atom but we use a string.
-    TyVar;
+    case is_atom(TyVar) of
+        true -> atom_to_list(TyVar);
+        false -> TyVar
+    end;
```
zuiderkwast, line 54: Maybe it's a matter of preference, but I prefer guards. This also avoids catch-all. (Better crash on unexpected data.)

```suggestion
    if
        is_atom(TyVar) -> atom_to_list(TyVar);
        is_list(TyVar) -> TyVar
    end;
```

zuiderkwast (commented):
Good.

Do you have a minimal example to trigger this? I suppose it needs the infer flag...

berbiche:
Sure, I added a test case in the commit message which I inline here:

``` erlang
-module(error).

-compile([nowarn_export_all, export_all]).

-record(some_record, { id :: integer() }).

-spec error([#some_record{}]) -> [integer()].
error(List) ->
    lists:map(fun(#some_record{id = Id}) -> Id end, List).
```

``` console
$ ./gradualizer error.erl
error.erl: escript: exception error: bad argument
  in function  io_lib:format/2
     called as io_lib:format("~sThe record pattern for record #~p~s is expected to have type ~s.~n",
                             [[],some_record,
                              [32,111,110,32,108,105,110,101,32,"9",32,97,116,
                               32,99,111,108,117,109,110,32,"19"],
                              ["\e[35m",'A',"\e[0m"]])
  in call from gradualizer_fmt:handle_type_error/2 (src/gradualizer_fmt.erl, line 534)
  in call from gradualizer_fmt:'-print_errors/2-lc$^0/1-0-'/2 (src/gradualizer_fmt.erl, line 520)
  in call from gradualizer_fmt:print_errors/2 (src/gradualizer_fmt.erl, line 520)
  in call from gradualizer:type_check_forms/3 (src/gradualizer.erl, line 209)
  in call from gradualizer:'-type_check_files/2-fun-1-'/4 (src/gradualizer.erl, line 159)
  in call from lists:foldl/3 (lists.erl, line 1267)
  in call from gradualizer_cli:main/1 (src/gradualizer_cli.erl, line 14)
```

src/typelib.erl
berbiche: Thanks!

erszcz:
Looks good 👍  I did the same change in #359, specifically lines https://github.com/josefs/Gradualizer/pull/359/files#diff-24fdf073a8645c401774bfbe51791d5c0c5f4aaecf41856284451c8498fbe938L50-R54.

src/typelib.erl
```diff
@@ -48,7 +48,10 @@ pp_type(Type = {type, _, bounded_fun, _}) ->
     "fun(" ++ S ++ ")";
 pp_type({var, _, TyVar}) ->
     %% TODO: In type(), TyVar should be an atom but we use a string.
-    TyVar;
+    if
```
erszcz, line 51:
```suggestion
    %% See gradualizer_type:af_type_variable/0 and typechecker:new_type_var/0
    if
```

berbiche: Thanks!

erszcz:
@berbiche Could you add the test as `test/should_pass/pp_type_error.erl`? I think this is good to merge, but let's have the test added to catch possible future regressions.

zuiderkwast:
> shouldn't it be under `test/known_problems/should_pass/`

You're right, but IIRC known problems even accepts crashes. If that is the case, then I think we need change that and mark the crashing examples in another way, e.g. move them to known_problems/crashes etc. (if there are any)

erszcz:
@berbiche We can prepare an ordinary unit test for this, not necessarily use the should pass / should fail framework. What do you think of https://github.com/josefs/Gradualizer/pull/401/commits/0083d1311b917551331c1ee295f2fd485710409c? We could add it to this PR and then merge it.

Issue #390, opened 2022-02-25
Not working with mix/elixir


lukaszsamson:
I don't know if mix/elixir support is the goal but Readme mentions gradualixir as a way to run gradualizer with mix. That solution no longer works and the autor doesn't seem to be actively developing gradualixir anymore (see [issue](https://github.com/OvermindDL1/gradualixir/issues/4))

zuiderkwast:
I you figure it out, feel free to remove the references to gradualixir and replace it with a better description on how to use it with Elixir. I don't use Elixir and neither does Josef AFAIK. Maybe @erszcz does?

erszcz:
Indeed, I do. We also work on https://github.com/esl/gradient, which is experimental, but allows to leverage Gradualizer for Elixir codebases. [Gradualixir author admitted he's no longer working on the project](https://elixirforum.com/t/gradualyzer-vs-dialyzer-main-differences/27343/13?u=erszcz), which likely makes Gradient the only alternative.

erszcz:
As of merging #396 the README now points at #esl/gradient. @lukaszsamson do you have any other suggestion on improving the experience for Elixir users? I'll close this for now, but feel free to reopen if there's more to discuss.

PR #396, opened 2022-03-23
Point README at Gradient instead of discontinued Gradualixir

erszcz:
As mentioned in #390 Gradualixir is discontinued by its main author. However, we're actively working on [Gradient](https://github.com/esl/gradient), which is an Elixir frontend to Gradualizer. I think it won't hurt to point at it from the main README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,11 +38,14 @@ use the functions in the `gradualizer` module.
 
     1> gradualizer:type_check_file("path/to/some_file.erl").
 
-## To run Gradualizer from rebar3 or Mix
+## To run Gradualizer from rebar3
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,11 +38,28 @@ use the functions in the `gradualizer` module.
...
+
+Please check out [Gradient](https://github.com/esl/gradient), the Elixir frontend for Gradualizer.
+It provides a Mix task to use in your project.
+
+Quickstart:
+
+```elixir
+def deps do
+  [
+    {:gradient, github: "esl/gradient", only: [:dev], runtime: false}
```
zuiderkwast, line 55: Isn't it better that users follow the link to gradient and find these examples there?

Issue #395, opened 2022-03-23
Unbound variable in presence of tuple pattern in function head binding it


erszcz:
This is a minimised example of one of the problems from #373 crashing Gradualizer on the Aeternity codebase.

```
12:56:53 erszcz @ x6 : ~/work/aeternity/aeternity (gradualizer-try-out *%)
$ cat apps/aechannel/src/aesc_chain_watcher.erl
-module(aesc_chain_watcher).

-record(st, {}).

handle_info({'DOWN', _MRef, process, Pid, _}, #st{} = St) ->
    ets:delete('?T_MREFS', Pid),
    {noreply, St}.
12:56:55 erszcz @ x6 : ~/work/aeternity/aeternity (gradualizer-try-out *%)
$ gradualizer --version
Gradualizer v0.1.3-137-g6e89b4e
12:56:58 erszcz @ x6 : ~/work/aeternity/aeternity (gradualizer-try-out *%)
$ gradualizer -- apps/aechannel/src/aesc_chain_watcher.erl
escript: exception error: bad key: 'Pid'
  in function  map_get/2
     called as map_get('Pid',#{'St' => {type,0,any,[]}})
     *** argument 1: not present in map
  in call from typechecker:do_type_check_expr_in/3 (src/typechecker.erl, line 2246)
  in call from typechecker:type_check_expr_in/3 (src/typechecker.erl, line 2239)
  in call from typechecker:'-type_check_call_ty/4-lc$^0/1-0-'/2 (src/typechecker.erl, line 2055)
  in call from typechecker:'-type_check_call_ty/4-lc$^0/1-0-'/2 (src/typechecker.erl, line 2056)
  in call from typechecker:type_check_call_ty/4 (src/typechecker.erl, line 2056)
  in call from typechecker:do_type_check_expr/2 (src/typechecker.erl, line 1615)
  in call from typechecker:type_check_expr/2 (src/typechecker.erl, line 1493)
12:57:01 erszcz @ x6 : ~/work/aeternity/aeternity (gradualizer-try-out *%)
$
```

I suspect something fishy going on in the tuple clause of `add_type_pat`.

erszcz:
This seems to be a duplicate of or at least very similar to #374.

erszcz:
More info:

```
$ cat apps/aechannel/src/aesc_chain_watcher.erl
-module(aesc_chain_watcher).

-record(st, {}).

handle_info({Pid}, #st{}) ->
[...]

Issue #373, opened 2022-01-17
Running on the Aeternity codebase


seanhinde:
Hi team. I had a go at running gradualizer against the aeternity codebase. It has a decent number of type specs and passes dialyzer cleanly. A few modules cause a crash, and there are lots of warnings.

https://github.com/aeternity/aeternity/tree/gradualizer-try-out

Just posting here in case you are after a decently large example to play with

/Sean

PR #397, opened 2022-03-23
Fix variable bind before a record pattern match in function head

erszcz:
Fix and test for #395. This is one of the issues found thanks to #373.


erszcz:
Reopening as GH Actions didn't kick in.

zuiderkwast (approved):

PR #401, opened 2022-03-23
Fix pp_type({var, ..., _})

erszcz:
This adds a unit test and supersedes #393. Thanks @berbiche!


erszcz:
Hmm, GH actions don't kick in...

zuiderkwast (approved):

PR #403, opened 2022-03-31
Salvage exhaustiveness checking improvements

erszcz:
This brings in tests, refactorings, and documentation improvements from #359, but doesn't introduce any functional changes.


zuiderkwast (approved):
This looks familiar. :-) I don't see any surprises. I think it has already been reviewed and discussed before.

erszcz:
Indeed :)

Issue #97, opened 2018-11-23
Intersection types in the type language


UlfNorell:
A function can have multiple specs, which [the documentation](http://erlang.org/doc/reference_manual/typespec.html#specifications-for-functions) calls overloading and whose semantics is intersection types. For instance,
```erlang
-spec foo(boolean()) -> its_a_bool;
         (integer()) -> its_an_int.
```
means that `foo` has both the type `boolean() -> its_a_bool` and the type `integer() -> its_an_int`.

We currently handle this for function definitions and explicit calls using the `fun_ty_intersection` return value from `expect_fun_type`, but not for `fun F/A` and `fun M:F/A` expressions. To do this we'd need to be able to represent intersection types in the type language.

zuiderkwast:
Yeah! So, what's the way forward?

Ideally, we convince the OTP team to include the intersection type in the official type language. My colleage Karol says that he has reported this years back but it has not been prioritized.

Until then, we can add it in our type, which we could export in the main interface module, i.e. `gradualizer:type()`. Then we just need to implement pretty-printing for it as a special case and handle it in the typechecker. I don't see that we need to do anything else. Do you?

Then, how should it be represented?

I suggest `{type, erl_anno:anno(), intersection, [type()]}`.

For the syntax (pretty-printing, documentation, etc.), I suggest the semicolon `T1 ; T2`. Can it only be used for fun types or do we for example ever need to intersect with `any()`? If it's only for fun types, I think the first of these syntaxes would be nice; otherwise the second:

```Erlang
-type t() :: fun((boolean()) -> its_a_bool;
                 (integer()) -> its_an_int).
-type t() :: fun((boolean()) -> its_a_bool);
             fun((integer()) -> its_an_int).
```

If we want to allow `T1 ; T2` for any types, is there any risk of syntax ambiguity when mixing specs with intersection types?

```Erlang
-spec foo(boolean() | atom()) -> fun((boolean() -> its_a_bool) ;
                                 fun((atom()) -> its_an_atom) ;
         (integer()) -> its_an_int.
```

gomoripeti:
I think intersection with `any()` could be temporary solution for handling conditional expressions before refinement types are fully implemented.

for exampe
```
  case get_int() of
     PosInt when PosInt > 0 -> handle_pos_int(PosInt);
    _ -> ...
  end.
-spec get_int() -> integer().
-spec handle_pos_int(pos_integer()) -> any().
```
[...]

PR #404, opened 2022-04-02
Fix list exhaustiveness checking regressions

erszcz:
This is another take at #359 and it finally seems to be successful. Apparently, I was confused with recursion in `normalize` and fixed on it too much, while ultimately the problem lay somewhere else - it was about quite subtle bugs in refinement.


erszcz:
@berbiche Could you check if this fixes #351-like issues in the project you mentioned there?

berbiche:
Sorry, I'll test this PR sometimes tomorrow (Friday).

src/typechecker.erl
```diff
@@ -2540,47 +2553,45 @@ do_type_check_expr_in(Env, ResTy, Expr = {'fun', P, {function, M, F, A}}) ->
...
-    {VarBinds2, Cs2} = type_check_expr_in(Env
-                                         ,type(integer)
-                                         ,After),
-    {VarBinds3, Cs3} = type_check_block_in(Env, ResTy, Block),
-    {union_var_binds([VarBinds1, VarBinds2, VarBinds3], Env)
-                    ,constraints:combine([Cs1, Cs2, Cs3])};
+    {Env1, Cs1} = check_clauses(Env, [type(any)], ResTy, Clauses, capture_vars),
+    {Env2, Cs2} = type_check_expr_in(Env1, type(integer), After),
+    {Env3, Cs3} = type_check_block_in(Env2, ResTy, Block),
+    {union_var_binds([Env1, Env2, Env3], Env), constraints:combine([Cs1, Cs2, Cs3])};
```
zuiderkwast, line 2594: I don't think this is correct. The variables bound in Clauses are not available in After (the timeout expression) nor the after Block. 

The vars bound in After are available in Block though. That's a bug in the legacy code. (I was unsure so I had to test it in an erl shell.)

Example:

```Erlang
receive
    V when is_atom(V) ->
        atom_to_list(V)
after (begin V=42, V+1000 end) ->     % a different V
    integer_to_list(V)
end.
```

src/typechecker.erl
```diff
@@ -4069,41 +4105,56 @@ position_info_from_spec(Form) ->
...
 %% is the singleton type. Otherwise, PatTy is none(). PatTy is a type exhausted
 %% by Pat. UBound is Ty or a subtype such that Pat :: UBound.
+-spec add_types_pats(Pats, Tys, Env, Caps) -> R when
+      Pats :: [gradualizer_type:abstract_pattern()],
+      Tys  :: [type()],
+      Env  :: env(),
+      Caps :: capture_vars | bind_vars,
+      R :: {PatTys      :: [type()],
+            UBounds     :: [type()],
+            NewVEnv     :: env(),
```
zuiderkwast, line 4115: ?

```suggestion
            NewEnv      :: env(),
```

src/typechecker.erl
```diff
@@ -4069,41 +4105,56 @@ position_info_from_spec(Form) ->
...
-do_add_types_pats(Pats, Tys, Env, VEnv) ->
-    add_types_pats(Pats, Tys, Env, VEnv, [], [], []).
+%% NB: Don't use this function directly. Use add_types_pats/4 instead.
+-spec do_add_types_pats(Pats, Tys, Env) -> R when
+      Pats :: [gradualizer_type:abstract_pattern()],
+      Tys  :: [type()],
+      Env  :: env(),
+      R :: {PatTys      :: [type()],
+            UBounds     :: [type()],
+            NewVEnv     :: env(),
```
zuiderkwast, line 4131:
```suggestion
            NewEnv      :: env(),
```

src/typechecker.erl
```diff
@@ -4165,87 +4217,95 @@ add_type_pat(Pat, ?type(union, UnionTys)=UnionTy, Env, VEnv) ->
...
+            {PatTy1, _UBound1, Env2, Cs2} =
+                add_type_pat(PH, normalize(ElemTy, Env), Env),
             TailTy = normalize(type(union, [ListTy, type(nil)]), Env),
-            {_PatTy2, _Ubound2, VEnv3, Cs3} = add_type_pat(PT, TailTy, Env, VEnv2),
+            {_PatTy2, _Ubound2, Env3, Cs3} = add_type_pat(PT, TailTy, Env2),
+            {PatTy, Env4} = case is_list_pat_exhaustive(CONS) of
+                                true ->
+                                    {type(nonempty_list, [PatTy1]), Env3};
+                                false ->
+                                    {type(none), Env3#env{exhaust = false}}
```
zuiderkwast, line 4294: Where do you set exhaust back to what it was before? Does this disable exhaustiveness checking for the rest of the module?

In a nested pattern (or nested expression) think we want to disable exhaustiveness checking only for the particular list where we don't have an exhaustive pattern in the clauses. Below, we still want it for the tuple and for ANotherList, don't we?

```Erlang
case List of
    {[X, Y | Tail], foo} ->
        case ANotherList of [] -> empty; [_|_] -> nonempty end;
    {[X], foo} ->
        ...
    {[], foo} ->
        ...
end
```

zuiderkwast (commented):
Interesting and large! I didn't proof-read everything completely, but I hopefully spotted the relevant parts.

What I had in mind for disabling exhaustiveness-checking was merely to add another element to the return tuple of add_type_pat (i.e. `{_, _, _, _, DisableExhaust}`). But this solution probably works too.

src/typechecker.erl
```diff
@@ -1612,9 +1605,11 @@ do_type_check_expr(Env, {call, _, {atom, _, TypeOp}, [Expr, {string, _, TypeStr}
     end;
 do_type_check_expr(Env, {call, _, {atom, _, record_info}, [_, _]} = Call) ->
     Ty = get_record_info_type(Call, Env),
-    {Ty, #{}, constraints:empty()};
+    {Ty, Env, constraints:empty()};
```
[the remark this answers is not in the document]
erszcz: This response is also connected with your other one from above, so I'll respond to both here:

> What I had in mind for disabling exhaustiveness-checking was merely to add another element to the return tuple of add_type_pat (i.e. {_, _, _, _, DisableExhaust}).

> Arguably, it's a more functional style to pass as little data as possible to and from each function.

The thing is this value is not used immediately after return from `add_type_pat`. Possibly, it could even be only used after return from an N-level nested `add_type_pat` call, not the topmost one. Then, it has to be returned up back to `check_clauses`. Would it make sense to modify each of these to return the 5th tuple element? I think that with the current change, even though it is bigger in scope, the code is more future-proof, as whatever we might need to store/change in `Env` as we type check will be available in all the subsequent code.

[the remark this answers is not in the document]
zuiderkwast: Right, it affects all outer patterns! I didn't think of that.

Why does it affect expressions though?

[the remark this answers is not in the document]
zuiderkwast: > Then, it has to be returned up back to `check_clauses`.

How do we distinguish between levels here? We have exhaustiveness-checking not only for function cases but also for case, receive and try clauses. We want to disable it only for the current "clause-related" expression, not for e.g. other (inner or outer) case expressions. Or am I missing something?

src/typechecker.erl
erszcz: > Where do you set exhaust back to what it was before? Does this disable exhaustiveness checking for the rest of the module?

That's a bit tricky. I don't yet, but the next form will be checked with a different copy of `Env` where `exhaust` is `true` by default:
https://github.com/josefs/Gradualizer/blob/3cda2e5aa1b1313b3e02fce67986fcda3ab9a02c/src/typechecker.erl#L4892
In other words, the current toggle disables exhaustiveness checking "for the rest of the current top-level form".

Based on your example, if we toggle `exhaust` to `true` when leaving `check_clauses`:

```erlang
case List of
    {[X, Y | Tail], foo} -> %% we set exhaust=false here
        case ANotherList of %% we won't check exhaustiveness on the nested case expr
            [] -> empty;
            [_|_] -> nonempty
        end; %% we exit from check_clauses -> set exhaust=true
    {[X], foo} -> %% we have exhaust=true, even though we wanted exhaust=false and did set it on the previous clause of this case expr
        ...
    {[], foo} ->
        ...
end
```

We could probably keep a stack of `exhaust` values in `Env` and push the previous value when entering `check_clauses` and pop when exiting 🤔  WDYT?

zuiderkwast: Stack! :+1: :100: I was thinking about that too. We don't need to change exhaust=true (the user's option) but instead we can use a different field (`clause_stack = [#clause{exhaust=true}|Clauses]`). WDYT?

erszcz: OK, this is done now and seems to work well :+1:

src/typechecker.hrl
```diff
@@ -1,13 +1,17 @@
...
+
 -record(env, {fenv              = #{}   :: map(),
               imported          = #{}   :: #{{atom(), arity()} => module()},
               venv              = #{}   :: typechecker:venv(),
               tenv                      :: gradualizer_lib:tenv(),
               infer             = false :: boolean(),
               verbose           = false :: boolean(),
               exhaust           = true  :: boolean(),
+              %% Options tunable per clause block (fun/case/try-catch/receive).
+              blocks            = []    :: [#clause_block{}],
```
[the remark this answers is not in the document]
erszcz: I think most programming languages have the concept of a code block, e.g. in C (almost?) everywhere where we can use a single expression we can also use a `{ ... }` delimited block.

> Maybe clause_stack is a better name here?

I was thinking of that and also wondering about `clauses_stack`, but ultimately decided not to make it more complicated than necessary. `clause_stack` (singular clause) is not correct, because the stack element is pushed/popped not per clause, but per a group of all clauses making up the higher-level expression (fun, case, try ... catch, or receive) - I think it would be misleading. I was also considering a `form_stack`, but that's not correct either, since we actually have multiple levels in a top-level form, and in general forms are more general than clauses or clause blocks.

The type checker code uses `Clauses` (plural) to name the concept for which we push/pop the things. However, using a plural noun to describe other nouns depending on it makes room for ambiguity (does the description refer to a single instance or to the grouping?). I think that "code block" might not be perfect, but IMO is good enough :)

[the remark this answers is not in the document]
erszcz: BTW, I'll fix the comment, you're right it's a bit misleading.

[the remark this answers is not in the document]
zuiderkwast: In C and Elixir, a block is always `{...}` and `do-end` respectively (regardless of whether it's a list of statements `{A; B; C;}` or a `while (...) {...}` or `switch {...}`), so it makes some sense.

In C, you can also have a sequence of expressions, delimited by comma as in Erlang, but that's not a block since there are no braces. Example: `(i = 0, j = 42, i + j)`.

In Erlang, the're so do-end; there's different syntax for functions, expressions, clauses, etc. More importantly, the word "block" is not used to describe anything in Erlang grammar in the official docs.

I think `clauses_stack` is perfect, especially if "clauses" is used in the grammar docs. :-)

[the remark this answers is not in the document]
erszcz: Ok, good point that it's used like that in the grammar. I'll change it :+1:

src/typechecker.erl
erszcz: Hmm... So I checked again:

> The vars bound in After are available in Block though.

They don't seem to be:

```erlang
1> receive
1>     V when is_atom(V) ->
1>         atom_to_list(V)
1> after (begin V=42, V+1000 end) ->     % a different V
1>     integer_to_list(V)
1> end.
* 5:21: variable 'V' is unbound
```

Maybe it depends on the Erlang version? Though it would be quite surprising 🤔

For now, I've pushed a fix which makes each of `Clauses`, `After`, and `Block` independent.

src/typechecker.hrl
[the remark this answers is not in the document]
erszcz: Fixed, WDYT?

src/typechecker.erl
```diff
@@ -3364,51 +3395,75 @@ check_clauses_union(Env, [Ty|Tys], Clauses) ->
...
+    {VarBindsList, Css, RefinedArgsTy, Env2} =
+        lists:foldl(fun (Clause, {VBs, Css, RefinedArgsTy, EnvIn}) ->
+                            {NewRefinedArgsTy, Env2, Cs} =
+                                check_clause(EnvIn, RefinedArgsTy, ResTy, Clause, Caps),
+                            VB =
+                                refine_vars_by_mismatching_clause(Clause, Env2#env.venv, Env2),
+                            {[Env2#env{venv = VB} | VBs], [Cs | Css], NewRefinedArgsTy, Env2}
                     end,
-                    {[], [], ArgsTy, Env#env.venv},
+                    {[], [], ArgsTy, Env1},
```
zuiderkwast, line 3440: I think this is another case where variables bound in one clause survive into the next one, because they're added to Env2 returned by check_clause, which is passed on to the next clause.

Example:

```Erlang
-spec f(integer()) -> integer().
f(X) ->
    case X of
        Y when Y > 42 -> Y;
        Y             -> Y + 1  % a different Y
    end.
```

Solution: Don't add varbinds directly in the returned Env2, alternatively restore venv within Env2 before passing it to the next clause.

src/typechecker.erl
```diff
@@ -4165,87 +4240,95 @@ add_type_pat(Pat, ?type(union, UnionTys)=UnionTy, Env, VEnv) ->
         _SomeTysMatched ->
             %% TODO: The constraints should be merged with *or* semantics
             %%       and var binds with intersection
-	    {Ty, Cs} = glb(PatTys, Env),
+            {Ty, Cs} = glb(PatTys, Env),
+            NewEnv = union_var_binds(Envs, Env),
             {Ty,
              normalize(type(union, UBounds), Env),
-             union_var_binds(VEnvs, Env),
+             NewEnv,
```
zuiderkwast, line 4247: Why this change?

src/typechecker.erl
```diff
@@ -4339,12 +4422,16 @@ add_type_pat(OpPat = {op, _Anno, _Op, _Pat1, _Pat2}, Ty, Env, VEnv) ->
...
-    add_type_pat_literal(OpPat, Ty, Env, VEnv);
-add_type_pat(Pat, Ty, _Env, _VEnv) ->
+    add_type_pat_literal(OpPat, Ty, Env);
+add_type_pat(OpPat = {op, _Anno, _Op, _Pat}, Ty, Env) ->
+    add_type_pat_literal(OpPat, Ty, Env);
+add_type_pat(Pat, Ty, _Env) ->
     throw({type_error, pattern, element(2, Pat), Pat, Ty}).
 
+is_list_pat_exhaustive({nil, _}) -> true;
+is_list_pat_exhaustive({cons, _, {var, _, _}, {var, _, _}}) -> true;
```
zuiderkwast, line 4432: `[X|Y]` only exhausts a nonempty list type if X and Y are unbound and distinct. Need to check that. Or does this function just check that a pattern doesn't disable exhaustiveness checking?

src/typechecker.erl
```diff
@@ -4165,87 +4240,95 @@ add_type_pat(Pat, ?type(union, UnionTys)=UnionTy, Env, VEnv) ->
...
+            {PatTy1, _UBound1, Env2, Cs2} =
+                add_type_pat(PH, normalize(ElemTy, Env), Env),
             TailTy = normalize(type(union, [ListTy, type(nil)]), Env),
-            {_PatTy2, _Ubound2, VEnv3, Cs3} = add_type_pat(PT, TailTy, Env, VEnv2),
+            {_PatTy2, _Ubound2, Env3, Cs3} = add_type_pat(PT, TailTy, Env2),
+            {PatTy, Env4} = case is_list_pat_exhaustive(CONS) of
+                                true ->
+                                    {type(nonempty_list, [PatTy1]), Env3};
+                                false ->
+                                    {type(none), disable_exhaustiveness_check(Env3)}
```
zuiderkwast, line 4317: Do we really always need to disable exhaustiveness-checking when a pattern is nonexhaustive? For example we could give an error for f below, but for g we can't know.

```Erlang
-spec f(string()) -> ok;
f([$a]) -> ok;
f([$a | _]) -> ok.

-spec g(string()) -> ok;
g([]) -> ok;
g([_] -> ok;
g([_, _ | _]) -> ok.
```

zuiderkwast (commented):
The clauses-stack looks good, but I took a closer look and found some more things that might need to be fixed. (I may be wrong in some cases though.)

src/typechecker.erl
```diff
@@ -2588,10 +2588,11 @@ do_type_check_expr_in(Env, ResTy, {'receive', _, Clauses}) ->
     {Env1, Cs} = check_clauses(Env, [type(any)], ResTy, Clauses, capture_vars),
     {Env1, Cs};
 do_type_check_expr_in(Env, ResTy, {'receive', _, Clauses, After, Block}) ->
+    VB = Env#env.venv,
```
zuiderkwast, line 2591: VB is short for var binds, i.e. variables bound in a the checked piece of code.

I think VEnv is a better name for this variable.

src/typechecker.erl
erszcz: There's definitely room for improvement here, but I'd prefer going in small steps, i.e. in subsequent PRs.

src/typechecker.erl
erszcz: > Or does this function just check that a pattern doesn't disable exhaustiveness checking?

Exactly, we use this to disable exhaustiveness checking for the current clauses list. We could definitely make it more sophisticated, but with the current tests this is enough.

src/typechecker.erl
erszcz: Hmm, good catch. Thanks!

erszcz:
Thanks for a thorough review, @zuiderkwast! Indeed, with the `VEnv` -> `Env` changes I mostly focused on making sure the code builds and passes tests, but there might be some mistakes with passing `Env` around. It's good to have a second pair of eyes catching those!

src/typechecker.erl
zuiderkwast: OK, so in the following example, Gradualizer thinks that the case clauses are exhaustive and doesn't return an error, but it could as well have disabled exhaustiveness checking. For the exhaustiveness checking, there's no difference, but for refinement there might be a difference. Maybe we get a warning for unreachable clauses?

```Erlang
-spec f([atom()]) -> ok.
f(X) ->
    case X of
        [] -> ok;
        [X|_] -> ok     % <--- Not exhaustive, but Gradualizer thinks it is.
        [hello|_] -> ok % <--- False positive: clause not reachable?
    end.
```

zuiderkwast: This function is not actually checking if the list pattern is exhaustive. We can leave it as is for now, but then at least rename the function or add a comment about it. It can be hard to understand this in the future otherwise.

I think it's only called with CONS patterns, so the first clause is not used AFAICT.

src/typechecker.erl
zuiderkwast: The example above might not show anything in a test case. Maybe this is better, where Y are of different types:

```Erlang
-spec f(foo | integer()) -> integer().
f(X) ->
    case X of
        Y = foo -> length(atom_to_list(Y));
        Y       -> Y + 1  % a different Y
    end.
```

src/typechecker.erl
```diff
@@ -3340,17 +3371,17 @@ infer_clause(Env, {clause, _, Args, Guards, Block}) ->
...
-    {#{}, constraints:empty()};
+check_clauses_intersect(Env, [], _Clauses) ->
+    {Env, constraints:empty()};
 check_clauses_intersect(Env, [Ty|Tys], Clauses) ->
-    {VarBinds1, Cs1} = check_clauses_fun(Env, Ty, Clauses),
-    {VarBinds2, Cs2} = check_clauses_intersect(Env, Tys, Clauses),
-    {union_var_binds(VarBinds1, VarBinds2, Env), constraints:combine(Cs1, Cs2)}.
+    {Env1, Cs1} = check_clauses_fun(Env, Ty, Clauses),
+    {Env2, Cs2} = check_clauses_intersect(Env1, Tys, Clauses),
+    {union_var_binds(Env1, Env2, Env), constraints:combine(Cs1, Cs2)}.
```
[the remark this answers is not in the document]
erszcz: I'm fixing it, but the proposed test fails with an unrelated error (also on current master):

```
$ cat clauses_intersect.erl
-module(clauses_intersect).

-export([foo/1]).

-spec foo(bar) -> bar; (baz) -> baz.
foo(X = bar) -> X;
foo(X = baz) -> X.
$ ./bin/gradualizer --verbose clauses_intersect.erl
Checking module clauses_intersect
Spawning async task...
Checking function foo/1
Pushing {clauses_controls,true}
6:1: Checking clause :: bar
6:17: Checking that X :: bar
Task returned from function foo/1 with [{type_error,unreachable_clause,{7,1}}]
clauses_intersect.erl: The clause on line 7 at column 1 cannot be reached
```

Actually, now that I look at it I think `check_clauses_intersect` is buggy (or just incomplete? I think there's a comment somewhere in the code which suggests function intersections are not fully supported yet). Take an example like this:

```
-spec foo(bar) -> bar;
      foo(list()) -> baz | qux.
foo(bar) -> bar;
foo([]) -> baz;
foo([a]) -> qux;
foo([_, _ | T]) -> foo(T).
```

The question is: after how many fun clauses should we switch from checking against `foo(bar) -> bar` to `foo(list()) -> baz | qux`? Well, once our fun clauses exhaust `bar`. This is not done currently, hence the error returned when checking your example. We should probably do something more like this:

```erlang
try
    {Env1, Cs1} = check_clauses_fun(Env, Ty, Clauses),
catch
    throw:{type_error,unreachable_clause,...} ->
        %% Now try to type check the fun clauses against the next spec clause.
end
```

All in all, though, I think this is outside the scope of this PR.

src/typechecker.erl
erszcz: Probably just a debugging artefact. Reverting.

src/typechecker.erl
erszcz: Hmm, interesting. I saved this example as a test and was assuming we're done, but...

```erlang
$ cat test/should_pass/variable_binding_leaks.erl
-module(variable_binding_leaks).

-export([case_clauses/1]).

-spec case_clauses(foo | integer()) -> integer().
case_clauses(X) ->
    case X of
        Y = foo -> length(atom_to_list(Y));
        Y       -> Y + 1  % a different Y
    end.
$ cat test/should_pass/type_refinement.erl
-module(type_refinement).

-spec refine_bound_var_by_guard_bifs(ok | integer() | pid() |
                                     #{x => y} | {x, y}) -> ok.
refine_bound_var_by_guard_bifs(X) ->
    if
        is_map(X)     -> ok;
        is_tuple(X)   -> ok;
        is_integer(X) -> ok;
        is_pid(X)     -> ok;
        true          -> X
    end.
```

The former requires vars to be separate between clauses, the latter depends on refinement by mismatching clause, so the particular var should be updated and propagated 🤷  It needs some more thinking.

erszcz: So currently, we can choose this:

```
$ ./bin/gradualizer test/should_pass/variable_binding_leaks.erl
test/should_pass/variable_binding_leaks.erl: The variable on line 9 at column 9 is expected to have type integer() but it has type foo

    case X of
        Y = foo -> length(atom_to_list(Y));
        Y       -> Y + 1  % a different Y
        ^

$ ./bin/gradualizer test/should_pass/type_refinement.erl
$
```

Or this:

```
$ ./bin/gradualizer test/should_pass/variable_binding_leaks.erl
$ ./bin/gradualizer test/should_pass/type_refinement.erl
test/should_pass/type_refinement.erl: The variable on line 13 at column 26 is expected to have type ok but it has type integer() | ok | #{x => y} | pid() | {x, y}

        is_integer(X) -> ok;
        is_pid(X)     -> ok;
        true          -> X
                         ^

$
```

erszcz: Even more interesting that the current master (3cda2e5) passes both of the above tests 🤔

src/typechecker.erl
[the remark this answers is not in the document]
zuiderkwast: Agree that it's outside the scope of this PR.

We shouldn't switch, but choose spec depending on param pattern somehow. It's a function intersection type.

According to Erlang docs, the clauses are not allowed to overlap (though there's no check for that either) but this implies that a spec like this is illegal, because [] is in both domains: `-spec f([a]) -> a; ([b]) -> b.`

src/typechecker.erl
erszcz: Solved with the today's changes :)

src/typechecker.erl
erszcz: This is handled correctly in each case. Consider the following examples.

Variable X is bound and we try to match on it as the first element of itself-the-list:

```
17:09:25 erszcz @ x6 : ~/work/erszcz/gradualizer (fix-list-exhaustiveness-checking-regressions-3 %)
$ cat t.erl
-module(t).

-compile(export_all).

-spec f([atom()]) -> ok.
f(X) ->
    case X of
        [] -> ok;
        [X|_] -> ok;     % <--- Not exhaustive, but Gradualizer thinks it is.
        [hello|_] -> ok  % <--- False positive: clause not reachable?
    end.
17:09:27 erszcz @ x6 : ~/work/erszcz/gradualizer (fix-list-exhaustiveness-checking-regressions-3 %)
$ ./bin/gradualizer t.erl
t.erl: The variable on line 9 at column 10 is expected to have type atom() but it has type [atom()]

    case X of
        [] -> ok;
        [X|_] -> ok;     % <--- Not exhaustive, but Gradualizer thinks it is.
         ^
```

A true positive error is reported - `X is expected to have type atom() but it has type [atom()]`.

We use a free variable Y to match on the first element of the list X:

```
17:09:28 erszcz @ x6 : ~/work/erszcz/gradualizer (fix-list-exhaustiveness-checking-regressions-3 %)
$ cat t.erl
-module(t).

-compile(export_all).

-spec f([atom()]) -> ok.
f(X) ->
    case X of
        [] -> ok;
        [Y|_] -> ok;     % <--- Not exhaustive, but Gradualizer thinks it is.
        [hello|_] -> ok  % <--- False positive: clause not reachable?
    end.
17:09:33 erszcz @ x6 : ~/work/erszcz/gradualizer (fix-list-exhaustiveness-checking-regressions-3 %)
$ ./bin/gradualizer t.erl
t.erl: The clause on line 10 at column 9 cannot be reached
```

A true positive error is reported - since we match on anything with Y, the following clause cannot be reached.

What if we invert the order of Y and `hello`?

```
17:09:35 erszcz @ x6 : ~/work/erszcz/gradualizer (fix-list-exhaustiveness-checking-regressions-3 %)
$ cat t.erl
-module(t).

-compile(export_all).

-spec f([atom()]) -> ok.
f(X) ->
    case X of
        [] -> ok;
        [hello|_] -> ok;  % <--- False positive: clause not reachable?
        [Y|_] -> ok     % <--- Not exhaustive, but Gradualizer thinks it is.
    end.
17:09:52 erszcz @ x6 : ~/work/erszcz/gradualizer (fix-list-exhaustiveness-checking-regressions-3 %)
$ ./bin/gradualizer t.erl
17:09:55 erszcz @ x6 : ~/work/erszcz/gradualizer (fix-list-exhaustiveness-checking-regressions-3 %)
$
```

Nothing is reported as it's valid code - we have a specific clause matching on `hello` and a following catch-all with Y.

All good :D

src/typechecker.erl
erszcz: This is now tracked as #409.

erszcz:
This is green now and I think it's finally ready to merge if you confirm @zuiderkwast :)

src/typechecker.erl
```diff
@@ -2436,18 +2448,19 @@ do_type_check_expr_in(Env, ResTy, {record_index, Anno, Name, FieldWithAnno} = Re
...
     end;
 
 do_type_check_expr_in(Env, ResTy, {'case', _, Expr, Clauses}) ->
     {ExprTy, VarBinds, Cs1} = type_check_expr(Env, Expr),
-    Env2 = Env#env{ venv = add_var_binds(Env#env.venv, VarBinds, Env) },
-    {VB, Cs2} = check_clauses(Env2, [ExprTy], ResTy, Clauses, capture_vars),
-    {union_var_binds(VarBinds, VB, Env), constraints:combine(Cs1,Cs2)};
+    Env1 = add_var_binds(Env, VarBinds, Env),
+    {Env2, Cs2} = check_clauses(Env1, [ExprTy], ResTy, Clauses, capture_vars),
+    {union_var_binds(VarBinds, Env2, Env), constraints:combine(Cs1,Cs2)};
```
zuiderkwast, line 2460: Isn't VarBinds included in Env2 already (since it's added to Env1 and Env2 is build from Env1)? So then, `union_var_binds(VarBinds, Env2, Env)` is a no-op. Am I missing something?

erszcz: Indeed, it seems to be redundant :+1:

erszcz:
Any other comments, @zuiderkwast, or is it ok to merge?

src/typechecker.erl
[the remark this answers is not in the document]
erszcz: For the record, this is related to #97, but not exactly the same thing. I've also created #410 to track this.

zuiderkwast:
I guess it's fine to merge it now. If there are any more minor things that we didn't cover in tests, they can be handled later.

erszcz:
And voila!

This merge, by itself, leads to reduction from

```
$ make gradualize | wc -l
     634
```

to

```
$ make gradualize | wc -l
     485
```

🎉

PR #407, opened 2022-04-03
Refactor throw_orig_type to avoid guards -> improve self-gradualization

erszcz:
Interesting, that even with `make tests` and `make gradualize` a lot of cases of `expected_type` are not covered (btw, they are based on `format_type_error`).

BTW, this is a spin-off from #408, but it ended up being a significant task on its own, so I'm creating a separate ticket to track it.


erszcz:
Done in #427, though in a somewhat different way.

Issue #405, opened 2022-04-02
"The clause ... cannot be reached" when matching on list patterns


erszcz:
Even with #404 when running `make gradualize` we get this report:

```
ebin/constraints.beam: The clause on line 43 at column 1 cannot be reached
```

The problematic code from `constraints.erl` when minimised gives the following example:

```erlang
-spec i([atom()]) -> ok.
i([]) -> ok;
i([Cs]) -> ok;
i([C1, C2 | Cs]) -> ok.
```

The thrown error is:

```erlang
check_clause(_Env, [?type(none)|_], _ResTy, {clause, P, _Args, _Guards, _Block}, _) ->
    throw({type_error, unreachable_clause, P});
```

Which means that if the argument type is `none` when checking a clause, it's considered as unreachable. Now, let's see how the argument type is refined as we check subsequent clauses:

1. We start with `type(list, [atom()])` from the function spec.
2. After the first clause matching on an empty list the type gets refined to `type(nonempty_list, [atom()])`. The pattern of the second clause is inferred as `type(nonempty_list, [atom()])`. This leads to the following refinement after the second clause:

```
{trace,<0.86.0>,call,
    {typechecker,refine_ty,
        [{type,0,tuple,[{type,0,nonempty_list,[{type,0,atom,[]}]}]},
         {type,0,tuple,[{type,0,nonempty_list,[{type,0,atom,[]}]}]},
         #{{type,0,tuple,[{type,0,nonempty_list,[{type,0,atom,[]}]}]} => {}},
         {env,
             #{{i,1} =>
                   [{type,
                        {7,8},
                        bounded_fun,
                        [{type,
                             {7,8},
                             'fun',
                             [{type,
                                  {7,8},
                                  product,
                                  [{type,{7,9},list,[{type,{7,10},atom,[]}]}]},
                              {atom,{7,22},ok}]},
                         []]}]},
             #{},#{},
             #{module => list_exhaustiveness_checking_regressions,
               records => #{},types => #{}},
             false,false,true,30,
             [{type,
                  {7,8},
                  bounded_fun,
                  [{type,
                       {7,8},
                       'fun',
                       [{type,
                            {7,8},
                            product,
                            [{type,{7,9},list,[{type,{7,10},atom,[]}]}]},
                        {atom,{7,22},ok}]},
                   []]}]}]}}
{trace,<0.86.0>,return_from,{typechecker,refine_ty,4},{type,0,none,[]}}
```

[...]

PR #408, opened 2022-04-03
Fix some self-gradualization errors

erszcz:
This builds on top of #404, which cleans up a considerable number of regressions from `make gradualize` output, and improves on it by fixing spec errors and other type issues in Gradualizer itself.

It's also a good exercise to find bugs in Gradualizer or spot hard to type check Erlang idioms.

Before this PR (but already with #404):

```
$ make gradualize | wc -l
     619
```

With this PR:

```
$ make gradualize | wc -l
     352
```

### DIFF
--- a/src/gradualizer_int.erl
+++ b/src/gradualizer_int.erl
@@ -24,6 +24,15 @@
 -type int_range() :: {int(), int()}.
 -type type() :: gradualizer_type:abstract_type().
 
+-type extended_int_type() :: {type, erl_anno:anno(), range, [{integer, erl_anno:anno(), int()}]}
+                           | {'integer', erl_anno:anno(), integer()}.
+%% `extended_int_type' is needed to describe type representations
+%% which are not part of `gradualizer_type:abstract_type()'.
+%% We get rid of it before returning from this module.
+%% TODO: Actually, we don't, though we should :/ https://github.com/josefs/Gradualizer/issues/406
+
+-include("gradualizer.hrl").
+
 %% +----------------------------------------+
 %% |  Functions operating on integer types  |
 %% +----------------------------------------+
@@ -58,13 +67,15 @@ int_type_glb(Ty1, Ty2) ->
 
 %% Range difference, like set difference. The result may be zero, one or two
 %% types.
--spec int_type_diff(type(), type()) -> type().
+%% TODO: https://github.com/josefs/Gradualizer/issues/406
+-spec int_type_diff(type(), type()) -> [type()].
 int_type_diff(Ty1, Ty2) ->
     IntRanges = int_range_diff(int_type_to_range(Ty1),
                                int_type_to_range(Ty2)),
     %% Make sure the result is a standard erlang type.
     %% Perhaps we can include generalized ranges such as 10..pos_inf
     ExpandedRanges = lists:map(fun int_range_expand_to_valid/1, IntRanges),
+    %% TODO: This fails self-gradualisation, see https://github.com/josefs/Gradualizer/issues/406
     int_ranges_to_type(ExpandedRanges).
 
 %% Merges integer types by sorting on the lower bound and then merging adjacent
@@ -103,7 +114,7 @@ int_type_to_range({char, _, I})                        -> {I, I};
 int_type_to_range({integer, _, I})                     -> {I, I}.
 
 %% Converts a range back to a type.
--spec int_range_to_type(int_range()) -> type().
+-spec int_range_to_type(int_range()) -> type() | extended_int_type().
 int_range_to_type(Range) ->
     union(int_range_to_types(Range)).
 
@@ -113,7 +124,7 @@ int_range_to_type(Range) ->
 
 %% Converts a range to a list of types. Creates two types in some cases and zero
 %% types if lower bound is greater than upper bound.
--spec int_range_to_types(int_range()) -> [type()].
+-spec int_range_to_types(int_range()) -> [type() | extended_int_type()].
 int_range_to_types({neg_inf, pos_inf}) ->
     [type(integer)];
 int_range_to_types({neg_inf, -1}) ->
@@ -121,11 +132,15 @@ int_range_to_types({neg_inf, -1}) ->
 int_range_to_types({neg_inf, 0}) ->
     [type(neg_integer), {integer, erl_anno:new(0), 0}];
 int_range_to_types({neg_inf, I}) when I > 0 ->
+    %% I > 0 - see the guard above - so we can safely assert that
+    I = ?assert_type(I, pos_integer()),
     [type(neg_integer),
      {type, erl_anno:new(0), range, [{integer, erl_anno:new(0), 0}
                                     ,{integer, erl_anno:new(0), I}]}];
 int_range_to_types({neg_inf, I}) when I < -1 ->
-    %% Non-standard
+    %% Non-standard - see extended_int_type()
+    %% I < -1 - see the guard - so we can safely assert that
+    I = ?assert_type(I, neg_integer()),
     [{type, erl_anno:new(0), range, [{integer, erl_anno:new(0), neg_inf}
                                     ,{integer, erl_anno:new(0), I}]}];
 int_range_to_types({I, pos_inf}) when I < -1 ->
@@ -155,7 +170,7 @@ int_range_to_types({I, J}) when I > J ->
     [].
 
 %% Merges ranges and returns a single type (possibly a union).
--spec int_ranges_to_type([int_range()]) -> type().
+-spec int_ranges_to_type([int_range()]) -> int_range().
 int_ranges_to_type(Ranges) ->
     union(int_ranges_to_types(Ranges)).
 
@@ -165,7 +180,7 @@ int_ranges_to_types(Ranges) ->
     MergedRanges = merge_int_ranges(Ranges),
     lists:flatmap(fun int_range_to_types/1, MergedRanges).
 
--spec union([type()]) -> type().
+-spec union([type() | extended_int_type()]) -> type() | extended_int_type() | int_range().
 union([]) -> type(none);
 union([T]) -> T;
 union(Ts) -> type(union, Ts).
@@ -233,6 +248,7 @@ int_range_diff({Lo1, Hi1}, {Lo2, Hi2}) ->
 
 %% Makes sure a range can be represented as a syntactically valid Erlang type,
 %% by expanding it if necessary.
+-spec int_range_expand_to_valid(int_range()) -> int_range().
 int_range_expand_to_valid({neg_inf, N}) when is_integer(N),
                                              N < -1 ->
     {neg_inf, -1}; % neg_integer()
--- a/src/gradualizer_lib.erl
+++ b/src/gradualizer_lib.erl
@@ -8,12 +8,14 @@
          remove_pos_typed_record_field/1]).
 -export_type([graph/1, tenv/0]).
 
+-type type() :: gradualizer_type:abstract_type().
+
 %% Type environment, passed around while comparing compatible subtypes.
 -type tenv() :: #{ module => module(),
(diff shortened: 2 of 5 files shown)


src/gradualizer_int.erl
```diff
@@ -24,6 +24,15 @@
 -type int_range() :: {int(), int()}.
 -type type() :: gradualizer_type:abstract_type().
 
+-type extended_int_type() :: {type, erl_anno:anno(), range, [{integer, erl_anno:anno(), int()}]}
+                           | {'integer', erl_anno:anno(), integer()}.
+%% `extended_int_type' is needed to describe type representations
+%% which are not part of `gradualizer_type:abstract_type()'.
```
zuiderkwast, line 30: Why is this not a subtype to type()?

src/gradualizer_type.erl
```diff
@@ -364,4 +365,8 @@
 
 -type type_name() :: atom().
 
+-type af_constraint() :: {'type', anno(), 'constraint',
```
zuiderkwast, line 368: For everything that is not copied from OTP here,  please add a comment about it.

Don't export the af types but add an abstract_constraint typedef as for the other exported ones.

src/gradualizer_lib.erl
```diff
@@ -100,11 +102,13 @@ reverse_graph(G) ->
...
 -spec get_type_definition(UserTy, Env, Opts) -> {ok, Ty} | opaque | not_found when
-      UserTy :: gradualizer_type:abstract_type(),
+      UserTy :: type(),
       Env :: typechecker:env(),
       Opts :: [annotate_user_types],
-      Ty :: gradualizer_type:abstract_type().
+      Ty :: type().
 get_type_definition({remote_type, _Anno, [{atom, _, Module}, {atom, _, Name}, Args]}, _Env, _Opts) ->
+    %% We matched out the non-type() elements above so we can assert Args :: [type()]
+    Args = ?assert_type(Args, [type()]),
```
zuiderkwast, line 111: Why? Atom literals are part of abstract_type().

src/typechecker.erl
```diff
@@ -1304,7 +1311,7 @@ expect_fun_type_union(Env, [Ty|Tys]) ->
     Any :: any,
     FieldsTy :: {fields_ty, [typed_record_field()], constraints:constraints()},
     FieldsTys :: {fields_tys, [[typed_record_field()]], constraints:constraints()},
-    TypeError :: {type_error, type()}.
+    TypeError :: {type_error, atom() | type()}.
```
zuiderkwast, line 1314: Atoms are types IUCC

src/typechecker.erl
```diff
@@ -3659,13 +3668,25 @@ refine_ty({atom, _, _}, ?type(atom), _, _) ->
     type(none);
 refine_ty(?type(list, E), ?type(nil), _, _Env) ->
     type(nonempty_list, E);
-refine_ty(?type(list, E), ?type(nonempty_list, E), _, _Env) ->
-    type(nil);
+refine_ty(?type(list, [ElemTy1]), ?type(nonempty_list, [ElemTy2]), Trace, Env) ->
+    case refine(ElemTy1, ElemTy2, Trace, Env) of
```
zuiderkwast, line 3672: 🍰🎂🥧🍺

src/typechecker.erl
```diff
@@ -4227,14 +4250,15 @@ add_type_pat(CONS = {cons, P, PH, PT}, ListTy, Env, VEnv) ->
             TailTy = normalize(type(union, [ListTy, type(nil)]), Env),
             {_TailPatTy, _TauUBound, VEnv3, Cs} = add_type_pat(PT, TailTy, Env, VEnv2),
             NonEmptyTy = rewrite_list_to_nonempty_list(ListTy),
-            {NonEmptyTy, NonEmptyTy, VEnv3, Cs};
+            {type(none), NonEmptyTy, VEnv3, Cs};
```
zuiderkwast, line 4253: `[_|_]` exhausts nonempty list. Have you saved that for another PR?

zuiderkwast (commented):
Very nice!

src/gradualizer_type.erl
erszcz: This is copied from OTP, it's a part of `erl_parse:abstract_form()`. To my surprise, the `abstract_form` is the only `abstract_*` thing not copied from `erl_parse` to `gradualizer_type` verbatim 🤷

> Don't export the af types but add an abstract_constraint typedef as for the other exported ones.

I could do this, but then we're deviating from the naming used in `erl_parse`. Do you think that's ok?

src/gradualizer_lib.erl
erszcz: Atom literals can be represented by `type()`, but literal atoms are not of type `type()`, see https://github.com/josefs/Gradualizer/blob/3cda2e5aa1b1313b3e02fce67986fcda3ab9a02c/src/gradualizer_type.erl#L290-L292

src/typechecker.erl
erszcz: That's not the case, as above. Atom literals can be represented by `type()`, but literal atoms are not of type `type()`, see https://github.com/josefs/Gradualizer/blob/3cda2e5aa1b1313b3e02fce67986fcda3ab9a02c/src/gradualizer_type.erl#L290-L292

src/gradualizer_type.erl
zuiderkwast: > (...) surprise (...) copied (...) verbatim

We needed to copy the whole module to be able to export types not exported by default.

> Do you think that's ok?

I changed my mind. You can export the `af_` name. I see now that `erl_parse` has a few exported `af_` types: https://www.erlang.org/doc/man/erl_parse.html#data-types

src/typechecker.erl
erszcz: Hmmm, I have to think it through 🤔

src/gradualizer_lib.erl
erszcz: Ahh, sorry, the above is bollocks, as the atoms here are actually in the AST representation. Let me check again...

src/typechecker.erl
zuiderkwast: I think it's only the pattern `[A | B]` where A and B are free variables (not in VEnv) that works.

src/typechecker.erl
```diff
@@ -4227,14 +4250,15 @@ add_type_pat(CONS = {cons, P, PH, PT}, ListTy, Env, VEnv) ->
...
         {elem_ty, ElemTy, Cs1} ->
-            {_PatTy1, _UBound1, VEnv2, Cs2} =
+            {PatTy1, _UBound1, VEnv2, Cs2} =
                 add_type_pat(PH, normalize(ElemTy, Env), Env, VEnv),
             TailTy = normalize(type(union, [ListTy, type(nil)]), Env),
             {_PatTy2, _Ubound2, VEnv3, Cs3} = add_type_pat(PT, TailTy, Env, VEnv2),
+            PatTy = type(nonempty_list, [PatTy1]),
             NonEmptyTy = rewrite_list_to_nonempty_list(ListTy),
-            {NonEmptyTy, NonEmptyTy, VEnv3, constraints:combine([Cs1, Cs2, Cs3])};
+            {PatTy, NonEmptyTy, VEnv3, constraints:combine([Cs1, Cs2, Cs3])};
```
zuiderkwast, line 4261: Here, I think we should return `{none(), ...}` too, in the normal case.

Only if the pattern is match-all (`[A|B]` where A and B are free, or A is bound and ElemTy is a singleton type) the pattern exhausts the type. In that case, we can return PatTy I suppose.

src/gradualizer_lib.erl
erszcz: Ok, so we start with this error:

```
$ gradualizer -I include/ -- src/gradualizer_lib.erl
src/gradualizer_lib.erl: The variable on line 112 at column 43 is expected to have type [type()]
but it has type [abstract_type()] | {atom, anno(), atom()}

    %% We matched out the non-type() elements above so we can assert Args :: [type()]
    %Args = ?assert_type(Args, [type()]),
    gradualizer_db:get_type(Module, Name, Args);
                                          ^^^^
```

We check the type of a remote call node:
https://github.com/josefs/Gradualizer/blob/3cda2e5aa1b1313b3e02fce67986fcda3ab9a02c/src/gradualizer_type.erl#L254-L257

So `Args` is a `[type()]`, not a `type()`, which is fine for passing down. Singleton atoms are not fine, as we need a list. We know we've already matched on the singleton atoms, so by elimination we're safe to assert `Args :: [type()]`.

zuiderkwast: Great! Only the comment needs an update then. E.g. `%% We matched out the single atom arguments, so only Args (list-of-type) remains`.

src/typechecker.erl
zuiderkwast: See the comment above `add_types_pats` about the return tuple `{PatTys, UBounds, NewVEnv, Constraints}`:

```Erlang
%% The returned lists of types are interpreted like this:
%% PatTy :: Pat as if Pat were a type. For match-all patterns, PatTy
%% is the same as the type. For patterns matching a singleton type, PatTy
%% is the singleton type. Otherwise, PatTy is none(). PatTy is a type exhausted
```

src/typechecker.erl
erszcz: Exactly :+1:

erszcz: This return value is not used so far, so maybe let's decide about what to return once we have a use for it.
https://github.com/josefs/Gradualizer/blob/3cda2e5aa1b1313b3e02fce67986fcda3ab9a02c/src/typechecker.erl#L4283

It's fine to return even just `type_error`, I just didn't want to change too much in one go.

erszcz:
@zuiderkwast I've given your points about exhaustive list types some thought. I also considered if just looking at whether the patterns match fixed length lists or any lists is sufficient - I'm afraid it's not :/ It seems that to model cases like these:

```erlang
-spec i([atom()]) -> ok.
i([]) -> ok;
i([Cs]) -> ok;
i([C1, C2 | Cs]) -> ok.

-spec j([atom()]) -> ok.
j([]) -> ok;
j([C1, C2 | _]) -> ok;
j([Cs]) -> ok.
```

it's necessary to track the lengths of the patterns. Please see https://github.com/josefs/Gradualizer/issues/405#issuecomment-1088904385 for a bit more detail.

zuiderkwast:
> it's necessary to track the lengths of the patterns

For clauses that exhaust a list using patterns `[]`, `[X]` and `[X, Y | Z]`, we can't perform exhaustiveness-checking without encoding the list length in the type, but that doesn't mean we can't silence the false positives and make it pass.

I believe we can run into similar problems for other types and there will always be something we can't do, unless we continue to extend the type language over and over. I'm not saying we can't do that, but I'm almost certain that we'll run into a lot of times when we need to decide how subtyping, GLB, refinement, etc. work for these types. We need draw the line somewhere. (We already have some overly fancy logic for arithmetic operators, for example, that I think we could live without, to make the type system simpler, but I know others think it's worth it.)

zuiderkwast:
Assume we add `list(T, N)` and `list(T, N+)` types.

Should we be able do exhaustiveness-checking for the following?

```Erlang
q([X | Xs]) when length(Xs) rem 2 == 0 ->
    X;
q(Xs) when length(Xs) rem 2 == 0 ->
    bananas.
```

(We could solve it if we encode odd and even lengths of lists in the type. Not saying we should.)

erszcz:
The current status is:

```
10:02:04 erszcz @ x6 : ~/work/erszcz/gradualizer ((a72e7f0...) %)
$ make gradualize | wc -l
     485
```

on master versus

```
10:02:24 erszcz @ x6 : ~/work/erszcz/gradualizer (fix-self-gradualization-errors %)
$ make gradualize | wc -l
     361
```

on this branch. Less of a difference, but still worth it.

PR #411, opened 2022-05-09
typechecker: check exhaustiveness against refined args

### LINKED ISSUE #414: exhaustiveness issue alert for wrong parameter
I found an issue related to the message printed when an exhaustive check finds non-exhaustive patterns.

The following test case incorrectly reports that the second parameter is not exhaustive while the issue lies in the first parameter:

``` erlang
-module(error3).

-export([my_function/2]).

-record(my_record, { something :: non_neg_integer() }).

-spec my_function(#my_record{} | undefined, non_neg_integer()) -> #my_record{}.
my_function(Record = #my_record{}, Number) ->
    Record#my_record{something = Number}.
```

As we can see, the `undefined` value is not handled by any function body but this is what Gradualizer reports:

```
/gradualizer/error3.erl: Nonexhaustive patterns on line 8 at column 1
	0
```

If the fix in #411 is applied, this is the issue reported:

```
/gradualizer/error3.erl: Nonexhaustive patterns on line 8 at column 1
	undefined
```

cc: @zuiderkwast it seems you were right in https://github.com/josefs/Gradualizer/issues/339#issuecomment-895786061.

berbiche:
This PR changes the logic in the exhaustiveness check to check for refinability of the refined args instead of the non-refined args.

This was suggested in a different issue here: https://github.com/josefs/Gradualizer/issues/339#issuecomment-895786061
The change resolves an issue I encounter where the type checked for exhaustiveness is accepted but the refined type fails with a `badargs` in `pick_value`.

As far as I understand, the `refinable3/` function also acts as a gatekeeper for `pick_value/2`.

edit: Fixes #414

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -3490,7 +3490,7 @@ disable_exhaustiveness_check(#env{} = Env) ->
 %% @end
 check_arg_exhaustiveness(Env, ArgsTy, Clauses, RefinedArgsTy) ->
     case exhaustiveness_checking(Env) andalso
-         all_refinable(ArgsTy, Env) andalso
+         all_refinable(RefinedArgsTy, Env) andalso
          no_clause_has_guards(Clauses) andalso
          some_type_not_none(RefinedArgsTy)
     of


berbiche:
One issue with checking refinement against the refined args is that `pick_value` is called against the refined type instead of the non refined type.

For instance, the refined type of a char is an integer range. The information is more exact but loses meaning for the user (IMO).

zuiderkwast:
> the refined type of a char is an integer range

Can we turn it into a char range? Or is it normalized here?

IIRC, normalize turns chars into integers, but in some cases (like this) it'd be good to keep them as chars. OTOH in other cases it's good to have a unique representation to pattern match on.

> The information is more exact but loses meaning for the user (IMO).

I agree, but maybe we can live with a more obscure message if we get better exactness?

PR #419, opened 2022-06-01
Fix false nonexhaustive record patterns warning

erszcz:
This stems from the research in #418 and supersedes any experiments done there. By dropping argument-wise exhaustiveness checking we get rid of the false positive warning and therefore go from:

```
$ make gradualize | wc -l
     361
```

to

```
$ make gradualize | wc -l
     307
```


berbiche:
Neat, this seems better than https://github.com/josefs/Gradualizer/pull/411

I will try the PR in the upcoming days 👍

erszcz:
> Do you think it is less important than the added test case?

Yes, I think it's less important. We get rid of a very annoying and very common false positive warning that way.

> Can you think of a way to solve both?

I don't know how to do that and I'm unsure if it's possible at all without really significant changes to the typechecker. The long explanation is in #418, together with some approaches I tried, but to no success :/

zuiderkwast:
Then I think we can document that we do exhaustiveness if and only if all of the parameters are typed, not any(). Does it make sense? It's good to have a clear rule to follow.

zuiderkwast (approved):
OK, feel free to merge when ready.

The edoc comments are mainly for internal use. The user documentation (as far as we have any) is on the wiki so we can consider updating that too.

PR #420, opened 2022-06-07
Make an empty map() refinable

erszcz:


zuiderkwast:
It looks like it would have been enough to remove `length(Assocs) =/= 0` in the next function clause 18 lines below.

Issue #422, opened 2022-06-08
Make atom() accepted for exhaustiveness checking


erszcz:
The following type cannot be checked for exhaustiveness:

```erlang
-type complex_t() :: {string, string()}
                   | {list, list(allergen())}
                   | {atom, atom()}
                   | {binary, binary()}
                   | {record, record()}
                   | {map, #{non_empty := true}}
                   | {integer, integer()}
                   | {float, float()}.
```

The check kicks in when we comment out the `atom` variant:

```erlang
-type complex_t() :: {string, string()}
                   | {list, list(allergen())}
                   %| {atom, atom()}
                   | {binary, binary()}
                   | {record, record()}
                   | {map, #{non_empty := true}}
                   | {integer, integer()}
                   | {float, float()}.
```

This is interesting, since a union of atoms can be checked for exhaustiveness.

erszcz:
For some context, `atom()` in general is not `refinable` - I'm guessing this is the sole reason:

```
9> Type = fun (T) -> typelib:remove_pos(typelib:parse_type(T)) end.
#Fun<erl_eval.44.40011524>
10> typechecker:refinable(Type("atom()"), []).
false
```

erszcz:
For more context, singleton atoms or a union of explicit atoms are accepted in complex types:

```erlang
-type complex_t() :: {string, string()}
                   | {list, list(allergen())}
                   | {atom, example_atom | another_atom}
                   | {binary, binary()}
                   | {record, record()}
                   | {map, #{non_empty := true}}
                   | {integer, integer()}
                   | {float, float()}.
```

PR #425, opened 2022-06-08
Make atom() refinable

erszcz:
This explores, apparently non-trivial, consequences of making `atom()` refinable. This addresses #422.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4941,6 +4941,7 @@ type_check_forms(Forms, Opts) ->
     AllErrors = lists:foldr(fun (Function, Errors) ->
                                     type_check_form_with_timeout(Function, Errors, StopOnFirstError, Env, Opts)
                             end, [], ParseData#parsedata.functions),
+    timer:sleep(200),
     lists:reverse(AllErrors).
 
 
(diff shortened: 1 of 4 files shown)


src/typechecker.erl
```diff
@@ -4941,6 +4941,7 @@ type_check_forms(Forms, Opts) ->
     AllErrors = lists:foldr(fun (Function, Errors) ->
                                     type_check_form_with_timeout(Function, Errors, StopOnFirstError, Env, Opts)
                             end, [], ParseData#parsedata.functions),
+    timer:sleep(200),
```
zuiderkwast, line 4944: :scream:

erszcz: This is a draft PR and this lines comes from a commit title "debug" ;)

erszcz: For some context, if we trace execution (as set up in `gradualizer_app`), we need this sleep, as otherwise the VM exits before the tracer process manages to print all the messages we want to see.

erszcz:
After giving it some thought I think that making `refinable(atom()) == true` currently brings more harm than good, so I'm closing this one. We might revisit it at some point, but I don't think it's essential now. Singleton atoms or a union of explicit atoms is accepted in complex types anyway:

```erlang
-type complex_t() :: {string, string()}
                   | {list, list(allergen())}
                   | {atom, example_atom | another_atom}
                   | {binary, binary()}
                   | {record, record()}
                   | {map, #{non_empty := true}}
                   | {integer, integer()}
                   | {float, float()}.
```

PR #427, opened 2022-06-11
Another batch of type error fixes

erszcz:
Before:

```
$ make gradualize  | wc -l
     307
```

After:

```
$ make gradualize  | wc -l
     253
```

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4756,7 +4756,8 @@ type_of_bin_element({bin_element, _P, Expr, _Size, Specifiers}, OccursAs) ->
 
 %%% Helper functions
 
--spec type(atom(), [any()]) -> type().
+-spec type(map | tuple, any) -> type();
+          (atom(), [any()]) -> type().
 type(Name, Args) ->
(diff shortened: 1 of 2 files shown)


erszcz:
@zuiderkwast WDYT?

src/typechecker.erl
```diff
@@ -5132,3 +5130,45 @@ number_of_exported_functions(Forms) ->
 
 line_no(Expr) ->
     erl_anno:line(element(2, Expr)).
+
+type_error(Kind) ->
+    {type_error, Kind}.
+
```
zuiderkwast, line 5146: What do these solve?

zuiderkwast: The're just a bunch of functions without a spec. Isn't it better to define a type for all the errors?

zuiderkwast: That's the next step perhaps? Anyway, this looks good so feel free to merge. :-)

erszcz: An `error` type won't hurt, indeed. I'll make it a bit more explicit in the next free time slot.

PR #439, opened 2022-08-24
Fix gradualizer_int:int_type_to_range/1 crash

### LINKED ISSUE #438: function_clause crash in `gradualizer_int:int_type_to_range`
I get a `function_clause` crash like this when running Gradualizer on our code:
```
Error: function_clause
[{gradualizer_int,int_type_to_range,
                  [{var,0,'T'}],
                  [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/gradualizer_int.erl"},
                   {line,106}]},
 {gradualizer_int,negate_int_type,1,
                  [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/gradualizer_int.erl"},
                   {line,95}]},
 {typechecker,do_type_check_expr,2,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,1869}]},
 {typechecker,type_check_expr,2,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,1550}]},
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,2330}]},
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,2321}]},
 {typechecker,check_clause,5,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,3581}]},
 {typechecker,'-check_clauses/5-fun-0-',4,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,3489}]}
```
If I run with `verbose` enabled, it tells me this immediately before the crash
```
Checking function commission/1
Pushing {clauses_controls,true}
94: Checking clause :: any()
94: Checking that -lists:min([C, 0]) :: any()
94: Checking that [C, 0] :: [T, ...]
94: Checking that C :: T
94: Checking that [0] :: [T]
94: Checking that 0 :: T
94: Checking that [] :: [T] | []
94: Propagated type of lists:min([C, 0]) :: T
Task reported error with trace from function commission/1
```
OTP: 23.3.4.2
Gradualizer: latest master (f00f1d2)

erszcz:
Fixes #438 by assuming that a type var is any number. It should probably also generate a constraint on the type var, but there's no path for returning constraints from `gradualizer_int` yet, so we might need to revisit this module anyway once there's a constraint solver in place.

### DIFF
--- a/src/gradualizer_int.erl
+++ b/src/gradualizer_int.erl
@@ -103,9 +103,6 @@ negate_int_type(RangeTy) ->
 
 %% Integer type to range.
 -spec int_type_to_range(type()) -> int_range().
-%% TODO: `{var, _, _}' should probably generate a constraint on the type var, too,
-%% but fixing a crash is more important at this point.
-int_type_to_range({var,  _, _})                        -> {neg_inf, pos_inf};
 int_type_to_range({type, _, integer, []})              -> {neg_inf, pos_inf};
 int_type_to_range({type, _, neg_integer, []})          -> {neg_inf, -1};
 int_type_to_range({type, _, non_neg_integer, []})      -> {0, pos_inf};
--- a/src/gradualizer_int.erl
+++ b/src/gradualizer_int.erl
@@ -103,6 +103,9 @@ negate_int_type(RangeTy) ->
 
 %% Integer type to range.
 -spec int_type_to_range(type()) -> int_range().
+%% TODO: `{var, _, _}' should probably generate a constraint on the type var, too,
+%% but fixing a crash is more important at this point.
+int_type_to_range({var,  _, _})                        -> {neg_inf, pos_inf};
 int_type_to_range({type, _, integer, []})              -> {neg_inf, pos_inf};
 int_type_to_range({type, _, neg_integer, []})          -> {neg_inf, -1};
 int_type_to_range({type, _, non_neg_integer, []})      -> {0, pos_inf};
--- a/test/should_pass/rigid_type_variables_pass.erl
+++ b/test/should_pass/rigid_type_variables_pass.erl
@@ -5,6 +5,5 @@
 -spec id(A) -> A.
 id(X) -> X.
 
-%% Caused a crash in int_type_to_range/1 on {var, _, 'A'}.
-%% It's now fixed by assuming that a type var means any integer whatsoever.
+%% Used to cause a crash in int_type_to_range/1 on {var, _, 'A'}.
 foo() -> -id(42).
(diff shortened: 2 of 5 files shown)


src/gradualizer_int.erl
```diff
@@ -103,6 +103,9 @@ negate_int_type(RangeTy) ->
 
 %% Integer type to range.
 -spec int_type_to_range(type()) -> int_range().
+%% TODO: `{var, _, _}' should probably generate a constraint on the type var, too,
+%% but fixing a crash is more important at this point.
+int_type_to_range({var,  _, _})                        -> {neg_inf, pos_inf};
```
zuiderkwast, line 108: I don't think we should ever get here for `var`, just like we never get here for `any()`. We should make sure we only have an integer type when we get to this module.

Instead, I think the problem should be fixed in `typechecker:negate_num_type/2`, by adding var to the guards in the first clause:

```Erlang
negate_num_type({type, _, TyName, []} = Ty, _Env) when
      TyName =:= any;
      TyName =:= integer;
      TyName =:= float ->
    Ty;
```

zuiderkwast: This function is absent in the stacktrace due to the stack frame being overwritten for tail calls, but it seems to be the only path where this is called and it makes sense for the stack trace posted in the issue IMO.

erszcz: Indeed, adding a clause here works :+1:

test/should_pass/rigid_type_variables_pass.erl
```diff
@@ -0,0 +1,10 @@
+-module(rigid_type_variables_pass).
+
+-compile([export_all, nowarn_export_all]).
+
+-spec id(A) -> A.
+id(X) -> X.
+
+%% Caused a crash in int_type_to_range/1 on {var, _, 'A'}.
+%% It's now fixed by assuming that a type var means any integer whatsoever.
+foo() -> -id(42).
```
zuiderkwast, line 9: Well, the type variable can also be `any()` or `float()`. What we assume is that it's a type that if an integer is included, its negation is also included, or sth like that.... We should perhaps add some constraints on the negated variable, but I don't know how that would look like or how it could work.

PR #440, opened 2022-08-24
Add specs for ++/2 and --/2

### LINKED ISSUE #437: `Call to undefined function :erlang.'++'/2` in OTP 22
Hi, just want to say how awesome this project is. I'm hoping to use it on a 300k+ LOC Elixir project for work -- just want to work out some issues first. Note that because I'm using Elixir with Gradient, the error message in the title is probably formatted differently -- but I've narrowed this down to 100% being an issue with Gradualizer, not Gradient.

The issue is that using the `++` operator on OTP 22 generates the error message `Call to undefined function :erlang.'++'/2`. This issue does not happen with OTP 24. I was able to narrow this down to `gradualizer_db:import_module/2` loading specs from the erlang BEAM file, and on OTP 22 that file doesn't contain a spec for `++`.

I verified this by disassembling the BEAM files using:
```
iex> {:ok, forms} = Gradient.ElixirFileUtils.get_forms_from_beam('/Users/myusername/.asdf/installs/erlang/22.3.4.22/lib/erts-10.7.2.14/ebin/erlang.beam')
iex> IO.puts(:erl_prettypr.format(:erl_syntax.form_list(forms)))
```
Again, sorry for using Elixir here rather than Erlang, but it's probably pretty simple to do the same thing from Erlang using `:erl_prettypr.format(:erl_syntax.form_list(forms))`. Anyways, I logged out what file it's loading the `erlang` module from, and when it loads from `.asdf/installs/erlang/22.3.4.22/lib/erts-10.7.2.14/ebin/erlang.beam`, the `++` spec is not present, but when it loads from `.asdf/installs/erlang/24.2.1/lib/erts-12.2.1/ebin/erlang.beam`, the `++` spec IS present.

(I guess it's possible that in the very latest version of OTP 22, since 22.3.4.22, this spec has been added. But that seems unlikely. I can double check on this.)

What seems like the easiest fix is to add a thing somewhere inside `gradualizer_db` when we import the `erlang` module, which is only enabled for OTP version 22 and earlier. (Maybe also OTP 23 -- I'll have to check.) At that point we'd import some hard-coded function specs for missing functions, such as `++`, and any others we find. (I'm assuming there are quite a few more.)

But, I'm open to other ways to fix this. I'm happy to submit an MR for this sometime in the next few days/weeks, just wanted to agree on the method of solving the issue before I start coding it. What do you think?

erszcz:
Hopefully fixes #437.

### DIFF
--- a/priv/prelude/erlang.specs.erl
+++ b/priv/prelude/erlang.specs.erl
@@ -43,7 +43,7 @@
 %%        in operator  ++/2
 %%           called as [a|b] ++ [c]
 %%
--spec erlang:'++'(list(T1), list(T2)) -> list(T1 | T2);
-                 (list(T1), nonempty_improper_list(T2, T3)) -> nonempty_improper_list(T1 | T2, T3);
-                 ([], T) -> T;
-                 (nonempty_list(T1), T2) -> nonempty_improper_list(T1, T2).
+-spec erlang:'++'([T1, ...], [T2])                           -> [T1 | T2, ...];
+                 ([T1, ...], nonempty_improper_list(T2, T3)) -> nonempty_improper_list(T1 | T2, T3);
+                 ([T1, ...], T2)                             -> nonempty_improper_list(T1, T2);
+                 ([], T)                                     -> T.
--- a/priv/prelude/erlang.specs.erl
+++ b/priv/prelude/erlang.specs.erl
@@ -15,3 +15,35 @@
 
 %% -spec tl([term(), ...]) -> [term()].
 -spec tl([A, ...]) -> [A].
+
+-spec erlang:'--'(list(), list()) -> list().
+
+%% The original spec is:
+%%
+%% -spec erlang:'++'(list(), term()) -> term().
+%%
+%% Now, this is funny:
+%%
+%%   > [] ++ b.
+%%   b
+%%   > [a] ++ b.
+%%   [a|b]
+%%   > [a, b] ++ c.
+%%   [a,b|c]
+%%   > [a|b] ++ c.
+%%   ** exception error: bad argument
+%%     in operator  ++/2
+%%        called as [a|b] ++ c
+%%   > [] ++ [a].
+%%   [a]
+%%   > [a,b] ++ [c].
+%%   [a,b,c]
+%%   > [a|b] ++ [c].
+%%   ** exception error: bad argument
+%%        in operator  ++/2
+%%           called as [a|b] ++ [c]
+%%
+-spec erlang:'++'(list(T1), list(T2)) -> list(T1 | T2);
+                 (list(T1), nonempty_improper_list(T2, T3)) -> nonempty_improper_list(T1 | T2, T3);
+                 ([], T) -> T;
+                 (nonempty_list(T1), T2) -> nonempty_improper_list(T1, T2).
--- a/test/known_problems/should_pass/list_op_should_pass.erl
+++ b/test/known_problems/should_pass/list_op_should_pass.erl
@@ -1,36 +1,62 @@
 -module(list_op_should_pass).
 
--export([nil_concat_elem_gives_elem/0,
-         nonempty_concat_elem_gives_improper1/0, nonempty_concat_elem_gives_improper2/0,
-         nil_concat_nonempty_gives_proper/0,
-         nonempty_concat_nonempty_gives_proper/0]).
+-export([nil_concat_op_elem_gives_elem/0,
+         nil_concat_fun_elem_gives_elem/0,
+         nonempty1_concat_op_elem_gives_improper/0,
+         nonempty1_concat_fun_elem_gives_improper/0,
+         nonempty2_concat_op_elem_gives_improper/0,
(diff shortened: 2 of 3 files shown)


priv/prelude/erlang.specs.erl
```diff
@@ -15,3 +15,33 @@
...
+%%   [a,b,c]
+%%   > [a|b] ++ [c].
+%%   ** exception error: bad argument
+%%        in operator  ++/2
+%%           called as [a|b] ++ [c]
+%%
+-spec erlang:'++'(list(T1), list(T2)) -> list(T1 | T2);
+                 (list(T1), nonempty_improper_list(T2, T3)) -> nonempty_improper_list(T1 | T2, T3);
+                 ([], T) -> T;
+                 (nonempty_list(T1), T2) -> nonempty_improper_list(T1, T2).
```
zuiderkwast, line 47: According to docs, the clauses in a spec are not allowed to overlap. It's not like a case expression where the first matching clause is chosen.

I don't know if we can write such spec though. Perhaps we should go for this one.

Would you care to add some examples (test cases) to show that this fancy spec is useful?

japhib:
Confirmed locally that this fixes https://github.com/josefs/Gradualizer/issues/437

priv/prelude/erlang.specs.erl
erszcz: > Would you care to add some examples (test cases) to show that this fancy spec is useful?

Hmm, I tried coming up with some, but apparently this spec is not taken into account at all. [There are hardcoded rules for handling `++` and `--`](https://github.com/josefs/Gradualizer/blob/master/src/typechecker.erl#L2940-L2947). So for the sake of #437 the only important thing is to have any spec defined. I'll work on some known problems tests for this.

zuiderkwast: Yeah I thought there are special code for binops. Perhaps the spec ia needed only when called as a funcrion ie like `erlang:'++'(X, Y)`...?

erszcz: Ok, this seems to be a bigger topic previously being discussed in #163 (closed today by Pierre). I think in this PR we should focus on unblocking #437 and maybe providing known_problems for some list operators, but in general not let it linger for too long.

erszcz: Ok, I have added the tests following the examples and the sketched out spec. Indeed, the function forms seem to already be checked against the spec, but not the operator forms.

zuiderkwast: Well, officially Erlang's type/spec language is specific to Dialyzer. We just happen to reuse that.

According to some paper I've read (the one about gradual typing with union, intersection and negation types IIRC), it is problematic to have overlapping domains. (I don't remember exactly why, but there is a way to rewrite the code to exclude the domains of the previous clauses using intersection and negation types.)

Regarding custom code vs specs for operators, it was decided that it's useful to have some richer semantics for some operators than what the type language can express. I don't have a strong opinion but think we can allow custom code for operators and for functions too, for some bifs. The distinction between binary operators and function calls is just syntactic.

erszcz: > it's useful to have some richer semantics for some operators

So far it seems we're seeing incomplete implementations instead of richer semantics ;p As in the case of `++` whose type checking implementation is less accurate than the spec I sketched above. Anyway, I think that we can have custom implementations for "some operators" where it makes sense yet default to treating the operators and `erlang:*` invocations as equivalent.

> According to some paper [...]

Yeah, I remember reading that, too. That's a convincing point, indeed. If they overlap it's just not clear (not possible to tell deterministically) which spec to choose.

Ultimately, it's possible to rewrite the spec not to have overlapping domains. It's going to have more clauses, so will be slightly less readable, but it should still match the new tests.

erszcz: > Ultimately, it's possible to rewrite the spec not to have overlapping domains.

Ok, I take it back. It's not possible without negation types. `T2` in the third clause overlaps the 2nd param of the first and second clauses...

```
-spec erlang:'++'([T1, ...], [T2])                           -> [T1 | T2, ...];
                 ([T1, ...], nonempty_improper_list(T2, T3)) -> nonempty_improper_list(T1 | T2, T3);
                 ([T1, ...], T2)                             -> nonempty_improper_list(T1, T2);
                 ([], T)                                     -> T.
```

erszcz: The thing I wanted to achieve with the spec was the type of a proper `[a,b,c]` ++ `[d,e,f]` to be a proper list. Due to the implementation of the `++` operator and improper lists it turns out we cannot escape the `_ ++ _ :: term()` :/

zuiderkwast: I know. I believe Josef would just say that ++ should be used for proper lists only. Anything else is a type error. (But I'm not sure it's correct given you actually use it for other stuff even if we don't like it.)

zuiderkwast: We can go for the spec you came up with. We can always change it later if it poses any problems.

PR #447, opened 2022-08-30
Fix crash when a record union contains any()

### LINKED ISSUE #445: function_clause crash in typechecker:get_unassigned_fields/2
Another function_clause error:
```erlang
Error: function_clause
[{lists,map,
        [#Fun<typechecker.21.54609745>,{type,0,any,[]}],
        [{file,"lists.erl"},{line,1242}]},
 {typechecker,get_unassigned_fields,2,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,2013}]},
 {typechecker,type_check_fields,3,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,1964}]},
 {typechecker,type_check_record_union_in,3,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,3348}]},
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,2496}]},
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,2325}]},
 {typechecker,type_check_block_in,3,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,3310}]},
 {typechecker,check_clause,5,
              [{file,"/home/jespereskilson/dev/kred/_build/default/plugins/gradualizer/src/typechecker.erl"},
               {line,3586}]}]
```
Last message from Gradualizer when using `verbose`: 
```
97: Checking that PaccErrInfo#pacc_err_info{err = {OrigErr, Where}} :: #pacc_err_info{} | any()
Task reported error with trace from function parse_modified_error/1
```
The function `parse_modified_error/1` looks like this:
```erlang
-spec parse_modified_error(Error :: any()) -> #pacc_err_info{} | any().
parse_modified_error({{nocatch, PaccErrInfo = #pacc_err_info{}}, Where}) ->
    OrigErr = PaccErrInfo#pacc_err_info.err,
    PaccErrInfo#pacc_err_info{err = {OrigErr, Where}};
parse_modified_error(PaccErrInfo = #pacc_err_info{}) ->
    PaccErrInfo;
parse_modified_error({PaccErrInfo = #pacc_err_info{}, Where}) ->
    OrigErr = PaccErrInfo#pacc_err_info.err,
    PaccErrInfo#pacc_err_info{err = {OrigErr, Where}};
parse_modified_error(Reason) ->
    Reason.
```
Gradualizer version: 136f1ba5aba32c5cdaec4f6764e4897fbae57f96

erszcz:
Fixes #445.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -2462,7 +2462,10 @@ do_type_check_expr_in(Env, ResTy, {record, Anno, Name, Fields} = Record) ->
             {VarBinds, Cs2} = type_check_fields(Env, Rec, Fields),
             {VarBinds, constraints:combine(Cs1, Cs2)};
         {fields_tys, Tyss, Cs1} ->
-            case type_check_record_union_in(Env, Tyss, Fields) of
+            Tyss2 = lists:map(fun (?type(any)) -> get_record_fields(Name, Anno, Env);
+                                  (Other) -> Other
+                              end, Tyss),
+            case type_check_record_union_in(Env, Tyss2, Fields) of
                 none ->
                     {Ty, _VB, _Cs} = type_check_expr(Env#env{infer = true}, Record),
                     throw(type_error(Record, Ty, ResTy));
@@ -2493,7 +2496,10 @@ do_type_check_expr_in(Env, ResTy, {record, Anno, Exp, Name, Fields} = Record) ->
             {union_var_binds([VarBinds|VarBindsList], Env)
                 ,constraints:combine([Cs1, Cs2|Css])};
         {fields_tys, Tyss, Cs1} ->
-            case type_check_record_union_in(Env, Tyss, Fields) of
+            Tyss2 = lists:map(fun (?type(any)) -> get_record_fields(Name, Anno, Env);
+                                  (Other) -> Other
+                              end, Tyss),
+            case type_check_record_union_in(Env, Tyss2, Fields) of
                 none ->
                     {Ty, _VB, _Cs} = type_check_expr(Env#env{infer = true}, Record),
                     throw(type_error(Record, Ty, ResTy));
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -2462,10 +2462,7 @@ do_type_check_expr_in(Env, ResTy, {record, Anno, Name, Fields} = Record) ->
             {VarBinds, Cs2} = type_check_fields(Env, Rec, Fields),
             {VarBinds, constraints:combine(Cs1, Cs2)};
         {fields_tys, Tyss, Cs1} ->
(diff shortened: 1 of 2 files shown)


src/typechecker.erl
```diff
@@ -2462,7 +2462,10 @@ do_type_check_expr_in(Env, ResTy, {record, Anno, Name, Fields} = Record) ->
             {VarBinds, Cs2} = type_check_fields(Env, Rec, Fields),
             {VarBinds, constraints:combine(Cs1, Cs2)};
         {fields_tys, Tyss, Cs1} ->
-            case type_check_record_union_in(Env, Tyss, Fields) of
+            Tyss2 = lists:map(fun (?type(any)) -> get_record_fields(Name, Anno, Env);
+                                  (Other) -> Other
+                              end, Tyss),
+            case type_check_record_union_in(Env, Tyss2, Fields) of
```
zuiderkwast, line 2468: Why not add a clause in `type_check_record_union_in` instead?

```Erlang
type_check_record_union_in(Env, [?type(any) | Tyss], Fields) ->
    ... ;
```

zuiderkwast (commented):
Looks like this will solve the issue. I have some suggestion but you can choose which solution you prefer.

There is some redundant code here. What if `expect_record_type` could return `fields_tys` in more cases, e.g. in the case of no union (a single record type) and in the case of `any` as well, could that get rid of some duplication? (optional)

src/typechecker.erl
erszcz: We don't have `Name` and `Anno` there, so calling `get_record_fields` just seemed simpler where I did it. Not a big difference, though.

erszcz:
> There is some redundant code here. What if expect_record_type could return fields_tys in more cases, e.g. in the case of no union (a single record type) and in the case of any as well, could that get rid of some duplication? (optional)

Good point, we can simplify it all a bit. I'll make this change before we merge this.

src/typechecker.erl
zuiderkwast: OK

erszcz:
> There is some redundant code here. What if expect_record_type could return fields_tys in more cases [...]

Ok, I did give it a try, but it's no use. Whatever we gain in `do_type_check_expr_in` we lose in `add_type_pat`, which is another call site of `expect_record_type`, but unions are handled completely differently there so reusing their code path doesn't make sense.

> Why not add a clause in `type_check_record_union_in`?

But I did that, so there's a tiny bit less repetition :)

PR #450, opened 2022-09-01
Experimental constraint solver

erszcz:
This rebases #284 by @josefs onto the current master branch. It also adds some polymorphic function tests borrowed from [_Bidirectional typing for Erlang_](https://dl.acm.org/doi/10.1145/3471871.3472966) to show the solver works - at least in some cases.

Constraint solving is enabled by using a Gradualizer option `solve_constraints`. Without the option the behaviour is unchanged. I think this approach will allow to merge this PR early, yet make the functionality opt-in while it's being polished.

I'm making this PR a draft for now, since it shows some suspicious reports when running a self-check:

```
ebin/gradualizer_db.beam: The type 0..1114111 is not a subtype of {_TyVar-576460752303417119, _TyVar-576460752303417087}
ebin/gradualizer_db.beam: The type [char()] is not a subtype of {_TyVar-576460752303415487, _TyVar-576460752303415455}
```

In general, the number of warnings on a self-check has grown significantly with this change. However, that was to be expected.


erszcz:
Actually, I don't think `map_type_var` should pass - as the list element type is union of a map and an atom, I think the map pattern match might fail. I'll move it around as appropriate tomorrow.

erszcz:
I've rebased this on top of the current master and force pushed.

Solving constraints is currently enabled by a command-line flag or an option set via an attribute. It's enabled for the self-check, which bumps the number of errors above 100 lines - that's why CI fails. It's not a lot more - 116 lines - but still more than the self-imposed threshold of 100.

I'm also gathering examples that would be useful as tests (currently the `poly1.erl` file apart from the tests already committed under `test`), which allows me to understand the places where the type checker approach fails (apart from a lot of places currently just dropping constraints). One such example is:

```
%% false negative, i.e. there's an undetected bug here
-spec i([binary() | integer()]) -> [integer()].
i(L) ->
    map(fun
            (I) when is_integer(I) -> I * 2;
            (B) when is_list(B) -> list_to_integer(B)
        end, L).
```

`B` should be constrained to a binary, but this inconsistency is not caught. It's clear issues arise when union types, type refinement, and constraint solving overlap.

erszcz:
I think this is ready for review now. My rationale for merging this in is that on a feature branch it's subject to bit rot, whereas on master it will be regularly tested for compatibility with the rest of the project thanks to CI.

Main changes from #284:
- a number of tests, all test files enabling constraint solving start with `poly_` prefix for clarity
- locating a constraint error is more convenient:
  - constraint error message prints type variable name
  - type variable names are deterministic and descriptive - they start with the original type var name, e.g. `A_...` for a `(A) -> boolean()` function
  - constraint error message prints the location of the top-level function for which a constraint error was identified as that's the most precise location that's available to the type checker
  - see https://github.com/josefs/Gradualizer/actions/runs/3608900282/jobs/6081768978 ("Run self check") for an example output
- number of self-check error lines < 100, i.e. we're below the self-imposed CI limit
- constraint solving is opt-in, it requires a command-line `--solve_constraints` flag or an in-file Gradualizer `solve_constraints` attribute to be enabled

Constraint solving will certainly require more work to be fully functional, especially in the light of #488, but I think it's better to have whatever already works merged to master.

@josefs @zuiderkwast WDYT?

erszcz:
I've fixed a few more self-check errors. Do you have any comments on this, @josefs @zuiderkwast? Especially any comments vetoing a merge into master? I'd like to rebase this onto master and merge after #499.

src/typechecker.erl
```diff
@@ -4227,14 +4343,21 @@ check_guards(Env, Guards) ->
...
     case maps:find({Name, NArgs}, Env#env.fenv) of
         {ok, FunTy} ->
             NewEnv = Env#env{current_spec = FunTy},
             FunTyNoPos = [ typelib:remove_pos(?assert_type(Ty, type())) || Ty <- FunTy ],
             Arity = clause_arity(hd(Clauses)),
-            check_clauses_fun(NewEnv, expect_fun_type(NewEnv, FunTyNoPos, Arity), Clauses);
+            case expect_fun_type(NewEnv, FunTyNoPos, Arity) of
+                {type_error, NotFunTy} ->
+                    %% This can only happen if `create_fenv/2' creates garbage.
+                    erlang:error({invalid_function_type, NotFunTy});
```
zuiderkwast, line 4356: Dead code? Shoulnd't we let it crash instead?

test/known_problems/should_fail/poly_lists_map_should_fail.erl
```diff
@@ -0,0 +1,19 @@
...
+
+-export([i/1,
+         j/1]).
+
+-spec i([binary() | integer()]) -> [integer()].
+i(L) ->
+    lists:map(fun
+                  (I) when is_integer(I) -> I * 2;
+                  (B) when is_list(B) -> list_to_integer(B)
+              end, L).
```
zuiderkwast, line 13: Since these examples are slightly complex, it'd be interesting to see what errors message we expect here. Can you add it in a comment (or do you think it I should run it myself to find out)?

test/should_fail/poly_fail.erl
```diff
@@ -0,0 +1,25 @@
...
+lookup(K, [{K, V}|_]) -> V;
+lookup(K, [_|KVs])-> lookup(K, KVs).
+
+-spec find1() -> string().
+find1() ->
+    %% This should fail type checking.
+    "s" = lookup(0, [{0, 1}]).
+
+-spec poly_2(fun((A) -> A)) -> {integer(), boolean()}.
+poly_2(F) -> {F(42), F(false)}.
```
zuiderkwast, line 22: This one shouldn't fail. Should it?

zuiderkwast (commented):
The tests look good. It's nice to see that the solver does something useful.

I've only skimmed thought the implementation. I think it's a good idea to merge this with constraint solving off by default, as you suggested.

Is the solver implemented as if subtyping is transitive? This is what Josef mentioned, but I haven't really spotted this part of the implementation. It would be nice to add some example of what the solver can't do if it assumes typing is transitive, which it actually isn't when any() is involved. (X :: Y and Y :: Z doesn't imply X :: Z if Y is any()).

src/typechecker.erl
erszcz: If we let it crash we'll get a self-check warning. Throwing an error is more explicit and avoids the warning, though it's more verbose.

test/known_problems/should_fail/poly_lists_map_should_fail.erl
erszcz: Yeah, this one is a bit far fetched as it assumes smart enough inference of the inline fun's type. I imagine the error could look similar to the ones we currently report on this branch:

```
Lower bound binary() | integer() of type variable T_123 on line 456 is not a subtype of integer() | list()
```

We could also try to be smarter and infer the input type of the inline fun to be `integer() & list()`, but that's even more distant future.

However, more explicit examples would not rely on the inference of the inline fun's type at all, while still capture the problem. Let's take these which rely on just on top-level functions with specs:

```erlang
-spec i([binary() | integer() | string()]) -> [integer()].
i(L) ->
    lists:map(fun takes_a_union/1, L).

-spec takes_a_union(binary() | integer()) -> integer().
takes_a_union(B) when is_binary(B) -> binary_to_integer(B);
takes_a_union(I) when is_integer(I) -> I.
```

`i/1` should fail type checking, because the union it accepts is wider than `takes_a_union/1` allowed input type.

```erlang
-spec j([binary() | integer() | string()]) -> list().
j(L) ->
    lists:map(fun takes_an_intersection/1, L).

-spec takes_an_intersection(binary()) -> binary();
                           (integer()) -> integer().
takes_an_intersection(B) when is_binary(B) -> B;
takes_an_intersection(I) when is_integer(I) -> I.
```

`j/1` is similar, but it calls a function which takes an intersection type instead of a union.

I'll add these two simpler examples to tests, too.

src/typechecker.erl
zuiderkwast: OK

test/known_problems/should_fail/poly_lists_map_should_fail.erl
zuiderkwast: Great, just add that in a code comment then, something like

```Erlang
%% We expect an error like "Lower bound binary() | integer()
%% of type variable T_123 on line 456 is not a subtype of
%% integer() | list()"
-spec i...
```

test/should_fail/poly_fail.erl
erszcz: I don't feel completely comfortable with this topic yet, but I think it depends on the future goals: if we ever want to be able to infer the type of a function like `poly_2` (for example, when it's used as an inline fun), then we're limited to rank-2 polymorphism for which inference is decidable.

According to _Bidirectional Typing for Erlang_:

> Erlang allows higher-ranked polymorphism where polymorphic types occur anywhere in the type signature. The Hindley-Milner type inference only supports the inference of rank-1 (prenex) polymorphism. Type inference for the higher-ranked polymorphism is undecidable, which is a significant limitation for the Hindley-Milner type inference. For example, the `poly_2` function shown next has rank-2 polymorphism that ETC fails to type-check:
>
>     -spec poly_2(fun((A) -> A)) -> {integer(), boolean()}.
>     poly_2(F) -> {F(42), F(false)}.
>
> Similarly, ETC fails for `poly_fail/3` due to its rank-2 poly-
morphism.
>
>     -spec poly_fail(fun((A) -> A), boolean(), integer()) -> {boolean(), integer()}.
>     poly_fail(F, B, I) -> {F(I), F(B)}.
>
> In contrast, Dialyzer here finds the return type mismatch error.

As we can see from this test, Gradualizer is also able to detect this error.

I'd leave it as is until anyone of us is comfortable enough with constraint solving and polymorphism to confidently move it to `should_pass` tests.

erszcz:
I've merged #499 to master and rebased this PR on top of that.

The self-check has uncovered a number of issues with functions with multiple clauses registering too narrow constraints on type variables. I came up with the idea that an argument type var should be constrained with a union of types accepted by all of such a fun's clauses. Let's consider `flatmap/2` and the following example:

```erlang
  -spec flatmap(Fun, List1) -> List2
                   when
                       Fun :: fun((A) -> [B]),
                       List1 :: [A],
                       List2 :: [B].
```

```erlang
-type t1() :: {_, _, _, _}.
-type t2() :: {_, _, _}.

-spec f([t1() | t2()]) -> ok.
f(Fields) ->
    lists:flatmap(fun
                      ({_, _, _, _}) -> [t1];
                      ({_, _, _}) -> [t2]
                  end, Fields),
    ok.
```

When constraining A we must take all of the inline fun's clauses into account, i.e. constrain `A :: {_, _, _, _} | {_, _, _}`. The previous implementation used a constraint obtained from the first matching clause, which lead to Fields element `t1() | t2()` not type checking.

Another aspect of that can be seen in the following example:

```erlang
  -spec foldl(Fun, Acc, List) -> Acc
                 when
                     Fun :: fun((T, Acc) -> Acc),
                     List :: [T].
```

```erlang
-spec g([t1() | t2()]) -> ok.
g(Fields) ->
    lists:foldl(fun
                    ({_, _, _, _}, Acc) -> [t1 | Acc];
                    (_, Acc) -> [default | Acc]
                end, [], Fields),
    ok.
```

The `(_, Acc)` clause did not register any constraint on the first arg (`_`). Now, though, the catch-all clause is reflected in the union constraint as `any()`, i.e. the full constraint is `T :: {_, _, _, _} | any()`.

I was afraid the idea of a union upper bound constraint might be a bit fragile, but so far it seems to work quite well 👍

PR #451, opened 2022-09-13
Fix all known infinite loops

### LINKED ISSUE #449: Timeout checking function (performance issue related to jsx?)
I'm seeing a lot of `Timeout checking function` around in our code base:
```
./_build/test/lib/kafka_consumer_util/ebin/kafka_consumer_message.beam: Timeout checking function decode_json/1 on line 27
```
The only common pattern is that the functions mentioned in these errors all seem deal with either JSON or Avro data (using `jsx` or `erlavro`), at least the ones I have investigated a bit deeper. Could the rather large-ish type specs in those applications be causing Gradualizer performance issues?

erszcz:
This fixes #449 #231 and supersedes #376 and #324. In other words it fixes all known infinite loops in the type checker. This also means that all but one property tests pass, so I disabled the failing one, and made property test failures fail CI.

The main changes are in `normalize` and affect what's normalized and how deep. Maps are not normalized anymore, as well as records. Unions are normalized only on the top level, not recursively.

I also did some small changes which simplify troubleshooting:
- extracted `compat_seen`
- refactored `denormalize`
- switched from `sets` to `maps` in some places (maps are way more compact when displayed in traces, especially empty ones, and are backwards compatible until Erlang <20, whereas `sets` v2 are only available since Erlang 24.0)

On the first try I tried a different approach - https://github.com/erszcz/Gradualizer/tree/debug-jsx-infinite-loop-9 - but it was way more work, many more changes, and didn't give as good a result. It has led to this leaner and cleaner solution, though :)

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -4262,10 +4262,10 @@ add_types_pats([Pat | Pats], [Ty | Tys], Env, PatTysAcc, UBoundsAcc, CsAcc) ->
     UBound = denormalize(Ty, UBoundNorm, NormTy),
     add_types_pats(Pats, Tys, Env2, [PatTy|PatTysAcc], [UBound|UBoundsAcc], [Cs1|CsAcc]).
 
-denormalize(Ty, NormTy, OrigNormTy) ->
-    case NormTy of
-        OrigNormTy -> Ty;
-        _          -> NormTy
+denormalize(OrigTy, ComputedTy, NormTy) ->
+    case ComputedTy of
+        NormTy -> OrigTy;
+        _      -> ComputedTy
     end.
 
 %% Type check a pattern against a normalized type and add variable bindings.
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -790,13 +790,7 @@ normalize(Ty, Env) ->
 %% The third argument is a set of user types that we've already unfolded.
 %% It's important that we don't keep unfolding such types because it will
 %% lead to infinite recursion.
-%% TODO: shouldn't {type, _, tuple, Elems} also be normalized?
 -spec normalize_rec(type(), env(), map()) -> type().
-normalize_rec({type, _, record, [{atom, _, Name} | Fields]}, Env, Unfolded)
-  when length(Fields) > 0 ->
-    NormFields = [type_field_type(FieldName, normalize_rec(Type, Env, Unfolded))
-                  || ?type_field_type(FieldName, Type) <- Fields],
-    type_record(Name, NormFields);
 normalize_rec({type, _, union, Tys} = Type, Env, Unfolded) ->
     case maps:get(mta(Type, Env), Unfolded, no_type) of
         {type, NormType} -> NormType;
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -168,7 +168,7 @@ compatible(Ty1, Ty2, Env) ->
 
 -spec subtype(type(), type(), env()) -> compatible().
 subtype(Ty1, Ty2, Env) ->
-    try compat(Ty1, Ty2, sets:new(), Env) of
+    try compat(Ty1, Ty2, maps:new(), Env) of
         {_Memoization, Constraints} ->
             {true, Constraints}
     catch
@@ -189,9 +189,9 @@ any_subtype([Ty1|Tys], Ty, Env) ->
             any_subtype(Tys, Ty, Env)
     end.
 
--type acc(A) :: {A, constraints:constraints()}.
+-type acc(Seen) :: {Seen, constraints:constraints()}.
 
--type compat_acc() :: acc(sets:set(_)).
+-type compat_acc() :: acc(map()).
 
 % This function throws an exception in case of a type error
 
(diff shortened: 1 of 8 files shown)


erszcz:
Ahhh, apparently it's not ready for prime-time yet... There's a lot of these when running `make gradualize`:

```
ebin/gradualizer_db.beam: The function call on line 494 at column 50 is expected to have type type() but it has type type()
```

I'm switching the PR to a draft until it's fixed.

src/typechecker.erl
```diff
@@ -933,20 +934,15 @@ expand_builtin_aliases(Type) ->
 %% * Remove subtypes of other types in the same union; keeping any() separate
 %% * Merge integer types, including singleton integers and ranges
 %%   1, 1..5, integer(), non_neg_integer(), pos_integer(), neg_integer()
--spec flatten_unions([type()], env(), map()) -> [type()].
-flatten_unions(Tys, Env, Unfolded) ->
-    [ FTy || Ty <- Tys, FTy <- flatten_type(normalize_rec(Ty, Env, Unfolded), Env, Unfolded) ].
+-spec flatten_unions([type()], env()) -> [type()].
+flatten_unions(Tys, Env) ->
+    [ FTy || Ty <- Tys, FTy <- flatten_type(Ty, Env) ].
```
zuiderkwast, line 943: I guess we need to update the comment above this function to reflect the changed behaviour. (It is a few lines more than what's possible to select in a code review.)

Btw, flatten unions is still recursive in that it normalizes `A | (B | C)` into a flat union, right? But it doesn't try to normalize the elements in the union nor remove duplicates, is that right?

zuiderkwast (commented):
Very nice!!! :champagne:

Does this mean we now do what #376 tried to do, i.e. produce a Head Normal Form?

Nothing is broken by the fact that we don't normalize as much as we used to?

src/typechecker.erl
erszcz: > flatten unions is still recursive in that it normalizes A | (B | C) into a flat union, right?

Yes, I'll add some tests to make that explicit.

> But it doesn't try to normalize the elements in the union nor remove duplicates, is that right?

Exactly.

erszcz:
> Does this mean we now do what https://github.com/josefs/Gradualizer/pull/376 tried to do, i.e. produce a Head Normal Form?

AFAICT, that's the case, but a second pair of eyes can't hurt :)

erszcz:
I called it victory too early, I'm afraid :(

The self-check errors are really nasty and there's also quite a lot of them. Moreover, they exposed flaws in the proposed changes. Upon closer investigation I realised that the union flattening change was buggy. However, it was exactly that change that lead to all properties passing - they were terminating (the properties currently only check for termination), but returning incorrect results.

I've also investigated a little bit the normal forms and I think we're in a bit of confusion over terminology. According to [this article](https://medium.com/@aleksandrasays/brief-normal-forms-explanation-with-haskell-cd5dfa94a157) what you're referring to as [Head Normal Form](https://medium.com/@aleksandrasays/brief-normal-forms-explanation-with-haskell-cd5dfa94a157#:~:text=(Head)%20Normal%20Form,no%20thunks%20inside.) is actually [Weak Head Normal Form][weak-head-normal-form]. It gets tricky here, though. In our type representation, a `user_type` (even with no arguments) might be recursive. In literature, [a recursive type is denoted as _μx.t_][head-normal-form]. [We could express a recursive type in Erlang as `{rec, X, T}`](https://github.com/erszcz/tapl-erlang/blob/29acb1da9a824dee98484fdfb7f44f08d486b36b/apps/fullequirec/src/fullequirec_core.erl#L227) (this is my translation of an example type checker supporting equirecursive types from _Types and Programming Languages_ by B. Pierce). In Gradualizer, however, the references to recursive types are denoted as `{user_type, _, X, _}`, with `X` corresponding to the `X` in `{rec, X, T}`, but `T` only being available in `Env#env.tenv` or in `gradualizer_db`. If we interpret that that tuple as a recursive type constructor or abstraction, then it's already in Weak Head Normal Form. However, it cannot ever be in Head Normal Form, since that's a possibly infinite unfolding.

However, it gets even more tricky than that, since a `user_type` in Gradualizer is overloaded:
- it might be a recursive type as we already covered,
- it might be a generic type - then, the `Args` in `{user_type, _, _, Args}` come into play,
- due to the translation of `remote_type` in `normalize` it might refer to an external opaque type which should not be resolved further at all.

The infinite loops we've seen so far come from the interplay of sum types aka unions and recursive types. In other words, it's really tricky to tell when resolving a user type is correct and when it's not the right thing to do. For example, in `compat`:
- to avoid infinite loops we have to compare unresolved types
- but sometimes we have to cater for synthesised types, which are never user types, so `compat` has to expand user types to make a structural comparison.

To flatten and merge unions properly, though, we should resolve the user types which are legit to resolve (i.e. non-opaque ones), but we cannot do it indefinitely, as it might also lead to infinite loops, too.

All in all, I'm closing this one, but I have quite a bit of work in progress that I'll post as separate PRs.

[head-normal-form]: https://en.wikipedia.org/wiki/Recursive_data_type#:~:text=In%20type%20theory%2C%20a%20recursive%20type%20has%20the%20general%20form%20%CE%BC%CE%B1.T%20where%20the%20type%20variable%20%CE%B1%20may%20appear%20in%20the%20type%20T%20and%20stands%20for%20the%20entire%20type%20itself.

[weak-head-normal-form]: https://medium.com/@aleksandrasays/brief-normal-forms-explanation-with-haskell-cd5dfa94a157#:~:text=An%20expression%20is%20in%20weak%20head%20normal%20form%20when%20the%20outermost%20part%20has%20been%20evaluated%20to%20the%20data%20constructor%20or%20lambda%20abstraction.%20Sub%2Dexpressions%20may%20not%20have%20been%20evaluated.%20If%20all%20the%20sub%2Dexpressions%20would%20be%20evaluated%20then%20the%20expression%20would%20be%20in%20normal%20form.

josefs:
Just a brief comment here. Dealing with equi-recursive types was the first thing I tackled when I started working on Gradualizer because it was a problem I was not familiar with but I knew it had to be solved if Gradualizer was to be practical. I did indeed base my implementation on Chapter 21 Types and Programming Languages by B. Pierce. I looked at other papers but that book certainly has the simplest and most straight-forward algorithm for subtyping and equi-recursive types. 
I chose explicitly to not use a recursion operator (like your `{rec, X, T}`) because they don't scale very well. How do you handle mutual recursion for example? What if the mutual recursion goes across modules? It doesn't seem to work in practice.
Note that the algorithm in Pierce's book rely on *regular trees*. It doesn't actually need a recursion operator per se. So the important property is that when we unfold types, that they at some point become the same type again. That's so that the cache can recognise that we've seen the type before. 
I assume that the current normalisation and flattening mess things up slightly so that we don't see the same type again as we unfold things. 
One way to try to get at this problem would be to figure out what property is needed for normalization and write a property for it. Then it would be easier to debug with the help of proper.

erszcz:
Thanks for an insightful comment, @josefs!

> I chose explicitly to not use a recursion operator (like your `{rec, X, T}`) because they don't scale very well. How do you handle mutual recursion for example? What if the mutual recursion goes across modules? It doesn't seem to work in practice.
Note that the algorithm in Pierce's book rely on regular trees. It doesn't actually need a recursion operator per se.

Indeed, the interplay between `remote_type`, `user_type` and opaqueness complicates matters a lot in contrast with the simple implementation from TaPL. Comparing a real-world one and the one from the book is eye-opening! Thanks for confirmation that the recursive type representation was deliberate :+1:

In the meantime I carried on with looking for the needle in the haystack - #452 shows that the necessary fix was in `normalize` itself.

zuiderkwast:
> what you're referring to as [Head Normal Form](https://medium.com/@aleksandrasays/brief-normal-forms-explanation-with-haskell-cd5dfa94a157#:~:text=(Head)%20Normal%20Form,no%20thunks%20inside.) is actually [Weak Head Normal Form](https://medium.com/@aleksandrasays/brief-normal-forms-explanation-with-haskell-cd5dfa94a157#:~:text=An%20expression%20is%20in%20weak%20head%20normal%20form%20when%20the%20outermost%20part%20has%20been%20evaluated%20to%20the%20data%20constructor%20or%20lambda%20abstraction.%20Sub%2Dexpressions%20may%20not%20have%20been%20evaluated.%20If%20all%20the%20sub%2Dexpressions%20would%20be%20evaluated%20then%20the%20expression%20would%20be%20in%20normal%20form.)

Well, my interpretation is that the difference between HNF and WHNF is that WHNF can contain unevaluated thunks anywhere below the head while HNF can't. In Erlang we don't have lazy evauation and thunks, so HFN and WHNF are the same thing.

Anyway, the important thing is to normalize the head (i.e. don't normalize recursively), then compare two types on head level, do memoization (or cache) to break cycles and then go down into the branches. As Josef said, we rely on seeing the exact same type again to break cycles, including non-normalized subtrees.

erszcz:
> Well, my interpretation is that the difference between HNF and WHNF is that WHNF can contain unevaluated thunks anywhere below the head while HNF can't. In Erlang we don't have lazy evauation and thunks, so HFN and WHNF are the same thing.

I reinterpreted _thunk_ as the unevaluated part of the type term. The original article is about evaluation of terms, not types, but the same holds for type terms.

All clear apart from that, thanks for the comment :+1:

PR #452, opened 2022-09-14
Fix some infinite loops

### LINKED ISSUE #449: Timeout checking function (performance issue related to jsx?)

erszcz:
That's certainly a more humble title than #451, but also more appropriate :) This distills the minimum set of changes to fix #449 and #231.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -13,7 +13,7 @@ jobs:
     name: Erlang ${{matrix.otp}}
     strategy:
       matrix:
-        otp: ['22.3.4.20', '23.3.4.7', '24.1']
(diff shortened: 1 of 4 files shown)


josefs (approved):
This is good progress on fixing the loops. Thanks!

PR #454, opened 2022-09-14
Add gradualizer_tracer for efficient troubleshooting

erszcz:
This is some machinery I'm keeping on the git stash for when it's needed, but I think it doesn't hurt to share it here.

### DIFF
--- a/src/gradualizer_cli.erl
+++ b/src/gradualizer_cli.erl
@@ -11,10 +11,12 @@ main(Args) ->
             halt(1);
         {ok, Files, Opts} ->
             start_application(Opts),
-            case gradualizer:type_check_files(Files, Opts) of
-                ok -> ok;
-                nok -> halt(1)
-            end
+            R = case gradualizer:type_check_files(Files, Opts) of
+                    ok -> ok;
+                    nok -> halt(1)
+                end,
+            gradualizer_tracer:flush(),
+            R
     end.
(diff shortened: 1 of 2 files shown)


src/gradualizer_cli.erl
```diff
@@ -24,6 +26,8 @@ start_application(Opts) ->
     %% If gradualizer is run as an escript this should not be necessary, but better safe than sorry.
     ok = application:load(gradualizer),
     application:set_env(gradualizer, options, Opts),
+    %% We could start the tracer based on a CLI flag, but it's config is compile-time anyway.
+    %gradualizer_tracer:start(),
```
zuiderkwast, line 30: Is this tracing enabled by un-commenting this line?

It doesn't seem very convenient to me...

zuiderkwast (commented):
I'm fine that you add this if you think it's useful.

src/gradualizer_cli.erl
erszcz: Given the tracer has to be customised in code to make it useful for a practical test case / found bug, I think that uncommenting this single line is a minor detail :) Seriously, debugging the infinite loops, for example, took me literally hours of working with different examples, reading the code, comparing against hypothetical fixes on various branches, etc. Uncommenting this line IMO is not a blocker.

PR #455, opened 2022-09-14
Define g.erl for interative work with Gradualizer

erszcz:
This is another batch of convenience stuff for working with Gradualizer. We could say it expands on the idea of `user_default.erl` that's already present in `src`, but at least for now I placed the new file under `test`, so that's it doesn't pollute the module namespace for non-interactive use of Gradualizer as a dependency (that is, for ErlangLS).


erszcz:
See also a screenshot of how the docs would be rendered in #456.

zuiderkwast (commented):
I'm not sure I like the module name `g`. Can't we place these utility functions in `gradualizer`?

erszcz:
The idea is to make it handy in the shell, similar to `r3`, that's why the terseness.
Another idea is that if we use that convention, the we could rewrite the tests in `test/typechecker_tests.erl` from:

```erlang
     ?_assert(subtype(?t( pos_integer()     ), ?t( non_neg_integer()))),
     ?_assert(subtype(?t( 0..5              ), ?t( non_neg_integer()))),
```

to something like:

```erlang
     ?_assert(subtype(g:type( "pos_integer()"     ), g:type( "non_neg_integer()" ))),
     ?_assert(subtype(g:type( "0..5"              ), g:type( "non_neg_integer()" ))),
```

The reason is that the former cannot be copy-pasted to a shell to troubleshoot / debug in isolation, but the latter can. Now, we could use an `-import(gradualizer, [...])`, but then the call sites wouldn't be copy-pastable again, so a short module name (not a macro) offers the best source-vs-shell compatibility.

KennethL:
Maybe shell_default and user_default can be used to make things nice in the shell. See doc for shell_default

erszcz:
I implemented [`h()` for documentation access in docsh](https://github.com/erszcz/docsh/blob/b6d917e75270cd3220300fc00583e4d3cfa8b460/include/docsh_user_default.hrl) before `shell_docs` existed as a `user_default` extension and I think it's a viable option. However, I think `user_default` itself doesn't scale well in some scenarios, especially when working across multiple projects/repos:
1. When `user_default` is compiled with a newer Erlang version, but we switch to a project using an old enough version, we have to recompile `user_default` and when we switch back, we have to recompile again.
2. When using `user_default` we have to manage the Erlang code path manually, for example we have to put Gradualizer in `ERL_LIBS`. However, this might lead to issues when both developing Gradualizer (when I want to use the local copy to test it) and using Gradualizer (as part of ErlangLS, which is invoked in the background of a text editor - here we would like a stable version to run, not the one from `ERL_LIBS`).
3. When we want to use different shell extensions in different Erlang projects/repos. We could still resort to something like https://direnv.net/ to manage the path to different `user_default` instances, but then we fall again into point 2., i.e. managing `ERL_LIBS`.

I think we can provide a `user_default` file for people who prefer using it, but I also think a standalone module like the one proposed in this PR together with Rebar3 solve all of the above issues.

erszcz:
I've moved the utils to `gradualizer` and adjusted the module description. What do you think, @zuiderkwast?

src/gradualizer.erl
```diff
@@ -227,11 +276,82 @@ add_source_file_and_forms_to_opts(File, Forms, Opts) ->
...
+    type_of(Expr, env([infer])).
+
+%% @doc Infer type of an Erlang expression.
+%%
+%% ```
+%% > gradualizer:type_of("[ a || _ <- lists:seq(1, 5) ]").
+%% {type,0,list,[{atom,0,a}]}
+%% '''
+-spec type_of(string(), typechecker:env()) -> typechecker:type().
+type_of(Expr, Env) ->
```
zuiderkwast, line 354: Add an example where type_of/2 and env/1 are used together? They are only useful together in practice, right? `gradualizer:type_of("...", gradualizer:env("..."))`.

Btw, don't we need to expose pretty-printing of types too for this to be useful without reading abstract internal representations?

zuiderkwast (commented):
I like that these have moved to `gradualizer`.

src/gradualizer.erl
erszcz: > They are only useful together in practice, right?

Indeed. Example added.

> don't we need to expose pretty-printing of types too?

Pretty-printing is already exported, see the module doc at the top:

```
> gradualizer:type_of("fun (A) -> #{tag => my_map, list_of_as => [ A || _ <- lists:seq(1, 5) ]} end").
{type,0,'fun',
      [{type,0,product,[{type,0,any,[]}]},
       {type,0,map,
             [{type,0,map_field_assoc,[{atom,0,tag},{atom,0,my_map}]},
              {type,0,map_field_assoc,
                    [{atom,0,list_of_as},{type,0,list,[{type,0,any,[]}]}]}]}]}
> typelib:pp_type(v(-1)).
"fun((any()) -> #{tag => my_map, list_of_as => [any()]})"
```

erszcz: I've also realised that defining variable bindings and variable typings in the env would make `type_of/1,2` much more powerful, but I guess we have to start somewhere. For now I added a comment that defining varbinds is not yet possible with `env/1,2`.

PR #457, opened 2022-09-19
Fix map type inference

### LINKED ISSUE #378: Type of nested map creation expression is not inferred correctly
@eksperimental in https://github.com/esl/gradient/issues/30 reports an interesting bug, which I minimised to the following Elixir example:

```
iex(3)> Gradient.type_check_file("_build/dev/lib/beam_langs_meta_data/ebin/Elixir.Test.beam")
lib/test.ex: The map on line 10 is expected to have type %{required(:inner) => %{required(:a) => 5}} but it has type %{required(:inner) => %{optional(:a) => 5}}
8   @spec g :: %{inner: %{a: 5}}
9   def g do
10     %{inner: %{a: 5}}
11  end

iex(1)> Gradient.Debug.print_erlang(Test)
-module('Elixir.Test').

-spec g() -> #{inner := #{a := 5}}.
g() -> #{inner => #{a => 5}}.
```

It seems that nested map creation expressions don't get their type inferred properly, i.e. the associations are inferred as optional, whereas after map creation we know they must be present in the resulting map, therefore they should be inferred as required. One idea is that [we might need some recursion in `update_map_type`](https://github.com/josefs/Gradualizer/blob/87bb404113f3c9ef43715f0fcda8df6d408683e1/src/typechecker.erl#L2334-L2346), but this has to be confirmed.

### LINKED ISSUE #432: Map type inference not working when map creation expression is an argument
As a followup to #429 I've stumbled upon `test/should_fail/covariant_map_keys_fail.erl`. There's no error reported for `kaboom/0` by default (even with the `--infer` flag), but there is if we apply the following change:

```
$ git diff
diff --git a/test/should_fail/covariant_map_keys_fail.erl b/test/should_fail/covariant_map_keys_fail.erl
index 99c5883..98e38f0 100644
--- a/test/should_fail/covariant_map_keys_fail.erl
+++ b/test/should_fail/covariant_map_keys_fail.erl
@@ -10,4 +10,6 @@ good(#{ good := X }) -> X.
 not_good(M) -> good(M). %% This call should fail

 -spec kaboom() -> integer().
-kaboom() -> not_good(#{ bad => 0 }).
+kaboom() ->
+    M = #{ bad => 0 },
+    not_good(M).
```

This shows that map type inference does not always work as expected. It might or might not be related to #378.

erszcz:
This fixes #432 and fixes #378.

### DIFF
--- a/test/should_fail/covariant_map_keys_fail.erl
+++ b/test/should_fail/covariant_map_keys_fail.erl
@@ -3,11 +3,19 @@
 -compile([nowarn_unused_function]).
 -export([not_good/1]).
 
--spec good(#{ good := A }) -> A.
-good(#{ good := X }) -> X.
+-gradualizer([infer]).
+
+-spec good(#{good := A}) -> A.
+good(#{good := X}) -> X.
 
 -spec not_good(#{good | bad := A}) -> A.
 not_good(M) -> good(M). %% This call should fail
 
--spec kaboom() -> integer().
-kaboom() -> not_good(#{ bad => 0 }).
+-spec kaboom1() -> integer().
+kaboom1() ->
+    M = #{bad => 0},
+    not_good(M).
+
+-spec kaboom2() -> integer().
+kaboom2() ->
+    not_good(#{bad => 0}).
--- a/test/should_fail/covariant_map_keys_fail.erl
+++ b/test/should_fail/covariant_map_keys_fail.erl
@@ -1,21 +1,9 @@
 -module(covariant_map_keys_fail).
 
--compile([nowarn_unused_function]).
 -export([not_good/1]).
-
--gradualizer([infer]).
 
 -spec good(#{good := A}) -> A.
 good(#{good := X}) -> X.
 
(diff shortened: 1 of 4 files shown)


test/should_fail/covariant_map_keys_fail.erl
```diff
@@ -3,11 +3,19 @@
 -compile([nowarn_unused_function]).
 -export([not_good/1]).
 
--spec good(#{ good := A }) -> A.
-good(#{ good := X }) -> X.
+-gradualizer([infer]).
```
josefs, line 6: The test suite is a little disorganized wrt `infer`. But I think it would be helpful to have separate files for things which are tested with `infer` enabled and have `infer` in the name.

zuiderkwast (commented):
Looks good in general. The fix looks simpler than I though.

test/should_fail/covariant_map_keys_fail.erl
erszcz: Good point, I'll split out the parts relying on `infer`.

PR #458, opened 2022-09-21
Fix more infinite loops

erszcz:
Ok, this might finally be the PR solving all the infinite loops we're aware of. At least that's what the property tests indicate - they don't fail in CI and they don't fail on my machine even with `PROP_NUMTESTS=50000`. All the looping examples we've seen so far are also handled properly, i.e. terminate.

The idea allowing this to happen is the realisation that a once normalized type should be a fixpoint under `normalize()` - even a union type. This was already suggested by one of the normalize tests - `normalize(boolean()) == normalize(normalize(boolean))` - but not realised fully, i.e. not for unions. Newly added `normalize(union()) == normalize(normalize(union()))` ensures that's the case even for union types.

But how to achieve that? [Based on the insight that a `user_type` is overloaded and means multiple things](https://github.com/josefs/Gradualizer/pull/451#issuecomment-1246692967), we just don't recursively normalise user types inside unions. This, in turn, shows that we don't need `Unfolded` in `normalize()` at all.

This means there are a few places (and we might find a few more) where we have to normalize these unexpanded user types to get their definitions. Once that's done, though, all the tests pass and all property tests terminate.

There are two more, somewhat connected, results:
1. We've regressed a bit and report more warnings from `make gradualize` - it was a bit of a challenge to make it pass the CI check which fails at 100 and more lines of warnings.
2. We can see that the `erl_parse` type representation we've adopted is problematic.

Why is `type()` problematic?

Firstly, the "flexibility" to pass around `type()` or `[type()]`, the latter expressing multiple function clauses, starts to raise hard to fix type errors with a picky type checker. It's already problematic due to special cases that have to check for a list in place of a type, but now it's painfully visible on the type/spec layer, too.

Secondly, the AST nodes pack node details into lists. Lists are not positional, so unpacking them leads to vars having union types, which require manual refinement. For example, a remote type AST node is:

```erlang
-type af_remote_type() ::
        {'remote_type', anno(), [(Module :: af_atom()) |
                                 (TypeName :: af_atom()) |
                                 [abstract_type()]]}. % [Module, Name, [T]]
```

Our code handling it must refine the element type to make the type checker happy:

```erlang
normalize_rec({remote_type, _, [{atom, _, M}, {atom, _, N}, Args]}, Env) ->
    %% It's safe as we explicitly match out `Module :: af_atom()' and `TypeName :: af_atom()'.
    Args = ?assert_type(Args, [type()]),
```

Using tuples instead of lists would be much more convenient as tuple fields are typed explicitly.

Going in the other direction is also problematic. For example:

```erlang
    RetTy = if
                Env#env.infer ->
                    %% Infer the size parameters of the bitstring
                    gradualizer_bin:compute_type(BinExpr);
                not Env#env.infer ->
                    type(any)
            end,
    RetTy = ?assert_type(RetTy, type()),
```

Without the explicit assertion the type checker would not understand that `RetTy` can be returned as a `type()`. One case was completely fruitless - no amount of assertions and twisting the code allowed `FunTy` to pass as a `type()`:

```erlang
type_check_fun(Env, {remote, _, _Expr, _}, Arity) ->
    % Call to an unknown module. Revert to dynamic types.
    P = erl_anno:new(0),
    AnyArgs = ?assert_type(lists:duplicate(Arity, type(any)), [type()]),
    FunTy = {type, P, bounded_fun,
             [{type, P, 'fun',
               [{type, P, product, AnyArgs}, type(any)]},
              []]},
    {[FunTy], Env, constraints:empty()};
```

All in all, though, infinite loops and timeout warnings should be gone now.


erszcz:
BTW, this finally addresses #376.

src/gradualizer_tracer.erl
```diff
@@ -109,25 +114,32 @@ start() ->
     %dbg:tpl(typechecker, refine_clause_arg_tys, x),
     %dbg:tpl(typechecker, refine_mismatch_using_guards, x),
 
-    %dbg:tpl(typechecker, expect_tuple_type, x),
+    dbg:tpl(typechecker, expect_tuple_type, x),
```
josefs, line 117: Comment this out?

src/typechecker.erl
```diff
@@ -1170,17 +1162,17 @@ infer_literal_string(Str, Env) ->
                                {char, erl_anno:new(0), lists:last(SortedChars)}])])
     end.
 
-expect_tuple_type({type, _, any, []}, _N) ->
+expect_tuple_type({type, _, any, []}, _N, _Env) ->
```
josefs, line 1165: The `_Env` parameter that you've added doesn't seem to be used anywhere. 
I suggest adding a spec to this function, while you're at it.

josefs (commented):
This is amazing! It's nice to see the simplification in `normalize`.

src/typechecker.erl
erszcz: It's used in only one clause, but this clause is the reason for introducing `Env`:

```
expect_tuple_type(?user_type() = Ty, N, Env) ->
    expect_tuple_type(normalize(Ty, Env), N, Env);
```

erszcz: It's one of those places where we need to call `normalize()` explicitly now.

zuiderkwast:
This looks awesome! Just replying to some of the comments:

> Why is `type()` problematic?
>
> Firstly, the "flexibility" to pass around `type()` or `[type()]`, the latter expressing multiple function clauses

I think there were plans to replace `[type()]` with a "bounded fun type" which can contain multiple clauses (a function type containing a product type for the params, etc.).

Btw, I think we need to distinguish between value types and function signatures. A `-type` can't expression everything that a `-spec` can express and we should be explicit about what we want here --- either we extend the `type()` definition to include bounded function types or we distinguish between the two. It could be worth opening a separe issue for this.

> Secondly, the AST nodes pack node details into lists. Lists are not positional,

Yeah, that's one of the cons of using OTP's type definitions (with minor modifications). If we decide to move away from this representation, we'd definitely not have a list in this way.

>     RetTy = if
>                 Env#env.infer ->
>                     %% Infer the size parameters of the bitstring
>                     gradualizer_bin:compute_type(BinExpr);
>                 not Env#env.infer ->
>                     type(any)
>             end,
>     RetTy = ?assert_type(RetTy, type()),

Shouldn't we be able to infer the type of an `if` expression as the union of the return of the if clauses?

> One case was completely fruitless - no amount of assertions and twisting the code allowed `FunTy` to pass as a `type()`

Yeah, this is the area of bounded fun types that is quite messy. There is room for improvements in this area.

src/typechecker.erl
erszcz: Spec added :+1:

PR #459, opened 2022-09-24
Add install-asdf rule to Makefile

erszcz:
This makes it possible to easily install Gradualizer globally if we use [asdf](https://asdf-vm.com/) to manage Erlang installations. Pretty convenient and not a lot of code added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,3 +230,14 @@ check: tests dialyze gradualize
 
 .PHONY: travischeck
 travischeck: cover/coverage.xml dialyze nocrashongradualize
+
+.PHONY: install-asdf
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
I'm fine with this, although I'm not familiar with asdf.

erszcz:
It's a generalised `kerl` or `kiex`. Generalised means it can manage many different tools and multiple versions of them, so that switching between tool versions (e.g. Erlang versions) depending on a project/repo is easy.

Taken to the extreme, it can use `.tool-versions` files to set specific versions which should be used in a given project, e.g. for Erlang, Elixir, Kubernetes kubectl, etc.

PR #460, opened 2022-09-24
Housekeeping

erszcz:
More housekeeping, no significant changes here, just some small refactorings, spec and type fixes, etc.

Self-check errors down to 42 lines:
```
$ wc -l gradualize.f4d111f.log
      42 gradualize.f4d111f.log
```

### DIFF
--- a/src/absform.erl
+++ b/src/absform.erl
@@ -4,6 +4,8 @@
 -export([normalize_record_field/1,
          normalize_function_type_list/1]).
 
+-include("gradualizer.hrl").
+
 %% @doc Turns all record fields into typed record fields. Adds default
 %% 'undefined' if default value is missing.
 normalize_record_field({record_field, L, Name = {atom, _, _}}) ->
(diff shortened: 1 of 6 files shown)


src/absform.erl
```diff
@@ -27,13 +29,15 @@ normalize_record_field({typed_record_field,
 
 %% @doc Turns all function types into bounded function types. Add default empty
 %% constraints if missing.
+-spec normalize_function_type_list(FunTypeList) -> FunTypeList when
+      FunTypeList :: gradualizer_type:af_function_type_list().
 normalize_function_type_list(FunTypeList) ->
-    lists:map(fun normalize_function_type/1, FunTypeList).
+    ?assert_type(lists:map(fun normalize_function_type/1, FunTypeList), [B, ...]).
```
josefs, line 35: It'd be nice to propagate the fact that the list is non-empty through functions like `lists:map`. It would likely need special handling in the type checker though, similar to how we treat arithmetic operators.

src/gradualizer_lib.erl
```diff
@@ -159,14 +164,17 @@ pick_value(?type(non_neg_integer), _Env) ->
 pick_value(?type(pos_integer), _Env) ->
     {integer, erl_anno:new(0), 0};
 pick_value(?type(neg_integer), _Env) ->
-    {integer, erl_anno:new(0), -1};
+    L = erl_anno:new(0),
+    {op, L, '-', {integer, L, 1}};
```
josefs, line 168: Just out of curiosity, why is this an improvement?

josefs (approved):
LGTM! Good effort to reduce the number of self-check errors.

src/gradualizer_lib.erl
erszcz: If we're strict, and Gradualizer is, then `abstract_expr()` doesn't allow negative numbers - we always deal with non-negatives, possibly wrapped in unary op nodes. I'm just applying feedback from Gradualizer itself - the spec says we return an `abstract_expr()`, so let's do it.

src/absform.erl
erszcz: I think we could rely on multiple spec clauses to do it like this:

```erlang
-spec lists:map(fun(), nonempty_list()) -> nonempty_list();
               (fun(), list()) -> list().
```

Then it would be a matter of providing a few overrides in prelude. The same would be nice for proper and improper lists - we touched on that in the thread of https://github.com/josefs/Gradualizer/pull/440#discussion_r957586771.

I think we just have to make sure the function intersection types work well enough to handle such specs. I have something in progress in #461. In that draft this code

```erlang
-spec foo(bar) -> bar;
         (list()) -> baz | qux.
foo(bar) -> bar;
foo([]) -> baz;
foo([a]) -> qux.
```

is already accepted, but this

```erlang
-type t() :: {tag, any()}.

-spec f(t()) -> string();
       ([t()]) -> string().
f(Types) when is_list(Types) -> lists:map(fun f/1, Types);
f({tag, _}) -> "".
```

still fails. Still working on it.

zuiderkwast: `lists:map(fun f/1, Types)` returns a nested list of strings, which isn't in line with the spec. Is this why it fails?

erszcz: No, no, it fails checking the intersection:

```
test/should_pass/spec_and_fun_clause_intersection_pass.erl:
The pattern {tag, _} on line 16 at column 3 doesn't have the type [t()]
```

erszcz: Good point anyway! I think we'll only catch that with the constraint solver in place, though.

Issue #446, opened 2022-08-30
Warn on every cast from any() to a well known type?


erszcz:
I'm wondering if warning for all casts from `any()` to a static type isn't the only sensible conclusion from the discussion here - https://github.com/josefs/Gradualizer/pull/429#discussion_r920837518. It could be possibly guarded by a CLI flag.

The rationale is that given we can't inject checks/guards/assertions, raising a warning is the only remaining option. Later on we could possibly detect some constructs ensuring runtime safety. For example, if we start with `X :: any()`, then we know that `if is_integer(X) -> X; true -> erlang:error() end :: integer()`, so a warning wouldn't have to be generated.

zuiderkwast:
We can add a warning guarded by an option, perhaps off by default to start with(?)

Another possibility is to start generating Erlang code with assertions inserted. Used in this way, Gradualizer should be used as a step before the compiler.

PR #461, opened 2022-09-24
Function intersections

erszcz:
So yesterday @josefs remarked in https://github.com/josefs/Gradualizer/pull/460#discussion_r979279858 that it would be nice to preserve (non)emptiness through functions like `lists:map/2`. We touched on a similar note in a discussion with @zuiderkwast in https://github.com/josefs/Gradualizer/pull/440#discussion_r957586771, but with regard to the (non)proper property of lists. It seems both of these goals could be achieved with better support for function type intersections or, in other words, better handling of multiple spec clauses.

Apparently, just providing a spec override like this for `lists:map/2` (which this PR does) is sufficient to be able to preserve (non)emptiness without other changes in the type checker:

```erlang
-spec map(fun((A) -> B), [A, ...]) -> [B, ...];
         (fun((A) -> B), [A]) -> [B].
```

However, if we want to reap the benefits of such specs in call sites it would be nice to handle checking functions declaring them, too. This is what this PR tries to achieve. Let's see some examples:

```erlang
-spec f(bar) -> bar;
       (list()) -> baz | qux.
f(bar) -> bar;
f([]) -> baz;
f([_|_]) -> qux.

-spec g([t()]) -> string();
       (t()) -> string().
g([] = _Types) -> [];
g([_|_] = Types) -> lists:map(fun g/1, Types);
g({tag, _}) -> "".
```

In order to do it with reasonable effort on the implementation side, though, it's tempting to stray away a bit from the official Erlang docs and assume that the spec clauses have to be in the same order as the corresponding function clauses matching the specs. Moreover, this implementation doesn't care if the specs overlap, but if they do they should be listed in the strictest to most general order (like in the `map` example - `[A, ...]` first, then `[A]`). This is NOT valid with this PR's implementation:

```erlang
-spec g(t()) -> string();
       ([t()]) -> string().
g([] = _Types) -> [];
g([_|_] = Types) -> lists:map(fun g/1, Types);
g({tag, _}) -> "".
```

All the above examples are already included in tests.

~I think the code in this PR can also handle exhaustiveness checking with such multiple clause specs, but it's not done yet.~ I'm sharing this already to understand if we all agree that's the right direction. WDYT?


erszcz:
BTW, this PR has the lowest number of self-check lines since we started tracking it:

```
$ wc -l gradualize.27730f9.log
      34 gradualize.27730f9.log
```

erszcz:
https://github.com/josefs/Gradualizer/pull/461/commits/54d16c0d8d5ca0ec5888abe798f5f78753c634cb shows how we apply this to self-check a very nasty case.

erszcz:
I'm marking this as ready for review, since the idea is fleshed out completely. The only remaining things to do would be:
- proper constraint handling (a bit hard to check without the solver, but now the constraints are just dropped)
- possibly refactoring to leave only one path through `check_clauses`.

josefs:
Requiring that the intersection types follow the order of pattern matching is quite a big change. My philosophy has so far been for Gradualizer to be as lenient as possible and not have the types get in the way unnecessarily. Requiring a specific order of intersection types would change that. 
But I don't feel like I understand the tradeoffs here. I might be convinced that the cost is worth the benefits. Are there problems that are unsurmountable if we allow any order of intersections? 
How do programmers use intersection types right now in the wild? Is it a common pattern to follow the pattern matching clauses? Some kind of impact assessment would be nice; for instance, trying it out on the top 24 hex packages :-)

erszcz:
I tried flexing the coding muscles a bit more to make the spec clause and function clause order independent, but the results are so so. First, already complex code got even more complex. Second, there are two tests failing and I'm not sure there's anything we can do about it... I'll describe it in more detail tomorrow.

erszcz:
Ok, so generally when the spec aka fun type clauses and function clauses are in order the implementation looks roughly like this (oversimplified pseudo-Erlang):

```erlang
check_clauses([] = SpecClauses, [] = FunClauses) ->
    ok;
check_clauses([] = SpecClauses, [_|_] = FunClauses) ->
    throw(unreacahble_clauses);
check_clauses([_|_] = SpecClauses, [] = FunClauses) ->
    throw(nonexhaustive_patterns);
check_clauses([SpecClause | SpecClauses], [FunClause | FunClauses]) ->
    try
        RefinedSpecClause = check_clause(SpecClause, FunClause),
        case RefinedSpecClause of
            ?type(none) ->
                check_clauses(SpecClauses, FunClauses);
            _ ->
                check_clauses([RefinedSpecClause | SpecClauses], FunClauses)
        end
    catch
        TypeError ->
            check_clauses(SpecClauses, [FunClause | FunClauses])
    end.
```

That is:
- we go from left to right through both clause lists
- we might consume multiple `FunClauses` for each `SpecClause`
- we go to the next `FunClause` when the current one type checks
- we go the the next `SpecClause` when the current one is refined to `none()` or if we cannot type check the current `FunClause` against the current `SpecClause`
- there's no accumulator of any kind
- we finish when both clause lists are empty
- linear complexity.

If we allow mixed order of spec clauses and function clauses the code is more like this:

```erlang
init_check_clauses(SpecClauses, FunClauses) ->
    SpecArgs = [ args(SpecClause) || SpecClause <- SpecClauses ],
    Acc = {_Seen = #{},
           _RefinedSpecClauses = maps:from_list(lists:zip(SpecArgs, SpecArgs)),
           _OrigSpecClauses = SpecClauses},
    check_clauses(SpecClauses, FunClauses, Acc).

check_clauses(_ = SpecClauses, [] = FunClauses, Acc) ->
    case all_refined_spec_clauses_are_none(Acc) of
        true ->
            ok;
        false ->
            throw(nonexhaustive_patterns)
    end;
check_clauses([SpecClause | SpecClauses], [FunClause | FunClauses], Acc) ->
    try
        already_seen(SpecClause, FunClause, Acc)
            andalso throw(unreachable_clauses),
        StoredRefinedSpecClause = refined_spec_clause(SpecClause, Acc),
        MoreRefinedSpecClause = check_clause(StoredRefinedSpecClause, FunClause),
        Acc = store_refined_spec_clause(SpecClause, MoreRefinedSpecClause, Acc),
        case MoreRefinedSpecClause of
            ?type(none) ->
                check_clauses(SpecClauses, FunClauses, Acc);
            _ ->
                check_clauses([MoreRefinedSpecClause | SpecClauses], FunClauses, Acc)
        end
    catch
        TypeError ->
            Acc = store_seen(SpecClause, FunClause, Acc),
            case SpecClauses of
                [] ->
                    check_clauses(orig_spec_clauses(Acc), [FunClause | FunClauses], Acc);
                [_|_] ->
                    check_clauses(SpecClauses, [FunClause | FunClauses], Acc)
                end
    end.
```

In words:
- we cycle through spec clauses until we run out of fun clauses to check
- we have to carry a complex accumulator
- if we "run out of" spec clauses we start with the original list, but mark the current spec and fun clauses as already seen - if we see this pair again, it means we have gone through the entire spec clause list without moving forward - which means failure (in the actual implementation I also carry the error caught on the first visit to the spec and fun clause pair to rethrow it, but I omitted it above) 
- since spec clauses are not in order of fun clauses, we have to store refined args types in the acc - this means a lot of packing and unpacking - without that we would never be able to tell that a given spec clause's args are exhausted
- more complex logic
- quadratic complexity, where each step is relatively heavyweight - it's checking an entire fun clause against a spec clause.

All of the above are just nuisances, though.

The unsolved problem is that there's no way to reliably check exhaustiveness, because we might run into a situation where there are more spec clauses than fun clauses - some examples from our tests:

```erlang
-spec number(integer()) -> integer();
            (float()) -> float().
number(N) -> N.

-spec i(a, b) -> {a, b};
       (d, e) -> {d, e}.
i(V, U) -> {V, U}.
```

In such a case there's just not enough steps taken through fun clauses to register a refined type for each spec clauses' args. Therefore the exhaustiveness check fails. I don't see a clever way to work around that. I also don't see a convenient convention to suggest to the end users to avoid that. At least not a more convenient one than "just put the specs and pattern matches in the same order".

What do you think?

josefs:
Thanks for the fantastic pseudocode! Very helpful!

In your second implementation you recurse over the function clauses, by which I mean that for each recursive call, you look at one function clause in return. I'd imagined that one could recurse over the spec clauses instead. In some sense I think that would be more natural. Since the meaning of intersection types is that the function must have *all* the types in the spec, then a naive way to implement that is to just check the function against each spec clause, just as if they were single spec clauses. I hope that makes sense, and apoligies for not following your example and providing elegant pseudocode.
What about exhaustiveness checking? When checking against one of the specs I think that one can keep track of which function clauses have been covered by previous specs. It could just be a boolean for each function clause. Though there are no doubt examples where that would be too coarse information.

I hope the algorithm I outlined makes sense. What do you think?

zuiderkwast:
I don't think we should depend on the spec clauses and the function clauses to be in the sam order, so I think version 2 sounds good although more complex than version 1. I don't know if iy matters whether we recurse over spec first or function clauses first. It seems we need to keep track of stuff regardless. If the known problems are solved, I'm happy.

We should take care that in some cases we can't do exhaustiveness-checking, e.g. when there are complex guards (meaning our exhaustiveness checking is best effort rather than sound, I suppose).

erszcz:
Ok, finally an update on this one. I've tried following:

> one could recurse over the spec clauses instead

and it seems to have worked. The code's a mess now, so it definitely needs a cleanup, but it passes all the tests and works fine with any order of spec clauses 🚀

test/should_pass/spec_and_fun_clause_intersection_pass.erl
```diff
@@ -0,0 +1,39 @@
...
+g1([_|_] = Types) -> lists:map(fun g1/1, Types);
+g1({tag, _}) -> "".
+
+%% Order of spec clauses doesn't match the order of patterns
+%% in function clauses and function clause patterns are mixed - will that pass?
+-spec g2(t()) -> string();
+        ([t()]) -> string().
+g2([] = _Types) -> [];
+g2({tag, _}) -> "";
+g2([_|_] = Types) -> lists:map(fun g2/1, Types).
```
zuiderkwast, line 34: This clause returns a list of strings if I read correctly. Thus, it should not pass.

Or is a string() allowed to be a nested list nowdays? I thought string() :: [char()].

src/typechecker.erl
```diff
@@ -4162,12 +4177,22 @@ type_check_function(Env, {function, _, Name, NArgs, Clauses}) ->
...
     end.
 
+-spec clause_arity({clause, _, list(), _, _}) -> arity().
+clause_arity({clause, _, Args, _, _}) ->
+    arity(length(Args)).
+
+-spec arity(non_neg_integer()) -> arity().
+arity(I) ->
+    ?assert(I < 256, arity_overflow),
+    ?assert_type(I, arity()).
```
zuiderkwast, line 4194: Cool.

Is this what we ideally would want ?assert_type to do for all types in runtime?

zuiderkwast (commented):
It looks promising! It's a bit hard to review though.

When do we use `{intersection, ...}` and when do we use a list of types after this?

test/should_pass/spec_and_fun_clause_intersection_pass.erl
zuiderkwast: Ah, that's why. So use `list()` instead of `string()` in the spec?

src/typechecker.erl
erszcz: Well, a bit tricky question. If we use `mfa()` or `arity()` in any specs or types, and `erl_parse:abstract_type/0` does, then we run into the problem of convincing the type checker that any arbitrary integers we try to use as arities make sense, i.e. fall into 0..256 range. We could do the type assertion without the dynamic check, but it could be unsound. The way it's done here we have the assertion in place for testing, but we can still drop the dynamic check for release by defining `NOASSERT`.

From a different perspective, if we want to make the Gradualizer source code itself be an example of how to write as safe as possible code with Gradualizer, then I think this is what we would write to stay on the safe/conservative side. This does not mean everyone adopting Gradualizer has to apply the idiom.

All in all, though, without at least `assert_type` an `integer()` is not an `arity()`, so we need something to convince the type checker.

erszcz:
> It's a bit hard to review though.

Yeah, I'll clean up the code and git log, too. I just wanted to share the result just as I finally got it to work as expected.

src/typechecker.erl
zuiderkwast: Yeah I know we need `?assert_type` to silence the type error, which we already had in place, but we talked about "inserting runtime casts" as the papers mentioned, which would make the runtime type error happen where the assert is and not somewhere else later on.

With an advanced macro system (Elixir? or code generation?) we could make `?assert_type(Expr, arity())` expand to sth `is_integer(Expr) andalso Expr >= 0 andalso Expr < 256`, but it would require smartness beyond Erlang's macro system and it might not work for all types.

zuiderkwast: #446

erszcz:
@josefs @zuiderkwast 
Ok, I think this is finally ready for review:
  - the order of spec/function clauses doesn't matter, they can be mixed
  - (non)emptiness of lists can be preserved given an appropriate spec

    ```erlang
    -spec map(fun((A) -> B), [A, ...]) -> [B, ...];
             (fun((A) -> B), [A]) -> [B].
    ````

  - all the tests (including a few new ones) pass
  - self reported errors are at the lowest ever: 38 lines
  - the code is in a decent shape

The constraints aren't properly preserved yet. I think we can defer that to the work on the constraint solver, though.

src/typechecker.erl
erszcz: A few more comments on this:

> we talked about "inserting runtime casts"

I think that if/when we get to actually inserting such casts, we'll have the freedom to drop such explicit checks as the one we're commenting. For now, I think this is the safest we can do, yet still reasonably lightweight.

> With an advanced macro system [...]

Indeed, or if/when running type checking as a stage of a compiler.

zuiderkwast (approved):
LGTM!

I haven't understood all the details but I assume it's according to what we've discussed, so I'm fine with merging this. Yeyyyy!

PR #470, opened 2022-09-30
support excluding modules in gradualizer:type_check_file

tsloughter:
Not sure this is the best place to do it but I think it is the only place to do the filtering that doesn't require doing it in multiple places.

### DIFF
--- a/src/gradualizer.erl
+++ b/src/gradualizer.erl
@@ -58,24 +58,31 @@ type_check_file(File) ->
 %% @doc Type check a source or beam file
 -spec type_check_file(file:filename(), options()) -> ok | nok | [{file:filename(), any()}].
 type_check_file(File, Opts) ->
-    case filename:extension(File) of
-        ".erl" ->
-            Includes = proplists:get_all_values(i, Opts),
-            case gradualizer_file_utils:get_forms_from_erl(File, Includes) of
-                {ok, Forms} ->
-                    lint_and_check_forms(Forms, File, Opts);
-                Error ->
-                    throw(Error)
-            end;
-        ".beam" ->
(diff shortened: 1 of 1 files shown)


src/gradualizer.erl
```diff
@@ -58,24 +58,31 @@ type_check_file(File) ->
...
-            case gradualizer_file_utils:get_forms_from_beam(File) of
-                {ok, Forms} ->
-                    type_check_forms(File, Forms, Opts);
-                Error ->
-                    throw(Error)
-            end;
-        Ext ->
-            throw({unknown_file_extension, Ext})
+    ExcludeModules = proplists:get_value(exclude_modules, Opts, []),
+    AtomBaseFile = list_to_atom(filename:rootname(filename:basename(File))),
```
zuiderkwast, line 62: Perhaps we can avoid creating the atom until we know the extension is ".beam" or ".erl"?

zuiderkwast (commented):
Welcome new contributor Tristan!

I think it looks like the right place to do it.

Next steps:

* Make the rebar3 plugin use this instead of its own logic.
* Docs (edoc somewhere, not dure where we list options)
* Add corresponding CLI flags

src/gradualizer.erl
tsloughter: Is that a common case? That'd only happen if a user passed in a filename to something that isn't a beam or erl file and not a big deal in that case to create the atom before it throws the exception.

zuiderkwast: You're right, sorry.

erszcz (approved):
Looks good! This seems to be the best place to do the filtering.

PR #482, opened 2022-10-21
fix GitHub reference in rebar3 example README

kivra-fabber:
The git:// protocol is no longer supported, see:
https://github.blog/2021-09-01-improving-git-protocol-security-github/


erszcz:
Awesome, thanks for spotting this! 🚀

kivra-fabber:
🚀  have a good weekend!

PR #485, opened 2022-11-17
Expand erlang.++ function specs

luk-pau-es:
Expand function specs for erlangs ++ operator.

### DIFF
--- a/test/should_pass/list_concat_op_pass.erl
+++ b/test/should_pass/list_concat_op_pass.erl
@@ -5,7 +5,7 @@
          nonempty2_concat_fun_elem_gives_improper/0,
          nil_concat_fun_nonempty_gives_proper/0,
          nonempty_concat_fun_nonempty_gives_proper/0,
-         nonempty1_concat_fun_nonempty2_gives_proper/0]).
+         proper_concat_fun_proper_gives_proper/0]).
 
 -spec nil_concat_fun_elem_gives_elem() -> b.
 nil_concat_fun_elem_gives_elem() ->
@@ -27,8 +27,8 @@ nil_concat_fun_nonempty_gives_proper() ->
 nonempty_concat_fun_nonempty_gives_proper() ->
     erlang:'++'([a,b], [c]).
 
--spec nonempty1_concat_fun_nonempty2_gives_proper() -> [integer()].
-nonempty1_concat_fun_nonempty2_gives_proper() ->
+-spec proper_concat_fun_proper_gives_proper() -> [integer()].
+proper_concat_fun_proper_gives_proper() ->
     erlang:'++'(generate_list(), generate_list()).
 
 -spec generate_list() -> [integer()].
--- a/test/should_pass/list_concat_op_pass.erl
+++ b/test/should_pass/list_concat_op_pass.erl
@@ -4,7 +4,8 @@
          nonempty1_concat_fun_elem_gives_improper/0,
          nonempty2_concat_fun_elem_gives_improper/0,
          nil_concat_fun_nonempty_gives_proper/0,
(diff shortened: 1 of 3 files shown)


erszcz:
@luk-pau-es awesome! Could you rebase this on top of https://github.com/erszcz/Gradualizer/tree/split-list-concat-op-tests and add a test to `test/should_pass/list_concat_op_pass.erl` which exercises the new spec? I.e. a test which fails with the current spec, but works with the new one.

erszcz:
@zuiderkwast
> Wow! Lots of known problems fixed. This looks awesome.

It looks impressive, but it's actually code already present in master. Most of the tests in `test/should_pass/list_concat_op_pass.erl` already passed, but were previously under `known_problems`.

@luk-pau-es
We have to add a spec to `generate_list()` - otherwise it doesn't exercise the new spec clauses. You can verify that by running the new test, but keeping the old spec - the test will not fail. A spec to add could be:

```
diff --git a/test/should_pass/list_concat_op_pass.erl b/test/should_pass/list_concat_op_pass.erl
index 8af53a9..8b8c3db 100644
--- a/test/should_pass/list_concat_op_pass.erl
+++ b/test/should_pass/list_concat_op_pass.erl
@@ -31,4 +31,5 @@ nonempty_concat_fun_nonempty_gives_proper() ->
 nonempty1_concat_fun_nonempty2_gives_proper() ->
     erlang:'++'(generate_list(), generate_list()).
 
+-spec generate_list() -> [integer()].
 generate_list() -> lists:seq(1, 2).
```

zuiderkwast:
> Most of the tests in `test/should_pass/list_concat_op_pass.erl` already passed, but were previously under `known_problems`.

Oh, that's a problem in our test framework in that case. I thought a passing test in known_problems/should_pass would fail the test run. That's what I'd expect anyway.

erszcz:
@zuiderkwast It seems that if any test in a `known_problems/should_pass` file fails, the entire file is treated as a known problem, so some passing tests might hide in such a file. If no test fails in a `known_problems/should_pass` file, then it's reported as solved.

test/should_pass/list_concat_op_pass.erl
```diff
@@ -0,0 +1,35 @@
...
+-spec nil_concat_fun_nonempty_gives_proper() -> [atom()].
+nil_concat_fun_nonempty_gives_proper() ->
+    erlang:'++'([], [a]).
+
+-spec nonempty_concat_fun_nonempty_gives_proper() -> [atom()].
+nonempty_concat_fun_nonempty_gives_proper() ->
+    erlang:'++'([a,b], [c]).
+
+-spec nonempty1_concat_fun_nonempty2_gives_proper() -> [integer()].
+nonempty1_concat_fun_nonempty2_gives_proper() ->
```
erszcz, line 31:
```suggestion
-spec proper_concat_fun_proper_gives_proper() -> [integer()].
proper_concat_fun_proper_gives_proper() ->
```

test/should_pass/list_concat_op_pass.erl
```diff
@@ -0,0 +1,35 @@
+-module(list_concat_op_pass).
+
+-export([nil_concat_fun_elem_gives_elem/0,
+         nonempty1_concat_fun_elem_gives_improper/0,
+         nonempty2_concat_fun_elem_gives_improper/0,
+         nil_concat_fun_nonempty_gives_proper/0,
+         nonempty_concat_fun_nonempty_gives_proper/0,
+         nonempty1_concat_fun_nonempty2_gives_proper/0]).
```
erszcz, line 8:
```suggestion
         proper_concat_fun_proper_gives_proper/0]).
```

PR #487, opened 2022-11-23
Override filename specs

erszcz:
This uses function intersection type support to define better specs for some functions from the `filename` module. It therefore reduces the number of self-check errors and also allows to drop some assertions.

The type names in `file` and `filename` are quite hard to memorise: `file:name()` vs `file:filename()` vs `file:filename_all()` vs `file:name_all()`. I'm not using them in the new prelude file and strive for using the builtin type names instead.

### DIFF
--- a/priv/prelude/filename.specs.erl
+++ b/priv/prelude/filename.specs.erl
@@ -5,6 +5,9 @@
 -spec basename(binary()) -> binary();
               (string() | atom() | deep_list()) -> string().
 
+-spec dirname(binary()) -> binary();
+             (string() | atom() | deep_list()) -> string().
+
 -spec rootname(binary()) -> binary();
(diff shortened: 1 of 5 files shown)


zuiderkwast (approved):
LGTM.

I wonder if it's worth trying to upstream these specs to OTP...

erszcz:
> I wonder if it's worth trying to upstream these specs to OTP...

My concern is the possible impact on other tools and therefore pushback from OTP maintainers. I think Dialyzer doesn't make use of the extra precision achieved thanks to multiple spec clauses, but I hope it doesn't complain either.

KennethL:
I think we should have a dialogue about this. The important thing is that the specs in OTP are correct and as good as possible and still maintaining compatibility with all type checking tools. I think special notation to please a particular tool should be avoided.

PR #489, opened 2022-11-27
Add known problem for unsupported maybe expression

ferd:
The syntax for the new experimental maybe expression is currently not supported. Highlight and document this fact by putting it in the known problems test directory.

Avoid compiling with support for the maybe feature outside of test profiles to avoid causing issues in people depending on gradualizer trying to load modules.

I have no idea how to fix this but felt there should be a moderately helpful bug report by at least putting in a test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,8 @@
 .PHONY: all
 all: app escript
 
+ERL_OPTS = -enable-feature maybe_expr
+
 # Compilation
 
 erls = $(wildcard src/*.erl)
@@ -143,7 +145,7 @@ end
 endef
 
 eunit: compile-tests
-	erl -noinput -pa ebin -pa test -eval \
+	erl $(ERL_OPTS) -noinput -pa ebin -pa test -eval \
 	 '$(erl_run_eunit), halt().'
 
 cli-tests: bin/gradualizer test/arg.beam
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1903,7 +1903,12 @@ do_type_check_expr(Env, {'try', _, Block, CaseCs, CatchCs, AfterBlock}) ->
           end,
     {normalize({type, erl_anno:new(0), union, [Ty, TyC, TyS]}, Env)
     ,VB
-    ,constraints:combine([Cs1,Cs2,Cs3,Cs4])}.
+    ,constraints:combine([Cs1,Cs2,Cs3,Cs4])};
+
+%% Maybe - value-based error handling expression
+%% See https://www.erlang.org/eeps/eep-0049
+do_type_check_expr(Env, {'maybe', Anno, [{maybe_match, _, _LHS, _RHS}]} = MaybeExpr) ->
+    erlang:throw({unsupported_expression, Anno, MaybeExpr}).
 
 %% Helper for type_check_expr for funs
 -spec type_check_fun(env(), _) -> {type(), env(), constraints:constraints()}.
--- a/test/test.erl
+++ b/test/test.erl
@@ -57,7 +57,10 @@ gen_should_fail() ->
 gen_known_problem_should_pass() ->
     map_erl_files(
       fun(File) ->
-              ?_assertNotEqual(ok, safe_type_check_file(File))
+              {ok, Forms} = gradualizer_file_utils:get_forms_from_erl(File, []),
+              ExpectedErrors = typechecker:number_of_exported_functions(Forms),
+              ReturnedErrors = length(safe_type_check_file(File, [return_errors])),
+              ?_assertEqual(ExpectedErrors, ReturnedErrors)
       end, "test/known_problems/should_pass").
 
 % Test succeeds if Gradualizer crashes or if it does type check.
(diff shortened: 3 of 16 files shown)


erszcz:
@zuiderkwast WDYT?

Makefile
```diff
@@ -37,8 +37,20 @@ all: app escript
...
+io:format("~s\n", [erlang:system_info(otp_release)]), halt().
+endef
+OTP_VERSION := $(shell erl -noinput -eval '$(otp_version)')
+
+ERL_OPTS 		=
+ERLC_OPTS 		= -I include -I src -pa ebin +debug_info
+TEST_ERLC_OPTS 	= +debug_info
+
+ifeq ($(shell [ $(OTP_VERSION) -ge 25 ] && echo y), y)
+ERL_OPTS 		+= -enable-feature maybe_expr
```
zuiderkwast, line 50: This looks like Erlang.mk magic. :grin: I prefer simplicity...

Instead of passing this flag to `erlc`, I believe we can enable it using `-feature(maybe_expr).` in the erl file itself.

We can pass `-enable-feature maybe_expr` to `erl`. It seems to ignore this on older versions.

Then we don't need this makefile magic. Would it work?

zuiderkwast (commented):
I like the test case. Good to have!

But I don't like the Makefile tricks that much. I hope we can avoid it.

ferd:
Yeah that all works for me. Also gives room to eventually support more complex maybe expressions down the road. FWIW I tried to cover as many interleavings as significant in [the OTP Test suite](https://github.com/erlang/otp/blob/a8ea725d017317ac3835eda45797b80c6d263405/lib/compiler/test/maybe_SUITE.erl) (which I believe the OTP team expanded since last year's SpawnFest submission).

Makefile
erszcz: > I believe we can enable it using `-feature(maybe_expr, enable)`. in the `.erl` file itself.

Indeed, it can be skipped in the Makefile. We can't skip passing the flag to `erlc`, though.

> Then we don't need this makefile magic. Would it work?

We still need it to control that pre-25 we don't want to compile the test file in the extra path: `test/otp25/known_problems/should_pass/maybe_expr.erl`. Otherwise, because there's only one test in the `maybe_expr.erl` file that's feature gated, on older versions checking the file passes, therefore it's raised that the file shouldn't be a known problem. See CI logs: https://github.com/josefs/Gradualizer/actions/runs/3576727016/jobs/6014886961

We either need the condition on OTP version in the Makefile or we need to inject an `else` preprocessor clause and a bogus (failing) test to `maybe_expr.erl` itself, so it stays a known problem no matter which branch the preprocessor takes.

zuiderkwast: For the `should_fail` tests, I think the test framework needs to be changed to detect passing functions within a file of known problems. I think it should check that there is one error for each exported function, so there's 0 errors for 0 functions, that's "all functions fail".

src/typechecker.erl
```diff
@@ -1903,7 +1903,12 @@ do_type_check_expr(Env, {'try', _, Block, CaseCs, CatchCs, AfterBlock}) ->
           end,
     {normalize({type, erl_anno:new(0), union, [Ty, TyC, TyS]}, Env)
     ,VB
-    ,constraints:combine([Cs1,Cs2,Cs3,Cs4])}.
+    ,constraints:combine([Cs1,Cs2,Cs3,Cs4])};
+
+%% Maybe - value-based error handling expression
+%% See https://www.erlang.org/eeps/eep-0049
+do_type_check_expr(Env, {'maybe', Anno, [{maybe_match, _, _LHS, _RHS}]} = MaybeExpr) ->
+    erlang:throw({unsupported_expression, Anno, MaybeExpr}).
```
zuiderkwast, line 1911: Why not make this a catch-all? There may be more unsupported expressions in the future.

test/test.erl
```diff
@@ -57,7 +57,10 @@ gen_should_fail() ->
 gen_known_problem_should_pass() ->
     map_erl_files(
       fun(File) ->
-              ?_assertNotEqual(ok, safe_type_check_file(File))
+              {ok, Forms} = gradualizer_file_utils:get_forms_from_erl(File, []),
+              ExpectedErrors = typechecker:number_of_exported_functions(Forms),
+              ReturnedErrors = length(safe_type_check_file(File, [return_errors])),
+              ?_assertEqual(ExpectedErrors, ReturnedErrors)
```
zuiderkwast, line 63: Good!

zuiderkwast (commented):
Awesome. I was just thinking about a few more things.

We handle `maybe` (unsupported expression clause) in `do_type_check_expr`. Shouldn't we also have such clause in `do_type_check_expr_in`? I believer we can trigger this by adding the same maybe-test again but with a spec.

Should we enable the maybe expression in the escript too? We generate the emulator options line in the Makefile like `%%! -escript main gradualizer_cli`.

erszcz:
Yeah, good points, @zuiderkwast 👍 Since I already merged this one, let's have a ticket for the future not to forget about these things - #494.

PR #491, opened 2022-11-28
Print inferred argument types on call_intersect error

### LINKED ISSUE #484: Error message needs more info: `The type of the function _____, called on line ___ doesn't match the surrounding calling context.`
This particular error message is insufficiently precise, which makes it hard to debug. It shows the type of the function that is trying to be called, but not what the args to the function are that causes this check to fail.

Example:

```
lib/gradient.ex: The type of the function :erlang.++(), called on line 194 doesn't match the surrounding calling context.
It has the following type
(nonempty_list(T1), list(T2) -> nonempty_list(T1 | T2))
(nonempty_list(T1), nonempty_improper_list(T2, T3) -> nonempty_improper_list(T1 | T2, T3))
(nonempty_list(T1), T2 -> nonempty_improper_list(T1, T2))
([], T -> T)
```

(This comes from `gradualizer_fmt:format_type_error/2`, the clause that looks like this: `format_type_error({type_error, call_intersect, Anno, FunTy, Name}, Opts)`).

Here is a better error message:

```
The variable on line 113 is expected to have type binary() but it has type integer() | float()
```

(This comes from `gradualizer_fmt:format_expr_type_error/4`.) Note that it says what the type actually is, `integer() | float()`, along with the expected type `binary()`, so it's much easier to see what the problem is.

**How to reproduce?**

There is probably a better way to do this. I'm not familiar enough with Gradualizer to provide an isolated test case. However, I've been running into this on my `gradient` PR here: https://github.com/esl/gradient/pull/124 . Pull the branch specified there, which is https://github.com/japhib/gradient and branch `umbrella-config` , and run `mix gradient` on the project itself.

erszcz:
Fixes #484.

The argument types are inferred, so they might not be completely accurate in each case. Example printout from Gradualizer's self-check:

```
ebin/typechecker.beam: type_error/5 call arguments on line 3345 at column 11 don't match the function type:
fun((call_arity, anno(), atom(), arity(), arity()) -> error())
fun((type_error(), binary_op(), anno(), type(), type()) ->
                  error())
Inferred argument types:
call_intersect, any(), any(), any(), [type()]
```

@japhib what do you think of the above message format?

### DIFF
--- a/src/gradualizer_fmt.erl
+++ b/src/gradualizer_fmt.erl
@@ -174,15 +174,6 @@ format_type_error({type_error, call_intersect, Anno, ArgsTys, FunTy, Name}, Opts
        length(ArgsTys),
        format_location(Anno, verbose, Opts),
        pp_intersection_type(FunTy, Opts)]);
-format_type_error({type_error, expected_fun_type, Anno, Func, FunTy}, Opts) ->
-    Name = pp_expr(Func, Opts),
-    io_lib:format(
-      "~sExpected function ~s~s to have a function type,~n"
-      "but it has the following type:~n~s~n",
-      [format_location(Anno, brief, Opts),
-       Name,
-       format_location(Anno, verbose, Opts),
-       pp_type(FunTy, Opts)]);
 format_type_error({type_error, no_type_match_intersection, Anno, Func, FunTy}, Opts) ->
     Name = pp_expr(Func, Opts),
     io_lib:format(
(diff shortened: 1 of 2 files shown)


erszcz:
@zuiderkwast WDYT? This PR is useful for debugging #512, so I got back to finishing it now.

japhib:
Looks great!

PR #493, opened 2022-12-02
Release 0.2.0

erszcz:
It has been a looong time since the previous release. I think it wouldn't hurt to tag one, add the changelog to the repo, push a new package to hex.pm, push ExDocs to hexdocs.pm, etc. If this PR is fine for you, I'll fast forward master to it, tag 0.2.0 and do the above things.


erszcz:
Apparently, 0.2.0 already got pushed to hex.pm because of the GH actions workflow. I'm dropping that workflow here, but I'm open to keeping it, too, if that's the preferred method of publishing packages.

.github/workflows/publish-to-hex.yml
```diff
@@ -1,17 +0,0 @@
-on:
-  push:
-    tags:
-      - '*'
```
zuiderkwast, line 4: This was added by @gilbertwong96 who wanted to publish it to hex once upon a time. Josef and me never bothered. @erszcz I think you can do this in your prefered way. (Personally I never remember how to publish stuff to hex.)

Perhaps we can just change this to take action only when a tag is created on master? If it works, it seems convenient...

zuiderkwast (approved):
LGTM

.github/workflows/publish-to-hex.yml
erszcz: The tool I found to generate the changelog requires a tag present in the remote repo. But this hook automatically triggers publishing a package when such a tag is pushed. That's why I think doing this manually is fine. It's a task to do once in a few months, I think, so not a lot of overhead.

zuiderkwast: Thx, very good.

gilbertwong96: > The tool I found to generate the changelog requires a tag present in the remote repo. But this hook automatically triggers publishing a package when such a tag is pushed. That's why I think doing this manually is fine. It's a task to do once in a few months, I think, so not a lot of overhead.

Which tool did you use to generate the changelog? What about generating a changelog automatically with the GitHub workflow action?

erszcz: @gilbertwong96

> Which tool did you use to generate the changelog?

[`github-changelog`](https://github.com/cfpb/github-changelog) as I wrote in the changelog itself. There was also a bit of manual editing involved.

In general, neither the PR titles, nor the commit messages in the repo are formatted consistently enough - it's rather a free-for-all. This makes generating the changelog completely automatically quite hard - I did try with [`git-chglog`](https://github.com/git-chglog/git-chglog) first, but the result was unsatisfactory.

Anyway, I don't think the overhead of doing this once in a couple of months is worth the effort of automation. Unless somebody wants to step forward with the initiative and propose a plan how to tackle it in the future - I'd be happy to see it happening :)

erszcz:
OK, all up:
- 0.2.0 tag and GH release - https://github.com/josefs/Gradualizer/releases/tag/0.2.0
- hex.pm package - https://hex.pm/packages/gradualizer
- hexdocs.pm - https://hexdocs.pm/gradualizer/readme.html

Issue #362, opened 2021-10-22
`type_check_expr` vs `type_check_expr_in` - unify or not?


erszcz:
These two seem to be similar, but according to some comments here in issues and in the code they are significantly different.

The source code comments say:

```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
%% @doc Infer type of expression
%%
%% Type information should* stem from user type annotation (function specs
%% and type annotated record definitions). If an expression does not have a
%% subexpression that has a type inferred from these sources, its inferred type
%% will be `any()'.
%%
%%   *) When the option 'infer' is true, the types of literals and other
%%      constructs are also propagated.
%%
%% Arguments: An environment for functions, an environment for variables
%% and the expression to type check.
%% Returns the type of the expression, a collection of variables bound in
%% the expression together with their type and constraints.
-spec type_check_expr(env(), expr()) -> {any(), venv(), constraints:constraints()}.
type_check_expr(Env, Expr) ->
```

and

```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
%% Checking the type of an expression
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

type_check_expr_in(Env, ResTy, Expr) ->
```

As we can see already their names are a bit unfortunate, as `_in`, at least to me, suggests _inference_, but it's exactly the other way around. Maybe it's for _internal_?

@zuiderkwast, I've found this comment - https://github.com/josefs/Gradualizer/discussions/337#discussioncomment-885390 - mentioning that it might be worth merging these functions. In such a case, would we still want the `infer` option to exist or would inference become the new default with no way to opt out of it?

michallepicki:
@erszcz 
> For the record, there was a bidirectional typing talk at this year's SIGPLAN Erlang Workshop - https://www.youtube.com/watch?v=MAerjPQUKQQ - but I cannot find the implementation, it might not be open-source 

The impl is at https://github.com/vrnithinkumar/ETC but I found it a bit hard to understand with non descriptive names and quite a lot of modules that take different approach, probably for benchmarks and other comparisons, and I couldn't get the tests to pass

erszcz:
I finally started reading _Bidirectional Typing_ by Jana Dunfield and Neel Krishnaswami and here's how the paper introduces the problem:

[...]

PR #496, opened 2022-12-06
Add support for module_info functions

### LINKED ISSUE #473: Undefined function `module_info/0`
I can try to dig into this at some point but thought someone would know how gets exported functions for a module? Is it for some reason only getting those exported in the `export` attribute of the module and not builtin ones like in this case:

```
otel_exporter.beam: Call to undefined function grpcbox:module_info/0 on line 73 at column 40
```

xxdavid:
I finally come up with a PR that fixes #473.

I am an Elixir person and haven't written almost any Erlang code before so there may be some things that could be done more idiomatically.

I think I got some (quite little still) idea how Gradualizer works but one thing still isn't clear to me – why isn't `type_check_expr_in` implemented in terms of `type_check_expr`. From my limited perspective, it could just determine the type of the expression by calling `type_check_expr_in` and then check whether the type is indeed a subtype of `ResTy`. Why wouldn't that work?

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -1658,6 +1658,10 @@ do_type_check_expr(Env, {call, _, {atom, _, TypeOp}, [Expr, {string, _, TypeStr}
 do_type_check_expr(Env, {call, _, {atom, _, record_info}, [_, _]} = Call) ->
     Ty = get_record_info_type(Call, Env),
     {Ty, Env, constraints:empty()};
+do_type_check_expr(Env, {call, P, {remote, _, _Mod, {atom, _, module_info}} = Name, Args}) ->
+    Arity = arity(length(Args)),
+    FunTy = get_module_info_type(Arity),
+    type_check_call_ty(Env, expect_fun_type(Env, FunTy, Arity), Args, {Name, P, FunTy});
 do_type_check_expr(Env, {call, P, Name, Args}) ->
     Arity = arity(length(Args)),
     {FunTy, VarBinds1, Cs1} = type_check_fun(Env, Name, Arity),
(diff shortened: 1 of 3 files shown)


erszcz:
Hi, @xxdavid! This is great!

I'll currently leave the review for a bit later, but I think I can share something on the question you ask:

> why isn't `type_check_expr_in` implemented in terms of `type_check_expr`

In a nutshell, type inference is a significantly harder problem than type checking. In general, it might be possible to type check a fully annotated program in a hypothetical language in a reasonable time, but not to rebuild the annotations if they're missing. Actually, full type inference of practical languages like Haskell or OCaml is undecidable - that's why in some cases the programmer has to provide type annotations to enable a program to compile.

It's not a new problem, though, which has focused interest on local methods originally known as local type inference and more recently as bidirectional typing. In practice, it means a type checker alternates between inferencing types where it can and only checking types where it cannot do any better (these are the places where we need annotations to be provided by a programmer). A great paper on that topic is [_Bidirectional Typing_ by Dunfield and Krishnaswami](https://dl.acm.org/doi/10.1145/3450952).

BTW, this question was previously raised, possibly even multiple times, and [possibly by me the last time](https://github.com/josefs/Gradualizer/issues/362) :D @josefs replied back then that:

> It would be great if it was possible to merge the two functions in a way that also made the code simpler. But I don't know how to do that.

Recently I also noticed some repetition between these functions, so maybe in practice there's room for improvement. In general, though, `type_check_expr_in` (type checking) is easier, whereas `type_check_expr` (synthesizing aka inferencing a type) might not be possible at all for some expressions. I haven't compared them in detail to tell if there are possible simplifications to be made.

xxdavid:
Aha! I haven't thought of that when raising the question. Thanks for your explanation! :slight_smile: And also for linking the previous issue you created that provides additional context. I'm glad I can associate the mysterious `_in` with a word (_inward_). :smile: I noticed that you sprinkled the Bidirectional typing paper also in our earlier discussion but I still haven't read it, I look forward forward to reading it after I grasp some of the more basic stuff.

erszcz:
This looks great, thanks @xxdavid! Congrats on the first contribution 🎉

erszcz:
@xxdavid Ahh, of course I missed something in the review 😅 Fixed in #498.

src/typechecker.erl
```diff
@@ -1658,6 +1658,10 @@ do_type_check_expr(Env, {call, _, {atom, _, TypeOp}, [Expr, {string, _, TypeStr}
 do_type_check_expr(Env, {call, _, {atom, _, record_info}, [_, _]} = Call) ->
     Ty = get_record_info_type(Call, Env),
     {Ty, Env, constraints:empty()};
+do_type_check_expr(Env, {call, P, {remote, _, _Mod, {atom, _, module_info}} = Name, Args}) ->
+    Arity = arity(length(Args)),
+    FunTy = get_module_info_type(Arity),
```
zuiderkwast, line 1663: I suppose we have a problem if Arity > 1. Perhaps we should break this clause up into explicit matching on arity 0 (`Args = []`) and 1 (`Args = [_]`) and fall though for any other arity?

xxdavid: Changed in #500.

PR #498, opened 2022-12-17
Fix specs introduced with Mod:module_info/1,2 support

erszcz:
This brings self-check warnings back to 39 lines.

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -5112,7 +5112,7 @@ record_field_types(Fields) ->
               end, Fields).
 
 % Helper returning the type of module_info/0 and module_info/1 functions
--spec get_module_info_type(arity()) -> fun_ty().
+-spec get_module_info_type(arity()) -> type() | [type()].
 get_module_info_type(0) ->
     Options = get_module_info_options(),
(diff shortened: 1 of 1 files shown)


src/typechecker.erl
```diff
@@ -5112,7 +5112,7 @@ record_field_types(Fields) ->
               end, Fields).
 
 % Helper returning the type of module_info/0 and module_info/1 functions
--spec get_module_info_type(arity()) -> fun_ty().
+-spec get_module_info_type(arity()) -> type() | [type()].
```
zuiderkwast, line 5115: Can we use an two-clause spec (0) -> type(); (1) -> [type()] here?

xxdavid:
Ooops, thanks for fixing it. I think the specs were right in an initial stage, but I then rewritten the implementation and forgot to change the specs. This is a good example of a situation where Gradualizer comes handy!

src/typechecker.erl
zuiderkwast: Btw, what happens if a code contains a call to `module_info/N` where N > 1? Do we crash?

PR #499, opened 2022-12-17
Require all functions in src/typechecker.erl to have specs

erszcz:
This significantly raises the bar on self-check consistency by requiring ALL functions in the type checker to have specs. It uncovered a number of consistency issues, but most of them could be solved. Most importantly, though, it documents the interfaces between various parts of the type checker and eases discoverability.

Done in this PR:
- [x] enabled `warn_missing_spec_all` in `src/typechecker.erl`
- [x] added all missing specs 
- [x] added some tuple union and record union tests
- [x] removed dead record handling code for which no examples could be found (cross-checked with Cover)
- [x] fixed Cover Make rule
- [x] renamed `constraints:constraints()` to `constraints:t()` for brevity
- [x] reduced the number of self-check error lines back to below 100

This PR is also a good example of dog-fooding.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,20 +169,22 @@ cli-tests: bin/gradualizer test/arg.beam
 	bin/gradualizer --no_prelude --specs_override_dir priv/prelude \
 	  test/should_pass/cyclic_otp_specs.erl || (echo "CLI 10"; exit 1)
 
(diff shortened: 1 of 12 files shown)


erszcz:
This fixes quite a few self-check errors, so I'm merging it, but I'm still happy with any reviews to come.

PR #504, opened 2023-01-18
Fix yet more self check errors

erszcz:
23 lines of self-check errors at this point. It has never been this low.


src/gradualizer_prelude_parse_trans.erl
```diff
@@ -35,4 +36,21 @@ get_module_and_forms(File) ->
...
     [{attribute, _, file, _},
      {attribute, _, module, Module} | _] = Forms,
-    {Module, Forms}.
+    {Module, [ erl_anno_mark(F) || F <- Forms ]}.
+
+erl_anno_mark({eof, _} = Form) ->
+    Form;
+erl_anno_mark(Form) ->
+    Anno = element(2, Form),
+    setelement(2, Form, {'$anno', Anno}).
```
zuiderkwast, line 45: What's the point of this? I don't understand.

src/gradualizer_file_utils.erl
```diff
@@ -58,19 +58,12 @@ epp_parse_file(File, Includes) ->
...
-        true ->
-            epp:open(File, Fd, StartLocation, Includes, []);
-        false ->
-            epp:open([{name, File},
-                      {location, StartLocation},
-                      {includes, Includes},
-                      {fd, Fd}])
-    end.
+    epp:open([{name, File},
+              {location, StartLocation},
```
zuiderkwast, line 64: Apparently, this doesn't give us column numbers in OTP < 24.

Btw, code that depends on OTP version by `erlang:function_exported/3` is an interesting and valid use case. It's the dynamic nature of dynamic loading. If we work around it here, perhaps we can add it to known problems instead?

It can't really be solved, but we can come up with some recommended way to silence it.

zuiderkwast (commented):
Nice!

src/gradualizer_file_utils.erl
erszcz: It can and I solved it in another way, but then decided it's better to remove code than to add more. I'll push the alternative solution instead, since you mention this.

BTW, I don't know why, but dropping this leads to the CI failure on 23.3.4.7 - is that connected with the missing column numbers? The other solution I have in mind passed CI.

src/gradualizer_prelude_parse_trans.erl
erszcz: The point is that the annotation is just an integer or a pair of integers. After transforming into the `erl_syntax` representation it would be very hard to tell in `erl_syntax_lib:map()` mapper which integer (or which pair) that we see in the tree is an annotation and which is not. By leaving a marker like this one, we can then quite safely match on a tuple like `{tree, tuple, _, [{atom, _, '$anno'}, Anno]}` and finally substitute it with what we ultimately need here - an `erl_anno:new()` call. Why do we need that call? See the commit message on https://github.com/josefs/Gradualizer/pull/504/commits/9a5bd852f060f706f42d1e01a14a0a18bc48af2e.

src/gradualizer_file_utils.erl
zuiderkwast: Yes, I believe it has to do with column numbers. Some OTPs ago, there were no column numbers and then it was added gradually in different modules. (Our Peter Gömöri was involved in adding column numbers in OTP.)

erszcz: I've pushed the alternative approach. It's based on `--specs_override_dir` and providing specs of all the functions we expect to use in various OTP versions. WDYT?

erszcz:
20 lines as of now and some new known problems.

PR #506, opened 2023-01-24
Drop invalid lists:filtermap/2 known problem

erszcz:
The original case this test was based on is not really a problem with the spec or the constraint solver - it was an actual bug in how we use the function. It was originally merged in #504.


erszcz:
I've also cross checked with the Haskell standard library and see there isn't a combined `filtermap` there. I think that without union types it would be pretty hard to express the return value in the Haskell type system. However, this also hints that the union type is correct there and that the actual use case of `filtermap` was buggy, not it's type checking.

zuiderkwast (approved):
I'm fine with deleting this. The use case is not valid. It needs some type assertion or guard. The spec in OTP is correct, i.e.

```Erlang
  -spec filtermap(Fun, List1) -> List2
                     when
                         Fun :: fun((Elem) -> boolean() | {true, Value}),
                         List1 :: [Elem],
                         List2 :: [Elem | Value],
                         Elem :: term(),
                         Value :: term().
```

PR #516, opened 2023-03-03
Test fixes for #500

erszcz:


src/gradualizer.erl
```diff
@@ -171,13 +171,15 @@ type_check_file(File, Opts) ->
                     Includes = proplists:get_all_values(i, Opts),
                     case gradualizer_file_utils:get_forms_from_erl(File, Includes) of
                         {ok, Forms} ->
+                            ok = gradualizer_db:import_erl_files([File]),
                             lint_and_check_forms(Forms, File, Opts);
```
zuiderkwast, line 175: Can this line crash for bad erl files? We want to handle that gracefully, don't we?

We can do the lint step before import, if it helps.

WDYT?

Issue #501, opened 2023-01-05
No errors are flagged for guarded function clauses


duncanatt:
I am experimenting with Gradualizer on the code below, expecting it to complain.

```erlang
-spec inc_1(atom()) -> integer().
inc_1(X) when is_integer(X) -> X + 1.
```

However, no error is flagged. Similarly, no error is raised for this function:

```erlang
-spec inc_2(integer()) -> integer().
inc_2(X) when is_atom(X) -> X + 1.
```

Once the guards are removed from the above functions, the expected behaviour is obtained (*i.e.* an error is flagged in the case of `inc_1`, whereas `inc_2` type checks). Is there something that I am missing, please?

erszcz:
Hi, @duncanatt!

Thanks for raising this.

It's more complex than that, but a short answer is that Gradualizer is still experimental and just not complete. This means that there's some refinement happening, but not as much as there could be, and apparently obvious refinement failures are not signalled either.

Generally speaking refinement based on patterns (also known as exhaustiveness checking) is probably somewhat more advanced, but there's a little bit of refinement based on guard BIFs in place, too. The best, though not necessarily the easiest way to know what's already supported, is checking the tests. We have the following which somehow refer to type *refin*ement:

```
19:19:14 erszcz @ x6 : ~/work/erszcz/gradualizer (master %)
$ fd refin test
test/known_problems/should_fail/refine_ty_vars.erl
test/known_problems/should_pass/refine_bound_var_on_mismatch.erl
test/known_problems/should_pass/refine_list_tail.erl
test/known_problems/should_pass/refine_mismatch_using_guard_bifs.erl
test/should_fail/exhaustive_refinable_map_variants.erl
test/should_fail/record_refinement_fail.erl
test/should_fail/tuple_union_refinement.erl
test/should_fail/type_refinement_fail.erl
test/should_fail/unreachable_after_refinement.erl
test/should_pass/record_refinement.erl
test/should_pass/refine_comparison.erl
test/should_pass/type_refinement_pass.erl
```

Of these, `test/should_pass/type_refinement_pass.erl` has some tests for type refinement based on guards:

```erlang
-spec refine_bound_var_by_guard_bifs(ok | integer() | pid() |
                                     #{x => y} | {x, y}) -> ok.
refine_bound_var_by_guard_bifs(X) ->
    if
        is_map(X)     -> ok;
        is_tuple(X)   -> ok;
        is_integer(X) -> ok;
[...]

PR #520, opened 2023-03-07
New known problem: refine bound vars with guards

erszcz:
This defines a known problem for one of the type check errors we discussed in https://github.com/josefs/Gradualizer/pull/512#discussion_r1122806508.

In general, due to #512 there are many more union types being passed around in code, which leads to more places where we have to manually assert which union member we're dealing with. I think using guards more extensively would:
-  allow for avoiding mundane refactorings such as some of those done in #512, which are currently necessary to satisfy the type checker, but lead to clunkier and less natural Erlang,
- play well with the idea of gradual typing; guards carry extra information that could help our static analysis, but currently this information is most often discarded; #501 touches on how we could use guards to provide more feedback, too.

This PR defines just one known problem because so far I only managed to isolate one from #512 changes, but I remember there were more cases like this. Not all are equally easy to reproduce, i.e. sometimes it's not obvious why Gradualizer can't tell which union member it's dealing with. This is connected with how we represent `gradualizer_type:abstract_type()`.

### DIFF
--- a/test/known_problems/should_pass/refine_bound_var_with_guard_should_pass.erl
+++ b/test/known_problems/should_pass/refine_bound_var_with_guard_should_pass.erl
@@ -1,0 +1,37 @@
+-module(refine_bound_var_with_guard_should_pass).
+
+-export([f/1]).
+
+%% This type is simpler than gradualizer_type:abstract_type() by having less variants
+%% and by using tuples to contain deeper nodes. The latter frees us from having to deal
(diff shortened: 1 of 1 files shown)


erszcz:
I see no veto, so I'm merging this. Let me know, please, if you have any comments.

zuiderkwast (commented):
LGTM

I don't always check messages during the evening/night. :D

PR #524, opened 2023-03-29
Improve map exhaustiveness checking

xxdavid:
As I've promised in #513, I've looked at exhaustiveness checking for maps, and spent the past several weeks trying to figure out how the algorithms should work. This is like my third iteration of the implementation, the previous ones that I wrote were buggy, so there may be bugs even in this one (but I hope not).

I've tried to test it thoroughly and everything seems to work (except for two edge cases described in the commits). I've also tried to make booleans refinable (as in #513) and everything passes as well.

@erszcz @zuiderkwast What do you think?

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -18,6 +18,7 @@
          type/1, type/2,
          type_diff/3,
          refinable/2,
+         add_types_pats/4,
          compatible/3,
          collect_specs_types_opaques_and_functions/1,
          number_of_exported_functions/1,
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -111,6 +111,7 @@
 -type record_field() :: gradualizer_type:af_record_field(expr()).
 -type record_field_type() :: gradualizer_type:af_record_field_type().
 -type typed_record_field() :: {typed_record_field, record_field(), type()}.
+-type af_assoc_type() :: gradualizer_type:af_assoc_type().
 
 %% The environment passed around during typechecking.
 %% TODO: See https://github.com/josefs/Gradualizer/issues/364 for details.
@@ -4892,29 +4893,36 @@ add_type_pat({map, P, AssocPats} = MapPat, MapTy, Env) ->
             AssocTys = ?assert_type(AssocTys, [gradualizer_type:af_assoc_type()]),
             %% Check each Key := Value and bind vars in Value.
             CheckAssoc = fun ({map_field_exact, _, Key, ValuePat},
-                              {EnvIn, ExhaustedAssocsAcc, CsAcc}) ->
-                                 case add_type_pat_map_key(Key, AssocTys, EnvIn) of
-                                     {ok, KeyPatTy, ValueTy, Cs1} ->
+                              {EnvIn, ExhaustedAssocsAcc, RemainingAssocsIn, CsAcc}) ->
+                                 case add_type_pat_map_key(Key, RemainingAssocsIn, EnvIn, []) of
+                                     {ok, KeyPatTy, ValueTy, RemainingAssocsOut, Cs1} ->
                                          {ValPatTy, _ValUBound, EnvOut, Cs2} =
                                              add_type_pat(ValuePat, normalize(ValueTy, EnvIn), EnvIn),
-                                         ExhaustedAssoc = type(map_field_exact,
+                                         ExhaustedAssoc = type_assoc(map_field_exact,
                                                                [KeyPatTy, ValPatTy]),
                                          {EnvOut,
                                           [ExhaustedAssoc | ExhaustedAssocsAcc],
+                                          RemainingAssocsOut,
                                           [Cs1, Cs2 | CsAcc]};
                                      error ->
+                                         %% TODO: allow patterns like #{x := a, x := a}
                                          throw(type_error(badkey, Key, MapTy))
                                  end
                          end,
-            {NewEnv, ExhaustedAssocs, Css} =
-                lists:foldl(CheckAssoc, {Env, [], [Cs0]}, AssocPats),
+            {NewEnv, ExhaustedAssocsRev, RemainingAssocs, Css} =
+                lists:foldl(CheckAssoc, {Env, [], AssocTys, [Cs0]}, AssocPats),
             PatTy = case NormMapTy of
                         ?top() ->
                             top();
                         {var, _, _Var} ->
                             type(none);
                         ?type(map, Assocs) when is_list(Assocs) ->
-                            PatTy0 = type(map, ExhaustedAssocs),
+                            ExhaustedAssocs = lists:reverse(ExhaustedAssocsRev),
+                            AllExhaustedAssocs = case RemainingAssocs of
+                                any -> ExhaustedAssocs;
+                                _List -> ExhaustedAssocs ++ RemainingAssocs
+                            end,
+                            PatTy0 = type(map, AllExhaustedAssocs),
                             handle_possible_none_map_keys(PatTy0)
                     end,
             {PatTy, MapTy, NewEnv, constraints:combine(Css)};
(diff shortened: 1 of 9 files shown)


test/known_problems/should_pass/inner_union_subtype_of_root_union.erl
```diff
@@ -0,0 +1,18 @@
+-module(inner_union_subtype_of_root_union).
+
+-export([f/0]).
+
+% The problem is that {t, a} | {t, b} is not subtype of {t, a|b}.
```
erszcz, line 5: We could add somewhere here that there's a similar case for map values: https://github.com/josefs/Gradualizer/blob/master/test/known_problems/should_pass/map_union.erl

test/typechecker_tests.erl
```diff
@@ -592,15 +595,184 @@ add_type_pat_test_() ->
...
+                    typechecker:type_diff(?t(#{x => a|b, y := c}),
+                                          ?t(#{x := a, y := c}),
+                                          gradualizer:env())),
+
+      ?_assertEqual(?t(#{y => b}),
+                    typechecker:type_diff(?t(#{x => a, y => b}),
+                                          ?t(#{x := a, y => b}),
+                                          gradualizer:env())),
+
+      ?_assertEqual(?t(#{y => a}),
```
erszcz, line 744: Shouldn't it be this? Similar to the case just below and also to `type_diff(?t(#{x => integer()}), ?t(#{x => integer()}))`?
```suggestion
      ?_assertEqual(?t(none()),
```

erszcz (approved):
This is terrific work, @xxdavid! 🚀

src/typechecker.erl
```diff
@@ -1076,6 +1091,15 @@ merge_union_types(Types, _Env) ->
...
 is_top(_) -> false.
 
+%% Decides whether the type expresses exactly one value.
+-spec is_singleton_type(type()) -> boolean().
+is_singleton_type({atom, _, _Atom}) -> true;
+is_singleton_type({integer, _, _Integer}) -> true;
+is_singleton_type({char, _, _Char}) -> true;
+is_singleton_type(?type(range, [{integer, _, N}, {integer, _, N}])) -> true;
+is_singleton_type(?type(map, [])) -> true;
+is_singleton_type(_) -> false.
```
zuiderkwast, line 1101: The empty list (nil) is also a singleton type. The empty binary too.

test/known_problems/should_pass/inner_union_subtype_of_root_union.erl
```diff
@@ -0,0 +1,18 @@
...
+
+% It has surfaced because
+%     {'type', anno(), 'map_field_assoc' | 'map_field_exact', [abstract_type()]}
+% is not a subtype of
+%     {'type', anno(), 'map_field_assoc', [abstract_type()]}
+%   | {'type', anno(), 'map_field_exact', [abstract_type()]}
```
zuiderkwast, line 11: Could we try to solve this by `normalize`ing them to the same structure?

zuiderkwast (commented):
This is not a full review. I've just skimmed through it. It seems to solve a lot of things. :+1:

xxdavid:
Thank you for your reviews! 🙂

test/typechecker_tests.erl
erszcz: I'm convinced 👍

test/known_problems/should_pass/inner_union_subtype_of_root_union.erl
xxdavid: Aha! I'll merge them into a single file.

xxdavid, line 5: I'm wrong here. It's the other way around. `{t, a|b}` is (in our implementation) not subtype of `{t, a} | {t, b}`. I'll fix this.

test/known_problems/should_pass/inner_union_subtype_of_root_union.erl
xxdavid: Hmm, maybe we could. But I'm not sure what should be criterion/algorithm for merging. We probably agree that `{t, a} | {t, b}` should be normalized into `{t, a|b}`. But what about `{t, a, c} | {t, b, d}`, should it be merged into `{t, a|b, c|d}`? Probably not because this way, we would lose information. So what would be the criterion? Whether all the elements of the tuples but one are the same?

src/typechecker.erl
xxdavid: You're right! I'll add the case for `nil`. What's the empty binary type? `{type,Ann,binary,[{integer,0,0},{integer,0,0}]}`?

test/known_problems/should_pass/inner_union_subtype_of_root_union.erl
zuiderkwast: Right, it's not trivial in the generic case.

Maybe tuples that differ in only one element can be a good candidate, but it may be hard to detect them. We need some algorithm to group them. Maybe something similar for unions of records and maps where only one element differs. Let's leave it for another PR.

xxdavid: I agree. 🙂

test/known_problems/should_pass/inner_union_subtype_of_root_union.erl
xxdavid: Done.

xxdavid: Fixed.

src/typechecker.erl
erszcz: Exactly:

```
$ nl -b a z.erl
     1	-module(z).
     2
     3	-type z() :: <<>>.
$ erlc +dabstr z.erl
z.erl:3:2: Warning: type z() is unused
%    3| -type z() :: <<>>.
%     |  ^

$ cat z.abstr
{attribute,{1,1},file,{"z.erl",1}}.
{attribute,{1,2},module,z}.
{attribute,{3,2},
           type,
           {z,{type,{3,14},binary,[{integer,3,0},{integer,3,0}]},[]}}.
{eof,{4,1}}.
```

xxdavid: Great, I'll add that as well. :)

xxdavid: Done.

zuiderkwast: Wait, there's more:

The empty tuple and any tuple where all elements are singleton types. :) Don't know how far we should go, but I guess there are more of these, e.g. maps of only singletons, etc.

xxdavid: Good spot!

We probably could go to great lengths to define this function precisely but I think it's not needed now. At least until it's used somewhere else. IMHO, for the use in maps (to catch when a map contains non-singular required fields and disable exhaustiveness checking in that case), the first three clauses are enough (when considering practical use cases), with the first one being by far the most used.

Maybe we should also add the empty tuple type (when we already have the empty map type), but I wouldn't extend this function to be recursive until we really need it. What do you think?

zuiderkwast: I guess it's fine for now but in the long run it's be good if the typechecker is predictable, i.e. either does exhaustiveness checking whenever it's possible or it stops at some predictable depth or so.

xxdavid: Ok, I added the case for the empty tuple and a TODO note to add more cases.

erszcz:
Ok, it seems all comments are addressed, so I think it's ready to merge - I'm clicking the button.

test/known_problems/should_pass/inner_union_subtype_of_root_union.erl
erszcz: > We need some algorithm to group them.

I think we have to do the opposite, i.e. expand them. I've stumbled upon this in practice when dogfooding.

When we have a function with a lot of clauses, a clause that matches on a structure with an inner union needs an assertion from the programmer or an inner case expr to ensure the entire union is handled. This happens even if a previous clause handles a specific case of the structure with one of the union elements.

It's painful because it breaks exhaustiveness checking an requires an seemingly unnecessary annotation.

I think that if we expanded to a union of structures this would be refined and eliminated properly. However, it would come at a cost of having do deal with way bigger unions at the top level.

I'm writing from my phone, I'll follow up with specific examples from the codebase later.

zuiderkwast: Sounds good, but if we expand them, there a risk they will explode. (Too many elements.)

erszcz: Here's the promised example from dogfooding that I think is related to this problem:

https://github.com/josefs/Gradualizer/blob/74338639c0e0b6903aca0f14d5abb73841955b65/src/typechecker.erl#L406-L414

If we do:

```
 compat_ty({type, _, tuple, Args1}, {type, _, tuple, Args2}, Seen, Env) ->
     %% We can assert because we match out `any' in previous clauses.
     %% TODO: it would be perfect if Gradualizer could refine this type automatically in such a case
-    compat_tys(?assert_type(Args1, [type()]),
+    compat_tys(Args1,
                ?assert_type(Args2, [type()]), Seen, Env);
```

we get the following self-check error:

```
ebin/typechecker.beam: The variable on line 413 at column 16 is expected to have type [type()] but it has type any | [abstract_type()]

compat_ty({type, _, tuple, Args1},
          {type, _, tuple, Args2}, Seen, Env) ->
    compat_tys(Args1,
               ^^^^^
```

This happens even though, as the comment says, the `any` element of that union is explicitly matched two clauses above.

erszcz: Moreover, [Lambda Days 2023](https://www.lambdadays.org/lambdadays2023) took place yesterday and today, so I had a chance to talk with [Guillaume Duboc, who works on a type system for Elixir](https://www.irif.fr/users/gduboc/index). He pointed me to [this paper](https://www.irif.fr/~gc/papers/covcon-again.pdf), which suggests using a [disjunctive normal form](https://en.wikipedia.org/wiki/Disjunctive_normal_form), i.e. a union of intersection, to represent types in the subtyping algorithm. This seems to match my intuition about expanding unions for the sake of subtyping and calculating the `type_diff`. I'll try experimenting with the idea.

PR #528, opened 2023-04-11
Document the compat_seen function

xxdavid:
I finally understand why we need to carry the `Seen` variable. 😄 

Oftentimes, I find the documentation of Gradualizer's codebase unsatisfactory and not very welcoming for newcomers (much useful information is just scattered in GitHub issues and pull requests). Is it okay to document things as I gradually (haha) become familiar with Gradualizer? Also, I'm better at concrete than abstract thinking, and examples help me a lot to understand things. Is it alright to include them in the docs?

### DIFF
--- a/src/typechecker.erl
+++ b/src/typechecker.erl
@@ -263,6 +263,25 @@ fixpoint_normalize(Ty, Env) ->
         _ -> fixpoint_normalize(NormTy, Env)
     end.
 
+%% Stop the recursion if we've already been queried whether T1 <: T2.
+%%
+%% This is because of recursive types. For example, suppose we have
+%% two binary tree types defined like so:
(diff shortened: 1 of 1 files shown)


zuiderkwast:
> Is it okay to document things as I gradually (haha) become familiar with Gradualizer?

Yes, I think it is very much okay to document things. :-)

The 'Seen' trick is used in similar ways in various places, IIRC not only in compat.

erszcz:
I also think it's good to document things :+1:

On the other hand, I'm not sure we'll ever be able to document everything with this level of detail, since a typechecker is quite a dense piece of code with a lot of assumed background knowledge. The case of `Seen` is, for example, described in TaPL - see also the connected https://github.com/josefs/Gradualizer/pull/451#issuecomment-1246692967.

PR #535, opened 2023-05-03
Disallow timeouts in tests

xxdavid:
If a `should_fail` test timed out, it was considered as a success, even though it proably shouldn't as it usually means there's a bug in the typechecker. I've changed it.

Unfortunately, it turned out that there are probably still some infinite loops that are not handled, as after this change two test cases that used recursive types started failing. I've moved them to known problems.


erszcz:
Well spotted, thanks @xxdavid!

PR #538, opened 2023-05-08
Add `apps` option to rebar3 commands

gonzalobf:
This option allows to specify a list of apps where to run the type check. The list can be specify when running the command or in `rebar.config`.

This option can be really handy when trying to integrate Gradualizer on big umbrella project. If the project contains a lot of errors, it is possible to enable the type check app by app and fix the errors incrementally.

### DIFF
--- a/src/rebar_prv_gradualizer.erl
+++ b/src/rebar_prv_gradualizer.erl
@@ -28,7 +28,10 @@ init(State) ->
             {bare, true},
             {deps, ?DEPS},
             {example, "rebar gradualizer"},
-            {opts, [{use_beams, $b, "use_beams", boolean, "use beam files as input"}]},
+            {opts, [
+                {use_beams, $b, "use_beams", boolean, "use beam files as input"},
+                {apps, $a, "apps", string, "Comma separated list of applications to perform type check"}
+            ]},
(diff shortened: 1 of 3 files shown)


erszcz:
Hi, @gonzalobf!

Thanks for the PR - it's a great idea! The CI is almost good, the only non-passing check is Gradualizer itself. Could you take a look at it?

These errors:

```
ebin/rebar_prv_gradualizer.beam: Undefined remote type rebar_state:t/0 on line 40 at column 9
...
```

are ignored in `https://github.com/josefs/Gradualizer/blob/master/gradualize-ignore.lst` - since the line numbers have changed, this file also has to be updated.

src/rebar_prv_gradualizer.erl
```diff
@@ -93,7 +99,9 @@ files_to_check(App) ->
             not lists:member(File, ExpandedExclude)
         end, ExpandedFiles).
 
--spec format_error(any()) -> string().
+-spec format_error(any()) -> iolist().
```
gonzalobf, line 102: The documentation suggests this could be a iolist(). https://www.rebar3.org/docs/tutorials/building_plugins/

erszcz:
Looks good to me, great stuff!

PR #539, opened 2023-05-10
Add known problem: not checking for unreachable clauses with intersections

xxdavid:
When using intersection types, apparently no reachability checking happens.


erszcz:
Indeed, good catch 👍

PR #546, opened 2023-06-16
Prepare release 0.3.0

erszcz:
I hoped for releases every 3 months, but every 6 is still better than every 18, like with the last one ;) @zuiderkwast @xxdavid @josefs do you have any comments? Do you think there's anything else that should be included in this release before we tag it?


josefs:
Go ahead and release IMO. It's better to get releases out through the door so that people can benefit from the improvement than potentially endlessly trying to squeeze just one more feature in before the release.

PR #548, opened 2023-06-20
Fix type inference for `andalso` & `orelse`

xxdavid:
I'm not completely sure but I think it should be this way.


zuiderkwast (approved):
Makes sense!

PR #554, opened 2023-09-06
deprecate dbg:stop_trace/0

MarkoMin:
This function will be removed in OTP27, so this commit is necessary at some point.


erszcz:
Thanks, @MarkoMin 🙏

PR #557, opened 2023-09-21
Add a few test cases

xxdavid:
Discovered while testing Gradualizer on an Elixir repo.

### DIFF
--- a/test/known_problems/should_pass/elixir_list_first.erl
+++ b/test/known_problems/should_pass/elixir_list_first.erl
@@ -1,0 +1,17 @@
+-module(elixir_list_first).
+
+-export([f/0]).
+
+% This is how List.first/1 is defined in Elixir
+-spec first([]) -> nil;
+           ([Elem, ...]) -> Elem.
+first([]) -> nil;
+first([Element | _]) -> Element.
+
+-spec f() -> ok.
+f() ->
+    first(some_list()),
+    ok.
+
+-spec some_list() -> list().
+some_list() -> [anything].
--- a/test/known_problems/should_pass/refine_comparison_should_pass.erl
+++ b/test/known_problems/should_pass/refine_comparison_should_pass.erl
@@ -1,0 +1,13 @@
+-module(refine_comparison_should_pass).
+
+-export([comp_map_value3/1]).
+
+-type my_map() :: #{value := integer() | nil}.
+
+-spec comp_map_value3(my_map()) -> integer().
+comp_map_value3(State) when map_get(value, State) /= nil ->
+    case State of
+        #{value := Val} ->
+            Val + 1
+    end;
+comp_map_value3(#{value := nil}) -> 0.
--- a/test/should_pass/refine_comparison.erl
+++ b/test/should_pass/refine_comparison.erl
@@ -28,3 +28,20 @@ compatom1(_) -> ok.
 -spec compatom2(a | b | c | pid()) -> a | c | pid().
 compatom2(X) when X =/= b -> X;
 compatom2(_) -> a.
+
+-spec compatom3(a | b | c | pid()) -> a | c | pid().
+compatom3(X) when X /= b -> X;
+compatom3(_) -> a.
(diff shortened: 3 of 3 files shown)


test/known_problems/should_pass/elixir_list_first.erl
```diff
@@ -0,0 +1,17 @@
+-module(elixir_list_first).
+
+-export([f/0]).
+
+% This is how List.first/1 is defined in Elixir
+-spec first([]) -> nil;
+           ([Elem, ...]) -> Elem.
+first([]) -> nil;
+first(Element) -> Element.
```
zuiderkwast, line 9: Typo?

```suggestion
first([Element | _]) -> Element.
```

test/known_problems/should_pass/guard_negative_refinement_should_pass.erl
```diff
@@ -0,0 +1,15 @@
+-module(guard_negative_refinement_should_pass).
```
zuiderkwast, line 1: Negative? I believe `not_equal` is better when talking about the `/=` operator.

Btw, do we have tests for guards with `=/=`? And what about `==` and `=:=`?

test/known_problems/should_pass/guard_negative_refinement_should_pass.erl
```diff
@@ -0,0 +1,15 @@
...
+
+-export([i/1]).
+
+-type state() :: #{value := integer() | nil}.
+
+-spec i(state()) -> integer().
+i(State) when map_get(value, State) /= nil ->
```
zuiderkwast, line 8: This guard can fail. Shouldn't we have another clause for value nil? (Not that we can detect this non-exhaustiveness now, but the code makes more sense that way and maybe in the future we can check it.)

zuiderkwast (commented):
Nice

test/known_problems/should_pass/elixir_list_first.erl
xxdavid: Yes, of course, thank you. 🙂

test/known_problems/should_pass/guard_negative_refinement_should_pass.erl
xxdavid: Good point, I've added a clause for `nil` (d3b5fa06).

xxdavid: Fixed: 0c9307cc. I've also added `nil` cases for the other functions (2b2c5110).

test/known_problems/should_pass/guard_negative_refinement_should_pass.erl
xxdavid: Hmm, the *negative* part was meant like *type negation* (e.g. `(integer() | nil) \ nil`) but this is already contained in the *refinement*… So maybe it should be only `guard_refinement.erl`? There's one file that already tests guard refinements (including `=/=`, `==`, and `=:=`), `test/should_pass/refine_comparison.erl`, so maybe I should merge the files and put it there? What do you think?

zuiderkwast: Yeah... merge the two, or rename the other one, or both, or what you find appropriate. :) Now you have the full picture.

xxdavid: Okay. :) I merged it into the `test/should_pass/refine_comparison.erl` and created `test/known_problems/should_pass/refine_comparison_should_pass.erl` with the failing test. I've also squashed the fixups and force pushed.

zuiderkwast (approved):
LGTM

Btw, maybe you should consider patching your editor. :-) A decent editor automatically adds a newline at end-of-file.

xxdavid:
Oh, thanks for noticing. Newlines added, editor configured (I though it was already like that but apparently not).

And thank you for the review. :)

Issue #553, opened 2023-08-26
Solving local constraints at each function call to a polymorphic function


xxdavid:
Hi!

I've read the [Local Type Inference](https://www.cis.upenn.edu/~bcpierce/papers/lti-toplas.pdf) paper by Pierce and Turner, and I came to an interesting realisation I would like to share.

In the paper (mainly in chapter three), they discuss a way to infer missing type application annotations / type arguments (so that instead of `id [Int] (3)` we can write just `id (3)`, provided that `id` is of type `∀X. X -> X`). Their approach is *local* in the way that they only introduce a new inference rule for function applications without the type application annotation, the rule computes the best substitution (intuitively the one which preservers the most information; based on *locally gathered constraints* that are *immediately solved*) and *immediately uses the substitution to substitute type variables in the type of the function call expression*.

This is in contrast with what we are doing in Gradualizer where we only *gather* all the constraints when checking a function, and only *then*, after we checked the function, we solve the constraints. If the constraints cannot be solved, we conclude that *somewhere in the function* there was a polymorphic type error (usually that the lower bound of some type variable is not a subtype of some other type). If they can be solved, we are happy and discard the found substitution.

Let me illustrate it on a simple example:
```erlang
   1  -module(constraints_example).
   2  -export([g/1]).
   3
   4  -spec g({boolean(), map()}) -> ok.
   5  g(Tuple) ->
   6      X = first(Tuple),
   7      add_one(X),
   8      ok.
   9
  10  -spec first({A, any()}) -> A.
  11  first({X, _Y}) -> X.
  12
  13  -spec add_one(integer()) -> integer().
  14  add_one(N) -> N + 1.
```

Gradualizer (with `--solve_constraints` works like this):
```
Checking function g/1
Pushing {clauses_controls,true}
5:1: Checking clause :: ok
5:3: Setting var type Tuple :: {boolean(), map()}
6:15: Checking that Tuple :: {A_typechecker_3742_1, any()}
6:9: Propagated type of first(Tuple) :: A_typechecker_3742_1
6:5: Setting var type X :: A_typechecker_3742_1
6:5: Propagated type of X = first(Tuple) :: A_typechecker_3742_1
7:13: Checking that X :: integer()
7:5: Propagated type of add_one(X) :: integer()
8:5: Checking that ok :: ok
Exhaustiveness checking: true
Popping clauses controls
Constraints before solving:
{constraints,#{"A_typechecker_3742_1" =>
[...]

PR #560, opened 2024-01-22
Support for rigid type variables

xxdavid:
Hi!

I had almost no time to work on the local type inference (#553) in the past months, but I finally found some time last week. After thinking about the implementation, I've decided that a good first step towards LTI would be to add the missing support for rigid type variables. This way we can check bodies of polymorphic functions. And here's a PR just for that.

Rigid type variables are usually treated as type you know nothing about. You cannot assume anything about them and almost no subtyping rules apply to them (except for reflexivity, top, bottom, and `any()`).  When typechecking a function, all of the type variables mentioned in its spec should be considered as rigid, because they can be anything (it is up to the caller what they are instantiated to (in rank-1 polymorphism)).

With this, we can discover type errors like this one:

```erlang
-spec id(A) -> A.
id(_X) -> 42.
```

The implementation in this PR tells us that in this case:
```
The integer on line 2 at column 11 is expected to have type A but it has type 42
```

When implementing rigid type variables, we have to somehow distinguish between flexible and rigid type variables. I came up with three possible ways of doing so:

1. Adding a new `type()` shape: `{rigid_type, anno(), atom()}`
2. Adding a new `anno()` shape: `{is_rigid, boolean()}`
3. Keeping the kind (flexible/rigid) of type variables in the current scope in `Env`

I turned down option number two, as we could not use pattern matching or guards to distinguish between flexible and rigid type variables. Option three could come with a macro guard `is_flexible(Var, Env)` (which is nice) but would require us to use it in every place where type variables are used in typechecking. This seems quite fragile to me, and thus I went for the option number one. A good thing about this approach is that it required no changes to the typechecking core (`subtype`, `type_check_expr`, etc.), only a few utility functions were enriched with a new clause for `rigid_var`. The only hassle was that `erl_pp` doesn't expect `rigid_var` and prints `INVALID-FORM` in that case, so we have to transform every type before printing it, but this seems okay to me.

I added tests for rigid type variables and I think it works really well. I also moved `poly_2` test functions with explanation in the commit message, feel free to react to that as well.

What do you think about it? @erszcz @zuiderkwast


src/gradualizer_type.erl
```diff
@@ -308,6 +313,8 @@
 -type gr_type_var() :: atom() | string().
 -type af_type_variable() :: {'var', anno(), gr_type_var()}.
 
+-type gr_rigid_type_variable() :: {'rigid_var', anno(), atom()}.
```
zuiderkwast, line 316: Since most of this file is copied from OTP, we want to explain all our changes with comments. The prefix "gr" is good but maybe something more about why we want to add this type?

src/gradualizer_type.erl
```diff
@@ -338,6 +345,8 @@
                           [af_lit_atom('is_subtype') |
                            [af_type_variable() | abstract_type()]]}. % [IsSubtype, [V, T]]
 
+-type gr_any_type() :: {'type', anno(), 'any', []}.
```
zuiderkwast, line 348: Why do we need to add this?

I think it's covered by this definition:

```erlang
-type af_predefined_type() ::
        {'type', anno(), type_name(),  [abstract_type()]}.
```

test/should_fail/opaque_fail.erl
```diff
@@ -1,7 +1,24 @@
...
+add_to_opaque() ->
+    Val = user_types:new_opaque(),
+    Val + 1,
+    ok.
+
+-spec return_opaque() -> user_types:my_opaque().
+return_opaque() -> 3.
+
+-spec abc(A) -> A.
+abc(X) -> X.
```
zuiderkwast, line 24: Unused function?

zuiderkwast (commented):
This looks awesome!

I didn't really know how to implement and use rigid type vars but looks very strait forward.

xxdavid:
Thank you for the review! 🙂

src/gradualizer_type.erl
xxdavid: Good point, I added the missing comment, together with a forgotten doc for the `make_rigid_type_vars/1` (with a slightly different wording): 5a18b8eb. Is it better now?

test/should_fail/opaque_fail.erl
xxdavid: Good catch, this is probably some leftover from some of my testings. Removed in 71395366.

src/gradualizer_type.erl
xxdavid: It's not necessarily needed. I used it [here](https://github.com/josefs/Gradualizer/blob/5d9bfccaf4a16972f9ee28c1f7e798cd34642411/src/typechecker.hrl#L8) to emphasise that `fenv` carries either a spec or simply `any()` (wrapped in a list) for each function (as it wasn't very apparent to me). We could inline it there like this:
```erlang
-record(env, {fenv              = #{}   :: #{{atom(), arity()} =>
                                                  [gradualizer_type:af_constrained_function_type()]
                                                | [{'type', gradualizer_type:anno(), 'any', []}]
                                            },
```
but I'm not sure whether it's better. What would you do?

zuiderkwast: OK. I'm fine with any. :-)

xxdavid:
Hey guys, could we please move this forward? 🙂  Is there anything left you want to change or comment on?

I've just fixed the authorship of the commits (there was a wrong email address used in them; otherwise the commits are the same) and force-pushed them. I also added one small fixup commit (a103f65).

zuiderkwast (approved):
Sorry @xxdavid, I think me and @erszcz are too busy with other projects.

I don't have a full understanding of how rigid type variables affects other ideas such as constraint solver, but I agree you're solving important problems here.

Please add relevant documentation. We use the wiki for documentation: https://github.com/josefs/Gradualizer/wiki. I hope you have write permissions to the wiki.

xxdavid:
@zuiderkwast I get it. Thank you for merging the PR.

Okay, I'll add a chapter related to Polymorphism to the wiki. I think it can cover both directions: checking the polymorphic functions itself (this involves the rigid type variables) and checking calls to such polymorphic functions (this will hopefully use the LTI approach I described in #553, I will make a PR for that soon). But I am afraid I don't have an edit permission to the wiki.

zuiderkwast:
@josefs, do you want to give write permissions to @xxdavid in this repo?

He is currently the most active person in this project.

Alternatively, allow everyone to edit the wiki pages by unchecking "Restrict editing to collaborators only" under settings:

image

Thanks.

josefs:
> @josefs, do you want to give write permissions to @xxdavid in this repo?

Done

xxdavid:
Thanks!

PR #566, opened 2024-05-31
Define subtyping of unions with type difference

erszcz:
This borrows the idea on how to define the subtyping relation from set-theoretic types, but we only use it for subtype of unions. This issue was made clearly visible in @xxdavid's #564 and we spoke about it face to face on Lambda Days earlier this week.

However, there are some basic tests which stopped passing, like `test/should_pass/guard.erl`, so it needs more investigation. The main troublemaker, `different_normalization_levels.erl`, works now, though.


xxdavid:
Impressive, thanks for trying it out! It breaks when `any()` is used, for example `any() | not_found <: float() | not_found` and `false | true <: integer() | any()` do not hold any more, but this can probably be tweaked somehow.

PR #567, opened 2024-06-02
Use Common Test for testing when Gradualize should pass, fail, and its known problems

erszcz:
This PR adds ability to test Gradualizer functionality with Common Test, which reports results in a way that enables much more convenient comparisons between builds. CT results are persisted across runs in HTML. The Makefile rule, `make ct`, also labels runs with git commit info and a _modified_ annotation, if a run tested a modified working directory. All in all, this makes figuring out how test results differ based on a code change much easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,9 @@ eunit: compile-tests
 	erl $(ERL_OPTS) -noinput -pa ebin -pa test -eval \
 	 '$(erl_run_eunit), halt().'
 
+ct:
+	@rebar3 ct --label "git: $$(git describe --tags --always) $$(git diff --no-ext-diff --quiet --exit-code || echo '(modified)')"
+
 cli-tests: bin/gradualizer test/arg.beam
 	# CLI test cases
 	# 1. When checking a dir with erl files, erl file names are printed
--- a/test/gradualizer_dynamic_suite.erl
+++ b/test/gradualizer_dynamic_suite.erl
@@ -1,0 +1,50 @@
+-module(gradualizer_dynamic_suite).
+
+-export([reload/1]).
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("stdlib/include/assert.hrl").
+
+reload(Config) ->
+    Module = ?config(dynamic_suite_module, Config),
+    Path = ?config(dynamic_suite_test_path, Config),
+    ?assert(Module /= undefined),
+    ?assert(Path /= undefined),
+    Forms = get_forms(Module),
+    FilesForms = map_erl_files(fun (File) ->
+                                       make_test_form(Forms, File, Config)
+                               end, Path),
+    {TestFiles, TestForms} = lists:unzip(FilesForms),
+    TestNames = [ list_to_atom(filename:basename(File, ".erl")) || File <- TestFiles ],
+    ct:pal("All tests found under ~s:\n~p\n", [Path, TestNames]),
+    NewForms = Forms ++ TestForms ++ [{eof, 0}],
+    {ok, _} = merl:compile_and_load(NewForms),
+    {ok, TestNames}.
+
+map_erl_files(Fun, Dir) ->
+    Files = filelib:wildcard(filename:join(Dir, "*.erl")),
+    [{filename:basename(File), Fun(File)} || File <- Files].
+
+make_test_form(Forms, File, Config) ->
+    TestTemplateName = ?config(dynamic_test_template, Config),
+    ?assert(TestTemplateName /= undefined),
+    TestTemplate = merl:quote("'@Name'(_) -> _@Body."),
+    {function, _Anno, _Name, 1, Clauses} = lists:keyfind(TestTemplateName, 3, Forms),
+    [{clause, _, _Args, _Guards, ClauseBodyTemplate}] = Clauses,
+    TestName = filename:basename(File, ".erl"),
+    ClauseBody = merl:subst(ClauseBodyTemplate, [{'File', erl_syntax:string(File)}]),
+    TestEnv = [
+               {'Name', erl_syntax:atom(TestName)},
+               {'Body', ClauseBody}
+              ],
+    erl_syntax:revert(merl:subst(TestTemplate, TestEnv)).
+
+get_forms(Module) ->
+    ModPath = code:which(Module),
+    {ok, {Module, [Abst]}} = beam_lib:chunks(ModPath, [abstract_code]),
+    {abstract_code, {raw_abstract_v1, Forms}} = Abst,
+    StripEnd = fun
+                   ({eof, _}) -> false;
+                   (_) -> true
+               end,
+    lists:filter(StripEnd, Forms).
--- a/test/known_problems_should_fail_SUITE.erl
+++ b/test/known_problems_should_fail_SUITE.erl
@@ -1,0 +1,87 @@
+-module(known_problems_should_fail_SUITE).
+
+-compile([export_all, nowarn_export_all]).
+
+%% EUnit has some handy macros, so let's use it, too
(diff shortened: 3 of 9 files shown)


test/should_pass_SUITE.erl
```diff
@@ -0,0 +1,107 @@
...
+     spec_and_fun_clause_intersection_pass,stuff_as_top,'try',try_expr,tuple,
+     tuple_union_pass,tuple_union_pat,tuple_union_pattern_pass,type_decl,
+     type_pattern,type_refinement_pass,type_variable,type_vars_term,
+     typed_record_field_access,unary_negate_union_with_user_type_pass,unary_plus,
+     underscore,user_type_in_pattern_body,user_types,var,var_fun,varbind_in_block,
+     varbind_in_case,varbind_in_function_head,varbind_in_lc,variable_binding,
+     variable_binding_leaks].
+
+should_pass_template(_@File) ->
+    ?assertEqual(ok, gradualizer:type_check_file(_@File, [{form_check_timeout_ms, 2000}])).
```
xxdavid, line 107: The EUnit tests print a Gradualizer's error message describing the type error when a test fails. I find it useful for quickly glancing over what's going on without having to inspect all the failed test files one by one. Do you think it's possible with CT?

xxdavid (commented):
I am not really familiar with Common Test and I also don't fully understand all the magic in `gradualizer_dynamic_suite`. But under this disclaimer, this PR looks quite good to me (apart from a few comments I left). The output is nicer compared to EUnit and, as you said, it's probably easier to compare between different versions of Gradualizer. It's a bit slower than the EUnit tests (15 secs vs 10 secs on my machine) but that's not a problem I think. Overall, I consider it as an improvement over the EUnit tests. Thanks for implementing it!

Do you plan to replace the EUnit tests for `should_*` with these CT tests, or do you want them to coexist (which would cause duplicated code)?

Thanks for extending the PR description and adding screenshots, it made it much easier to grasp the intended usage. If this gets merged, maybe we should also document it somewhere.

test/should_fail_SUITE.erl
```diff
@@ -0,0 +1,111 @@
...
+    ok.
+
+all() ->
+    [{group, all_tests}].
+
+groups() ->
+    [{all_tests, [parallel], all_tests()}].
+
+all_tests() ->
+    [annotated_types_fail,arg,arith_op_fail,arity_mismatch,
```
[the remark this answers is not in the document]
erszcz: Yes, I managed to find a way to do it 👍

test/should_pass_SUITE.erl
erszcz: It's available in the CT HTML report, but that requires a little bit of clicking in the browser. It might be possible to get in the shell, too, but I'd have to think some more about it.

Makefile
```diff
@@ -149,6 +149,9 @@ eunit: compile-tests
 	erl $(ERL_OPTS) -noinput -pa ebin -pa test -eval \
 	 '$(erl_run_eunit), halt().'
 
+ct:
+	@rebar3 ct --label "git: $$(git describe --tags --always) $$(git diff --no-ext-diff --quiet --exit-code || echo '(modified)')"
```
zuiderkwast, line 153: We made this whole makefile work without rebar3 because of some annoyances we had with it and to get better control of what's happening. If we mix rebar3 with non-rebar3 we'll have files built in various place and a mess in general. I don't like that.

Isn't it fairly straitforward to run `ct` without rebar3? It's a command line tool.

Add ct to the `tests` target and to `.PHONY`.

We should be able to modify the logic we have in `erl_cover_run` function to have coverage computed on eunit and commontest combined, or only commontest if we port all eunit tests to commontest.

To run a specific suite, we can do something similar to what erlang.mk does, e.g. `make ct suite=known_problems_should_pass_SUITE`.

test/gradualizer_dynamic_suite.erl
```diff
@@ -0,0 +1,60 @@
...
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("stdlib/include/assert.hrl").
+
+reload(Config) ->
+    Module = ?config(dynamic_suite_module, Config),
+    ?assert(Module /= undefined),
+    case erlang:function_exported(Module, generated_tests, 0) of
+        true ->
+            {ok, Module:generated_tests()};
```
zuiderkwast, line 13: What's this test suite doing?

It looks like a lot of magic. I don't like magic. :)

test/known_problems_should_fail_SUITE.erl
```diff
@@ -0,0 +1,63 @@
+-module(known_problems_should_fail_SUITE).
```
zuiderkwast, line 1: Did you duplicate the eunit test suites as commontest suite?

I don't want duplicated logic. Delete the eunit test suites that have been ported to commontest.

test/should_pass_SUITE.erl
zuiderkwast: IIRC, CT supports plain text reports too. I don't know it well, but if it's too complicated, maybe it isn's worth it.

If we run it ourselves instead of relying on rebar3, we'll have better control over it.

zuiderkwast (changes requested):

Makefile
erszcz: > We made this whole makefile work without rebar3...

Generally, I think rebar3 is quite mature and I'd happily drop the Makefile, to be honest. though I admit I see some small benefits of using a tailored Makefile.

> Isn't it fairly straitforward to run ct without rebar3? It's a command line tool.

It's possible, but the nice CLI output is provided by a custom Rebar3 CT hook, so if we run CT directly, we'll get uglier printouts.

> Add ct to the `tests` target and to `.PHONY`.

If we're happy with moving completely to CT, I can do that. For now, I considered this a nicer alternative for local development (especially comparing test results across builds), but did not include it in the CI, nor did I remove the original EUnit tests this is based on. In this light, it didn't make sense to run them twice, so the CT variants are not in `tests`.

test/known_problems_should_fail_SUITE.erl
erszcz: I did, yes. If we're happy with moving completely to CT, I'll delete the EUnit tests.

test/gradualizer_dynamic_suite.erl
erszcz: There's a little bit of magic, but let's start from the standard CT conventions. CT treats files matching `*_SUITE.erl` as test suites, so `gradualizer_dynamic_suite.erl` is not a test suite, it's a helper file for use in test suites. The actual test suites match the previous tests defined in EUnit:
- `test/known_problems_should_fail_SUITE.erl`
- `test/known_problems_should_pass_SUITE.erl`
- `test/should_fail_SUITE.erl`
- `test/should_pass_SUITE.erl`.

The dynamic suite helper exports only one function - `gradualizer_dynamic_suite:reload/1` - which generates CT test cases based on passed in config and then dynamically reloads the module it's invoked from. It's idempotent, so calling it once, twice, or 100 times has the same effect. That's where the name comes from - "dynamic suite". The reason for this helper is that, unlike EUnit, CT has no standard way to generate tests dynamically.

The tests are generated for each of the test files defined under the respective should pass/fail/known problems directories.

The most enigmatic part is where to actually invoke `gradualizer_dynamic_suite:reload/1` in a client test suite - this is something I had to dive in OTP CT code to figure out. Apparently, `*_SUITE:groups/0` is the first function CT calls when running a test suite, so it's the best place to call `reload/1` from. Before @xxdavid's remark that listing cases by hand is tedious we could call `reload/1` from something more logical, like `init_per_suite/1`, but if we want to avoid listing the to-be-generated tests manually, we have to play with how CT runs the test suites.

PR #570, opened 2024-09-17
Skip too complex guards

erszcz:
This PR skips processing functions which use complex guards in order to make Gradualizer more convenient for everyday use, based on the fact that such complex guards aren't handled properly yet. It's important to underline that this PR doesn't lead to fewer or inferior warnings, since the warnings that are skipped could only be false positives anyway, due to incorrect/missing parameter refinement.

This enables Gradualizer users to more conveniently use the tool on "random", i.e. Gradualizer-unaware, projects, thanks to **fewer false positives being raised**. The PR also contains examples of refactoring (some) unhandled Erlang constructs into equivalent ones that are already handled completely by the tool.

Gradualizer developers can see which Erlang constructs and what guards would be worth handling next. We can also see that with this change some assertions can be dropped from Gradualizer code, yet maintaining that the tool can self-check.

### DIFF
--- a/src/typechecker.hrl
+++ b/src/typechecker.hrl
@@ -3,23 +3,26 @@
 
 -record(clauses_controls, {exhaust}).
 
--record(env, {fenv              = #{}   :: #{{atom(), arity()} =>
-                                                  [gradualizer_type:af_constrained_function_type()]
-                                                | [gradualizer_type:gr_any_type()]
-                                            },
-              imported          = #{}   :: #{{atom(), arity()} => module()},
-              venv              = #{}   :: typechecker:venv(),
-              tenv                      :: gradualizer_lib:tenv(),
-              infer             = false :: boolean(),
-              verbose           = false :: boolean(),
-              exhaust           = true  :: boolean(),
+-record(env, {fenv              = #{}           :: #{{atom(), arity()} =>
+                                                          [gradualizer_type:af_constrained_function_type()]
+                                                        | [gradualizer_type:gr_any_type()]
+                                                    },
+              imported          = #{}           :: #{{atom(), arity()} => module()},
+              venv              = #{}           :: typechecker:venv(),
+              tenv                              :: gradualizer_lib:tenv(),
+              infer             = false         :: boolean(),
+              verbose           = false         :: boolean(),
+              exhaust           = true          :: boolean(),
               %% Controls driving the type checking algorithm
               %% per clauses-list (fun/case/try-catch/receive).
-              clauses_stack     = []    :: [#clauses_controls{}],
+              clauses_stack     = []            :: [#clauses_controls{}],
               %% Performance hack: Unions larger than this limit are replaced by any() in normalization.
-              union_size_limit          :: non_neg_integer(),
-              current_spec      = none  :: erl_parse:abstract_form() | none,
-              solve_constraints = false :: boolean()
+              union_size_limit                  :: non_neg_integer(),
+              current_spec      = none          :: erl_parse:abstract_form() | none,
+              solve_constraints = false         :: boolean(),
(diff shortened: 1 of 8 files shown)


src/typechecker.hrl
```diff
@@ -3,23 +3,26 @@
...
               %% Performance hack: Unions larger than this limit are replaced by any() in normalization.
-              union_size_limit          :: non_neg_integer(),
-              current_spec      = none  :: erl_parse:abstract_form() | none,
-              solve_constraints = false :: boolean()
+              union_size_limit                  :: non_neg_integer(),
+              current_spec      = none          :: erl_parse:abstract_form() | none,
+              solve_constraints = false         :: boolean(),
+              %% Skip functions whose guards are too complex to be handled yet,
+              %% which might result in false positives when checking rest of the functions
+              no_skip_complex_guards = false    :: boolean()
```
xxdavid, line 25: I appreciate adding the option (as it enables us to test Gradualizer's behaviour with/without it easily). I would also add the CLI option for it (to `src/gradualizer_cli.erl` and to the `README.md`).

xxdavid (commented):
Hi again! I am also for lowering the number of false positives in cases our code is not smart enough yet. To be honest, I am not really sure of the scope/implications of the changes in this PR though. Do I understand it correctly that all functions that contain any guard other than a single `is_TYPE/1` (such as `is_integer/1`) guard are now skipped (as `refine_vars_by_mismatching_clause/3` throws the `{skip, too_complex_guards}` tuple?

src/typechecker.hrl
erszcz: I intentionally skipped the CLI option. I don't think it should ever be needed for users to toggle this and I think it's worth trying to limit the number of obscure options. I rather consider it as a means to still be able to test the full capacity of the typechecker, even if the users won't ever use the incomplete/buggy behaviour.

test/known_problems/should_fail/type_refinement_should_fail.erl
```diff
@@ -0,0 +1,19 @@
...
+-spec guard_prevents_refinement(1..2, boolean()) -> 2.
+guard_prevents_refinement(N, Guard) ->
+    case N of
+        1 when Guard -> 2;
+        M            -> M  %% 1 cannot be eliminated
+    end.
+
+-spec guard_prevents_refinement2(erlang:timestamp()) -> ok.
+guard_prevents_refinement2(X) when is_integer(X), X rem 7 == 0 -> ok;
+guard_prevents_refinement2(infinity) -> ok. % can still be an integer
```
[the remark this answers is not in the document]
erszcz: Good point, I think it's just a mistake in the spec which assumes that `erlang:timestamp()` is a Unix timestamp, i.e. an integer. I'll just fix it to make sense.

erszcz:
>  Do I understand it correctly that all functions that contain any guard other than a single `is_TYPE/1` (such as `is_integer/1`) guard are now skipped (as `refine_vars_by_mismatching_clause/3` throws the `{skip, too_complex_guards}` tuple?

Yes. I don't really see a better way to make the usability reasonable and at the same time not spend significant time on code which would just override the argument types to `any()` in presence of too complex guards. It's still possible to avoid skipping by refactoring code to use just patterns or simple guards, thanks to verbose logging.

Proper tackling of complex guards requires two things: precise inference of types based on all operators (e.g. binary `+` might currently "refine" to a broader type, see #572) and negative types. Neither is already available, so I think reaching for the lowest-hanging fruit is just to skip too complex syntax, yet be useful otherwise.

src/typechecker.hrl
xxdavid: Okay then. :)

PR #571, opened 2024-09-17
Skip functions using maybe expr, don't crash on maybe ... else

erszcz:
This handles `maybe ... else` instead of crashing. It also skips processing functions which use `maybe` similar to how #570 skips functions that use too complex guards.


src/typechecker.erl
```diff
@@ -2232,8 +2232,10 @@ do_type_check_expr(Env, {'try', _, Block, CaseCs, CatchCs, AfterBlock}) ->
 
 %% Maybe - value-based error handling expression
 %% See https://www.erlang.org/eeps/eep-0049
-do_type_check_expr(_Env, {'maybe', Anno, [{maybe_match, _, _LHS, _RHS}]} = MaybeExpr) ->
-    erlang:throw({unsupported_expression, Anno, MaybeExpr}).
+do_type_check_expr(_Env, {'maybe', _, _}) ->
+    erlang:throw({skip, maybe_expr_not_supported});
```
xxdavid, line 2236: I suppose you are going to rebase this on top of the changes in #570, right?

src/typechecker.erl
```diff
@@ -2232,8 +2232,10 @@ do_type_check_expr(Env, {'try', _, Block, CaseCs, CatchCs, AfterBlock}) ->
 
 %% Maybe - value-based error handling expression
 %% See https://www.erlang.org/eeps/eep-0049
-do_type_check_expr(_Env, {'maybe', Anno, [{maybe_match, _, _LHS, _RHS}]} = MaybeExpr) ->
-    erlang:throw({unsupported_expression, Anno, MaybeExpr}).
+do_type_check_expr(_Env, {'maybe', _, _}) ->
```
xxdavid, line 2235: I think that equivalent code should be also added to `do_type_check_expr_in`.

xxdavid (commented):
Hey Radek! I agree that not crashing on `maybe`/`maybe … else` is a good idea. 🙂 

I also agree that it is a bit better to skip these function silently (as in this PR) rather than with a warning (as is the current behaviour), as one may strive for zero Gradualizer warnings (for example in a CI) and the maybe warning would complicate that. However, maybe even a bit better solution would be to skip just the `maybe` *expressions* (instead of skipping whole functions) and assign the expression the type `any()`. I am not sure how much work would that be (if only a small change in `type_check_expr(_in)` would be enough or whether other places would need to be updated as well) though…

src/typechecker.erl
erszcz: Yes, exactly

erszcz: I applied your suggestion here:

> maybe even a bit better solution would be to skip just the maybe expressions (instead of skipping whole functions) and assign the expression the type any().

It might enable working with a bit more real-world code, indeed.

src/typechecker.erl
erszcz: Good point, added. In the case of `do_type_check_expr_in`, though, it's not possible to assume `any()`, since the operation is checking (as in _checking_ vs _inference_ in bidirectional typing), so it would certainly lead to false positives. I think it's better to skip such functions altogether for now.

src/typechecker.erl
xxdavid: Thanks!

PR #573, opened 2024-10-11
Drop the --infer flag

erszcz:
This removes the `--infer` CLI and `#env{}` option. The Erlang VM does dynamic type checking based on type tags embedded in values. This means that it's safe to infer type information from literals and operators according to the same rules that the VM enforces at runtime. Actually, these are one of the few completely credible sources of information, unlike type specs, in which there's room for human error. Removing the option along with the branching in the typechecker should somewhat minimise the amount of code to maintain. Moreover, requiring users to know the available options might lead to a worse impression of what the typechecker is capable of, since if they're not used, the tool is not showing it's full potential.

### DIFF
--- a/test/typechecker_tests.erl
+++ b/test/typechecker_tests.erl
@@ -365,20 +365,18 @@ normalize_e2e_test_() ->
     ].
 
 propagate_types_test_() ->
-    %% Checking type_check_expr, which propagates type information but doesn't
-    %% infer types of literals and language constructs.
-    [%% the inferred type of a literal tuple should be any()
-     ?_assertEqual("any()",
+    %% Checking type_check_expr, which propagates type information
+    %% and infers types of literals and language constructs.
+    [%% the inferred type of a literal tuple
+     ?_assertEqual("{1, 2}",
                    type_check_expr(_Env = "",
                                    _Expr = "{1, 2}")),
      %% the inferred type of a tuple with a untyped function call
-     %% should also be any()
-     ?_assertMatch("any()",
+     ?_assertMatch("{1, any()}",
                    type_check_expr(_Env = "h() -> 2.",
                                    _Expr = "{1, h()}")),
      %% the inferred type of a tuple with a typed function call
-     %% should be {any(), restype()}
-     ?_assertMatch("{any(), integer()}",
+     ?_assertMatch("{1, integer()}",
                    type_check_expr(_Env = "-spec h() -> integer().",
                                    _Expr = "{1, h()}")),
 
@@ -389,7 +387,7 @@ propagate_types_test_() ->
                                           "-spec c() -> 3.",
                                    _Expr = "[a(), b(), c()]")),
      %% Fun.
-     ?_assertMatch("any()",
+     ?_assertMatch("fun((any(), any()) -> any())",
                    type_check_expr(_Env = "",
                                    _Expr = "fun (_A, _B) ->\n"
                                            "    receive C -> C end\n"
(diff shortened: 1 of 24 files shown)


test/typechecker_tests.erl
```diff
@@ -456,52 +454,43 @@ type_check_in_test_() ->
...
      %% Although there is no spec for f/1 - inferred arity does not match
      ?_assertNot(type_check_forms(["-spec g() -> fun(() -> integer()).",
                                    "g() -> fun f/1.",
-                                   "f(_) -> ok."],
-                                  [infer]))
+                                   "f(_) -> ok."]))
     ].
 
 infer_types_test_() ->
     %% Checking type_check_expr with inference enabled
```
xxdavid, line 465: I think comment can be now reworded.

test/typechecker_tests.erl
```diff
@@ -456,52 +454,43 @@ type_check_in_test_() ->
...
-                                   _Expr = "\"0123456789abc\"",
-                                   [infer])),
+                                   _Expr = "\"0123456789abc\"")),
      %% infer exact type of record index
      ?_assertMatch("2",
                    type_check_expr(_Env = "-record(r, {f}).",
-                                   _Expr = "#r.f",
-                                   [infer]))
+                                   _Expr = "#r.f"))
     ].
```
xxdavid, line 494: This is now checked in the `propagate_types_test_` as well.

xxdavid (approved):
To be honest, I never fully understood why `infer` wasn't the default (and only) mode. If I recall the discussions correctly, it was added by @zuiderkwast but never fully supported by @josefs, as @josefs wanted to rely solely on typespecs as part of the gradual typing paradigm ("without typespecs, no checking happens, it is up to the programmer to enable checking by adding typespecs, the programmer is in control of what is checked and what is not"). I don't really share this perspective on gradual typing (but I do respect it!) and I agree with you that literals are a trustworthy source of type information.

The pull request itself seems legit to me and I personally approve it, but that's only my take on it.

josefs:
Since I'm no longer active in this project, I don't have an opinion on this PR. But since I was pinged, I thought I'd like to clarify my position here. 

When I set out build gradualizer, an important goal for me was to be able to start with an existing Erlang code base and run gradualizer without any errors (or perhaps, if there was already some erroneous type specs, they would be highlighted, but nothing more). Being able to apply gradualizer with minimal cost to an existing code base without being overwhelmed with type errors immediately would lower the threshold to adoption. Then, one could progressively add more and more type specs and get gradually more benefit from gradualizer. The goal was to provide a pay-as-you-go experience. Enabling inference could be added later, as it could potentially generate a lot of type errors up front. It certainly would have for the code base I used to work on.

I had planned, but never implemented, a strict mode which would require type specs everywhere in a file, no `any()` types anywhere, inference, and all the imports to also be completely typed. The goal for any code base would be to eventually be checked in strict mode everywhere.

zuiderkwast:
Hey Josef! I'm not that active either anymore.

Eqwalizer has a strict and a non-strict mode, I believe. They seem to provide some real guarantees by not supporting the detailed type hierarchy (no subtyping?), but it may be a more useful tool in practice? (I don't know.)

David, thanks for reminding me of the discussions around this. I recall that Josef's idea initially was to rely on typespecs only and nothing else, and I supported that. The more we implemented and tested it, though, the more it became apparent that this approach had some problems, like all the OTP library functions having type specs, while arithmetic operators and literals not having any, was a bit problematic. I argued that we should consider other sources of type information too. (What should be used as type information and why?)

The graduality guarantee ("gradual soundness") vs eliminating false positives was another balance between theoretical correctness and practical usefulness, and I believe still is. I found it hard to accept that we can't fully achieve both. (I still haven't accepted it. :grin:)

PR #574, opened 2024-10-15
Unify type_check_expr and type_check_expr_in

erszcz:
This attempts the unification discussed in #362. The are some obvious takeaways from this exercise, some already mentioned by @xxdavid in that thread:
- exhaustiveness checking is not done in `type_check_expr` (somewhat recovered in this draft)
- operator support is not as good in `type_check_expr`
- it seems to me that two "parallel" implementations (`type_check_expr` and `type_check_expr_in`) lead to features being added/improved ad-hoc in just one of them, because of similar yet duplicate code and noncomprehensive testing.

What do I mean by the last point? Roughly speaking, to check if syntax for `Expr` is handled properly by the parallel impls, we need two test cases:

```erlang
-spec f1() -> some_ty().
f1() ->
    Expr.

-spec f2() -> some_ty().
f2() ->
    X = Expr,
    X.
```

Where in `f1` `type_check_expr_in` would check that `Expr :: some_ty()` and in `f2` `type_check_expr` would infer the type of `Expr` to assign it to `X` and only then check if `X :: some_ty()`. This is not obvious nor documented anywhere and I think it has lead to the disparity in syntax support in the two impls. From this perspective, merging the two seems to be compelling.

In order to make tests pass in this draft some tests had to be disabled or moved to known problems. I think that if we want to unify these parallel impls, we should bring them all into should pass/fail categories in this PR.


erszcz:
Not for self - an example of a class of the self-check errors we now see is this:

```
$ cat tzt.erl
-module(tzt).

-type options() :: proplists:proplist().

-spec parse_opts([string()], options()) -> {[string()], options()}.
parse_opts([], Opts) ->
    {[], Opts};
parse_opts([A | Args], Opts) ->
    case A of
        "-h"                       -> {[], [help]};
        "--help"                   -> {[], [help]};
        "--verbose"                -> parse_opts(Args, [verbose | Opts])
    end.
$ ./bin/gradualizer tzt.erl
tzt.erl: The list on line 12 at column 56 is expected to have type options() but it has type [verbose | property(), ...]

        "-h"                       -> {[], [help]};
        "--help"                   -> {[], [help]};
        "--verbose"                -> parse_opts(Args, [verbose | Opts])
                                                       ^^^^^^^^^^^^^^^^
```

but

```
$ r3 shell
1> OptionTy1 = gradualizer:type("[verbose | proplists:property(), ...]").
{type,0,nonempty_list,
      [{type,0,union,
             [{atom,0,verbose},
              {remote_type,0,
                           [{atom,1,proplists},{atom,1,property},[]]}]}]}
2> OptionTy2 = gradualizer:type("gradualizer:options()").
{remote_type,0,[{atom,1,gradualizer},{atom,1,options},[]]}
3> typechecker:subtype(OptionTy1, OptionTy2, gradualizer:env()).
true
4>
```

Makefile
```diff
@@ -133,8 +133,8 @@ test/arg.beam: test/should_fail/arg.erl
 .PHONY: build_test_data
 test_data_erls = $(wildcard test/known_problems/**/*.erl test/should_fail/*.erl test/should_pass/*.erl)
 build_test_data:
-	mkdir -p "test_data"
-	erlc $(TEST_ERLC_OPTS) -o test_data $(test_data_erls)
+	@mkdir -p "test_data"
+	@erlc $(TEST_ERLC_OPTS) -o test_data $(test_data_erls)
```
zuiderkwast, line 137: Hehe, I've always been annoyed by build tools that don't show they are doing, i.e. don't print the commands that run (like CMake or Rebar, they hide everything by default). In make you at least see the commands by default. When the commands are printed, you can understand what happens and you can easily repeat the same commands manually.

I'm not going to argue, but I am going to unassign myself as a reviewer of this PR. :)

erszcz: In general I agree, I don't like opaque complexity either. With this number of repetitions, though, I could tell by heart what it's doing ;) I don't really need to see again and again the parts that don't change, I'm more interested in the output which does change. Anyway, this bit is a marginal detail.

zuiderkwast: If the files don't change, they targets shouldn't need to be rebuilt. Some dependency declaration missing here seems to be the cause of repetition.

test/should_pass/return_fun.erl
```diff
@@ -14,17 +17,5 @@ return_fun_union() -> fun nil/0.
 -spec return_fun_any_arity() -> fun((...) -> list()).
 return_fun_any_arity() -> fun nil/0.
 
--spec return_fun_intersection() -> fun((any()) -> integer()).
-return_fun_intersection() -> fun number/1.
```
xxdavid, line 18: I know you just moved this function around, but why should this pass? The spec says that `return_fun_intersection/0` returns a function that returns an integer, but it actually returns a function that can also return a float (if I pass it a float). A function that might return a float should not be a subtype of a function that returns an integer.

test/should_pass/return_fun.erl
```diff
@@ -14,17 +17,5 @@ return_fun_union() -> fun nil/0.
...
-
--spec return_fun_union_intersection()
-                                   -> fun((atom()) -> atom()) |
-                                      fun((1..3) -> integer()).
-return_fun_union_intersection() -> fun number/1.
```
xxdavid, line 23: The same goes for this function.

test/known_problems/should_fail/return_fun_should_fail.erl
```diff
@@ -0,0 +1,8 @@
+-module(return_fun_should_fail).
+
+-export([return_fun_no_spec/0]).
+
+-spec return_fun_no_spec() -> integer().
+return_fun_no_spec() -> fun no_spec/0.
```
xxdavid, line 6: This can be solved by this commit https://github.com/xxdavid/Gradualizer/commit/13d689363e96f0c4f74ebc70b292ac43657a3540
You can cherry pick it if you want.

xxdavid (commented):
Wow, good attempt! Indeed, the only tests failing after merging the two functions seem to be related to anonymous functions, arithmetical operations and exhaustiveness checking. I am not sure with the lambdas but the latter two can probably be implemented inside `type_check_expr`. I've skimmed over the changes and I see that a lot of known problems aren't known problems anymore. That's great!

What are we gonna do now with this draft? Do you still want to pursue merging these functions or was it just a demo of it?

Makefile
xxdavid: I agree that the Makefile can probably be tweaked so that tests are built only when they actually change. I would consider that a better option than just silencing the commands.

src/typechecker.erl
```diff
@@ -2634,11 +2638,28 @@ type_check_comprehension(Env, Compr, Expr, [Guard | Quals]) ->
...
-    ?assert_type(R, env()).
+            [gradualizer_fmt:format_location(Expr, brief),
+             erl_prettypr:format(Expr),
+             typelib:pp_type(ResTy)]),
+    case subtype(InferredTy, ResTy, NewEnv) of
+        true ->
+            NewEnv;
+        false ->
+            throw(type_error(Expr, InferredTy, ResTy))
+    end.
```
josefs, line 2651: The state of the art in writing typecheckers is to do it the other way around, compared to what you've done here. Always pass in the type that's expected, though if we don't know the type just pass a variable. You'll likely get better propagation of types that way.

See e.g. slide 43 in this presentation by Simon Peyton Jones: https://haskell.foundation/assets/other/Simon%20Peyton%20Jones%20-%202023-06-08%20-%20Type%20inference%20in%20GHC%20Jun%2023.pdf
Video can be found here: https://haskell.foundation/events/2023-ghc-development-workshop.html

